PR #11, opened 2023-02-16
Unit Test for FAR Controller

razo7:
- Create new controllers suite for testing with Ginkgo commands.
- Add Ginkgo target.
- Refactor Executer
- Unit Test the build of CLI command and functionality tests.


Makefile
```diff
@@ -162,7 +164,7 @@ go-verify: go-tidy go-vendor # Run go mod verify - verify dependencies have expe
 
 .PHONY: test
 test: manifests generate go-verify fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -i --bin-dir $(LOCALBIN) -p path)"  $(GINKGO) -v -r --keepGoing -requireSuite -coverprofile cover.out
```
slintes, line 167: doesn't work for me, I had to chaneg to:
`	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(ENVTEST_DIR)/$(ENVTEST_VERSION) -p path)"  $(GINKGO) -v -r --keepGoing -requireSuite -coverprofile cover.out`

slintes (changes requested):
So.... I left "some" comments already. But before even understanding the actual test, I discovered  a huge mistake in this operator. Looking further into the unit test does not make sense with it:
 
The FenceAgentsRemediation**Template**'s spec should only have a `template.spec` field which refers to the spec of  FenceAgentsRemediation. Everything else needs to go into the latter. See SNR template as an example.

mshitrit:
>The FenceAgentsRemediationTemplate's spec should only have a template.spec field which refers to the spec of FenceAgentsRemediation. Everything else needs to go into the latter. See SNR template as an example.

IIUC this is a mistake because that makes the naming/convention very confusing (i.e the FAR isn't an imaged of the FAR Template).
Is that indeed the case or were you thinking of different reasons ?

controllers/fenceagentsremediation_controller.go
```diff
@@ -47,8 +46,16 @@ var (
 // FenceAgentsRemediationReconciler reconciles a FenceAgentsRemediation object
 type FenceAgentsRemediationReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
+	Log          logr.Logger
+	Scheme       *runtime.Scheme
+	ExecutorHook func(*corev1.Pod) (cli.Executer, error) // for testing the cli command
```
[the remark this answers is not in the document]
mshitrit: I don't think so (at least not the way the code is constructed at the moment).
Since in the production code we are [fetching the executor](https://github.com/medik8s/fence-agents-remediation/blob/main/controllers/fenceagentsremediation_controller.go#L96) in order to mock the executor we need to mock the fetching part.

controllers/fenceagentsremediation_controller.go
```diff
@@ -68,69 +75,68 @@ type FenceAgentsRemediationReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.2/pkg/reconcile
 func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	r.Log.Info("started reconcile")
-	defer r.Log.Info("finished reconcile")
```
mshitrit, line 72: Since we log the start of reconcile I think it makes sense to print the finish as well

controllers/fenceagentsremediation_controller.go
```diff
@@ -68,69 +75,68 @@ type FenceAgentsRemediationReconciler struct {
...
+		}
+		r.Log.Error(err, "failed to get FAR Template CR")
+		return emptyResult, err
+	}
 
+	// Fetch the FenceAgentsRemediation instance
+	r.Log.Info("Fetch FAR CR")
 	far := &v1alpha1.FenceAgentsRemediation{}
-	if err := r.Get(ctx, req.NamespacedName, far); err != nil {
+	if err := r.Client.Get(ctx, req.NamespacedName, far); err != nil {
```
mshitrit, line 97: Nit: IMO it's nicer to use the fact that `client ` is embedded so I prefer `r.Get`

controllers/fenceagentsremediation_controller.go
```diff
@@ -68,69 +75,68 @@ type FenceAgentsRemediationReconciler struct {
...
 		if apiErrors.IsNotFound(err) {
 			// FAR is deleted, stop reconciling
-			r.Log.Info("Fence Agents Remediation not found, nothing to do")
-			return ctrl.Result{}, nil
+			r.Log.Info("FAR CR is deleted - nothing to do", "CR Name", req.Name, "CR Namespace", req.Namespace)
```
mshitrit, line 100: I don't think we can assume it's deleted :thinking: 
I think creation of the template alone will also get us here.

controllers/fenceagentsremediation_controller.go
```diff
@@ -68,69 +75,68 @@ type FenceAgentsRemediationReconciler struct {
...
+	// Fetch the FAR's pod
+	r.Log.Info("Fetch FAR's pod")
+	pod, err := r.getFenceAgentsPod(req.Namespace)
+	if err != nil {
+		return emptyResult, err
 	}
 
-	pod, err := r.getFAPod(req.NamespacedName.Namespace)
+	// Build CLI executer for FAR's pod
+	r.Log.Info("Build CLI executer for FAR's pod")
```
mshitrit, line 115: Nit: IMO this log is an overkill for each time reconcile is triggered.

controllers/fenceagentsremediation_controller.go
```diff
@@ -68,69 +75,68 @@ type FenceAgentsRemediationReconciler struct {
...
-
-	return fenceAgentParams
-}
-
-// SetupWithManager sets up the controller with the Manager.
-func (r *FenceAgentsRemediationReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&v1alpha1.FenceAgentsRemediation{}).
-		Complete(r)
+	r.Log.Info("Finish FenceAgentsRemediation Reconcile")
```
mshitrit, line 134: I suggest putting this in `defer` at the start otherwise it'll not be printed in case of an early exit.

controllers/fenceagentsremediation_controller.go
```diff
@@ -68,69 +75,68 @@ type FenceAgentsRemediationReconciler struct {
...
+	ex, err := r.ExecutorHook(pod)
 	if err != nil {
-		return ctrl.Result{}, err
+		return emptyResult, err
 	}
 
-	ex, err := cli.NewExecuter(pod)
+	//TODO: Check that FA is excutable? run cli.IsExecuteable
+
+	r.Log.Info("Create and execute the fence agent", "Fence Agent", farTemplate.Spec.Agent)
```
mshitrit, line 123: Nit: IMO this log is an overkill for each time reconcile is triggered.
(Keep in mind that we already have a log in the executor that has this info an more)

controllers/fenceagentsremediation_controller.go
```diff
@@ -146,17 +152,36 @@ func (r *FenceAgentsRemediationReconciler) getFAPod(namespace string) (*corev1.P
...
+// buildFenceAgentParams collects the FAR's parameters for the node based on the FARTemplate and FAR CRs
+func buildFenceAgentParams(farTemplate *v1alpha1.FenceAgentsRemediationTemplate, far *v1alpha1.FenceAgentsRemediation) ([]string, error) {
+	var fenceAgentParams []string
+	for paramName, paramVal := range farTemplate.Spec.SharedParameters {
+		fenceAgentParams = appendParamToSlice(fenceAgentParams, string(paramName), paramVal)
+	}
+
+	nodeName := v1alpha1.NodeName(far.Name)
+	for paramName, nodeMap := range farTemplate.Spec.NodeParameters {
+		if nodeMap[nodeName] != "" {
```
mshitrit, line 174: Nit:
1. I think it would be nicer to use go's built in map empty check:
```
if nodeVal, isFound := nodeMap[nodeName]; isFound {
			fenceAgentParams = appendParamToSlice(fenceAgentParams, string(paramName), nodeVal)
```
2.  I suggest keeping `paramName` type (if necessary convert is as late as possible).

controllers/fenceagentsremediation_controller.go
```diff
@@ -146,17 +152,36 @@ func (r *FenceAgentsRemediationReconciler) getFAPod(namespace string) (*corev1.P
...
+			fenceAgentParams = appendParamToSlice(fenceAgentParams, string(paramName), nodeMap[nodeName])
```
mshitrit, line 175: Nit:  I suggest keeping `paramName` type (if necessary convert is as late as possible).

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
slintes: I see. Refactorig the executor would make the code easier than using the func hook though IMHO.
E.g.:  https://github.com/slintes/fence-agents-remediation/commit/b8a9af412e7ecd9150c557a3bdd910615c8aeb6f

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -0,0 +1,214 @@
...
+		When("creating a resource", func() {
+			It("should have a namespace scoped CR", func() {
+				Expect(underTestFAR.Namespace).To(Not(BeEmpty()))
+			})
+		})
+	})
+
+	Context("Reconcilie", func() {
+		//Scenarios
+		nodeName :=  "master-0"
```
mshitrit, line 68: Nit: even though the name is irrelevant for the test functionality I think that using "master" might be a bit confusing.
I suggest going with the conventional "worker"

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -0,0 +1,214 @@
...
+		Image: "foo",
+	}
+	fenceAgentsPod.Spec.Containers = []corev1.Container{container}
+	return fenceAgentsPod
+}
+
+// cliCommandsEquality creates the command for CLI and compares it with the production command
+func cliCommandsEquality(farTemplate *v1alpha1.FenceAgentsRemediationTemplate, far *v1alpha1.FenceAgentsRemediation) bool {
+	//fence_ipmilan --ip=192.168.111.1 --ipport=6233 --username=admin --password=password --action=status --lanplus
+	command, err := buildFenceAgentParams(farTemplate, far)
```
mshitrit, line 169: better name for `command` would be `expectedCommand`

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -0,0 +1,214 @@
...
+	container := corev1.Container{
+		Name:  "foo",
+		Image: "foo",
+	}
+	fenceAgentsPod.Spec.Containers = []corev1.Container{container}
+	return fenceAgentsPod
+}
+
+// cliCommandsEquality creates the command for CLI and compares it with the production command
+func cliCommandsEquality(farTemplate *v1alpha1.FenceAgentsRemediationTemplate, far *v1alpha1.FenceAgentsRemediation) bool {
```
mshitrit, line 167: Since this method is already called from an `Eventually` block, I suggest removing inner `Expect` verification - just return `false` instead

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -0,0 +1,214 @@
...
+	//fence_ipmilan --ip=192.168.111.1 --ipport=6233 --username=admin --password=password --action=status --lanplus
+	command, err := buildFenceAgentParams(farTemplate, far)
+	Expect(err).NotTo(HaveOccurred())
+	command = append([]string{farTemplate.Spec.Agent}, command...)
+	Expect(executedCommand).ToNot(Equal(nil))
+	fmt.Printf("%s is the executedCommand in prod, and %s is the expected command in test.\n", executedCommand, command)
+	return contains(executedCommand, command) && contains(command, executedCommand)
```
mshitrit, line 174: I think This method is incorrect, check playground example [here](https://go.dev/play/p/VdqRfOXwPbg).
I suggest replacing `contains(executedCommand, command) && contains(command, executedCommand)` with a single method: (feel free to use better naming)
```
func isContainSameElements(s1 []string, s2 []string) bool {
	if len(s1) != len(s2) {
		return false
	}
	s1ElementsAndCount, s2ElementsAndCount := make(map[string]int), make(map[string]int)
	for i := 0; i < len(s1); i++ {
		s1ElementsAndCount[s1[i]]++
		s2ElementsAndCount[s2[i]]++
	}

	for key, _ := range s1ElementsAndCount {
		if s1ElementsAndCount[key] != s2ElementsAndCount[key] {
			return false
		}
	}
	return true

}
```

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
mshitrit: I agree, I think it simplify the code nicely :+1:

main.go
```diff
@@ -24,6 +24,8 @@ import (
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	"github.com/medik8s/fence-agents-remediation/pkg/cli"
```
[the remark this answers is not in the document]
razo7: Why to sort the imports?
Did you mean by using https://github.com/medik8s/node-healthcheck-operator/blob/main/Makefile#L156-L158 ?

controllers/controllers_suite_test.go
```diff
@@ -17,50 +16,51 @@ limitations under the License.
...
+		Development: true,
+	}
+	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseFlagOptions(&opts)))
 
 	By("bootstrapping test environment")
-	testEnv = &envtest.Environment{
-		CRDDirectoryPaths:     []string{filepath.Join("..", "config", "crd", "bases")},
-		ErrorIfCRDPathMissing: true,
-	}
+	testEnv = &envtest.Environment{CRDDirectoryPaths: []string{filepath.Join("..", "config", "crd", "bases")}, ErrorIfCRDPathMissing: true}
```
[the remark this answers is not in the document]
razo7: I can't recall what happened in the process, but it wasn't for the sake of better readability. I can undo this change if you prefer :)

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -0,0 +1,214 @@
...
+		templateTest := newFenceAgentsRemediationTemplate(" ", testShareParam, testNodeParam)
+		test := &v1alpha1.FenceAgentsRemediation{
+			ObjectMeta: metav1.ObjectMeta{Name: nodeName, Namespace: defaultNamespace},
+		}
+		fenceAgentsPod := buildFarPod()
+
+		JustBeforeEach(func() {
+			err := k8sClient.Create(context.Background(), test)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(test.Namespace).To(Not(BeEmpty()))
```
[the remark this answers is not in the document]
razo7: It doesn't hurt to this check ;)  But I agree that it doesn't add. 
Do you see a use for checking the namespace in E2E test?

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
razo7: Nice idea @slintes !

controllers/fenceagentsremediation_controller.go
razo7: Oho, now I understand the advantage of `defer`.

controllers/fenceagentsremediation_controller.go
razo7: Creation or deletion of FARTemplete won't trigger the reconcile since it listen/watch for FAR CRs.
If we have reached to [line 97](https://github.com/medik8s/fence-agents-remediation/pull/11/files/260d55261948ed3d0fee45279adee1d7f3a30640#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R97) then FAR CR has been created or deleted, and were able to fetch the FARTemplete. Otherwise we should have received an error from [line 84 ](https://github.com/medik8s/fence-agents-remediation/pull/11/files/260d55261948ed3d0fee45279adee1d7f3a30640#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R84) and never reach this far ;)

controllers/fenceagentsremediation_controller.go
razo7: Nice

controllers/fenceagentsremediation_controller.go
razo7: I don't understand your suggestion. May you elaborate please?

controllers/fenceagentsremediation_controller_test.go
razo7: Good example, and better function with much lower complexity (o(n) vs o(n^2))

main.go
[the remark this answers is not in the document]
slintes: > Why to sort the imports?

because thet are unsorted?!

controllers/fenceagentsremediation_controller_test.go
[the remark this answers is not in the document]
slintes: You want to test FAR functionality, do you? Having a namespace on a CR, after setting a namespace on a CR... doesn't do that 🤷🏼‍♂️ Neither in unit test, nor in e2e.

controllers/fenceagentsremediation_controller_test.go
slintes:
```
sort.Strings(s1)
sort.Strings(s2)
return reflect.DeepEqual(s1, s2)
```

main.go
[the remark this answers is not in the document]
razo7: I meant why do we care about sorting the imports? I don't mind of doing that
It is just new to me, and I just wonder since ATM in other operators we don't have it (e.g. NMO).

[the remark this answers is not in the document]
slintes: We always sort imports. If you find unsorted imports somewhere, it was just missed in reviews

controllers/fenceagentsremediation_controller.go
slintes: what a nice example for why we shouldn't use the template CR for anything but being the literally template for the real remediation CRs. It only causes confusion...

main.go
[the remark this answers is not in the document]
razo7: Ok, then I will do that in a follow up PR.
What do you thing about moving https://github.com/slintes/sort-imports to be under [medik8s tools](https://github.com/medik8s/tools)?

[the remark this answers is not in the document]
razo7: https://github.com/medik8s/fence-agents-remediation/pull/15

controllers/fenceagentsremediation_controller.go
mshitrit: Sure, here is a code snippet:

 
nodeName := v1alpha1.NodeName(far.Name)
for paramName, nodeMap := range farTemplate.Spec.NodeParameters {
	if nodeVal, isFound := nodeMap[nodeName]; isFound {
		fenceAgentParams = appendParamToSlice(fenceAgentParams, **paramName**, nodeVal)
...
 

 
func appendParamToSlice(fenceAgentParams []string, paramName **v1alpha1.ParameterName**, paramVal string) []string {
	if paramVal != "" {
		fenceAgentParams = append(fenceAgentParams, fmt.Sprintf("%s=%s", paramName, paramVal))
	} else {
		fenceAgentParams = append(fenceAgentParams, **string(paramName)**)
...

main.go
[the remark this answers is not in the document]
slintes: it's not releated to medik8s, so I don't think that would be the right home for it

controllers/fenceagentsremediation_controller.go
razo7: Got it :)

controllers/fenceagentsremediation_controller.go
razo7: https://github.com/medik8s/fence-agents-remediation/pull/16 will hopefully resolve that 😸

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/medik8s/fence-agents-remediation/api/v1alpha1"
+	"github.com/medik8s/fence-agents-remediation/pkg/cli"
+)
+
+const (
+	defaultNamespace = "default"
+)
+
+var (
+	executedCommand []string
```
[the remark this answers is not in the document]
razo7: I agree.

I stored the _**executedCommand**_ in the [mock executor field](https://github.com/medik8s/fence-agents-remediation/pull/11/files#diff-fac0cf56dda42683c1e24b46c87cfa4cca46c444145ee69de9e264db50f10289R198), and I am not sure how to call the _cliCommandsEquality_ function in a way that use the same mock executer (otherwise, I will have a pointer to mockExecuter but without the _executedCommand_ value).
Is there any similar example I can follow?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -0,0 +1,214 @@
...
+			It("should have a namespace scoped CR", func() {
+				Expect(underTestFAR.Namespace).To(Not(BeEmpty()))
+			})
+		})
+	})
+
+	Context("Reconcilie", func() {
+		//Scenarios
+		nodeName :=  "master-0"
+		testFields := []string{"--username", "--password", "--action", "--ip", "--lanplus", "--ipport"}
```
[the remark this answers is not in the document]
razo7: These fields are for the _shared_ and _node_ parameters, for instance, **"--ipport"** is not shared.
In https://github.com/medik8s/fence-agents-remediation/pull/11/files#diff-fac0cf56dda42683c1e24b46c87cfa4cca46c444145ee69de9e264db50f10289R115-R123 I keep iterating this string array for the node parameters...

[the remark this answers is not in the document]
razo7: Looking on this again -  I do understand the confusion with the **testValues** variable that only has the values of shared parameters.. 🤔

controllers/fenceagentsremediation_controller_test.go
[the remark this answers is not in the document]
razo7: No more global vars over there, thanks to @mshitrit  🙏🏻

controllers/fenceagentsremediation_controller_test.go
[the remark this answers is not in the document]
slintes: The confusion is more about how the parameters are build. First 4 string arrays, and then a magic buildFARParameters func which puts them into a mixer and returns 2 maps 😉 

IMHO building the params directly would be better readable

```
	testShareParam := map[v1alpha1.ParameterName]string{
		"name": "value",
	}
	testNodeParam := map[v1alpha1.ParameterName]map[v1alpha1.NodeName]string{
		"name": {
			"node": "value",
		},
	}
```

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -0,0 +1,195 @@
...
+func cliCommandsEquality(far *v1alpha1.FenceAgentsRemediation) (bool, error) {
+	//fence_ipmilan --ip=192.168.111.1 --ipport=6233 --username=admin --password=password --action=status --lanplus
+	if mocksExecuter.command == nil {
+		return false, errors.New("executedCommand is null")
+	}
+	expectedCommand, err := buildFenceAgentParams(far)
+	if err != nil {
+		return false, err
+	}
+	expectedCommand = append([]string{far.Spec.Agent}, expectedCommand...)
```
slintes, line 149: We use the same code for building the command here in the test and in the controller. So IIUC the only thing which is tested here is whether Execute was called or not. It's impossible that the command is not the same.

I would like to see a that buildFenceAgentParams() is tested as well. One possibily is to hardcode the expected command here instead of calling buildFenceAgentParams()

slintes (changes requested):

controllers/fenceagentsremediation_controller_test.go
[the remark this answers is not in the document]
slintes: thanks!

controllers/fenceagentsremediation_controller_test.go
razo7: > I would like to see that buildFenceAgentParams() is tested as well. One possibily is to hardcode the expected command here instead of calling buildFenceAgentParams()

I am not sure I follow. In the beginning I understand that you are suggesting to call buildFenceAgentParams() and then you suggest to hardcode the expected command and don't use that function.

razo7: > expectedCommand, err := buildFenceAgentParams(far)
> 	if err != nil {
> 		return false, err
> 	}

If the intention was to still using this function and then hardcode the expected command, so I would keep the above lines to test _buildFenceAgentParams()_  and then explicitly assign the expected command for the _expectedCommand_ variable as below:

 ```suggestion
	// hardcode expected command
	expectedCommand = []string{"","--lanplus","--password=password","--username=admin","--action=reboot","--ip=192.168.111.1","--ipport=6233"}
```

Did I get it right?

Comment (slintes) [reply controllers/fenceagentsremediation_controller_test.go]: > you are suggesting to call buildFenceAgentParams()
sorry if I did so (can't remember), I probably didn't understand the complete test yet because of all the confusion around FAR and FARTemplate.
The change now looks almost good, I think you need to add the agent's name as first string value (also hardcoded)?

File: controllers/fenceagentsremediation_controller_test.go [hunk 21]
```diff
@@ -140,14 +142,15 @@ func buildFarPod() *corev1.Pod {
 func cliCommandsEquality(far *v1alpha1.FenceAgentsRemediation) (bool, error) {
 	//fence_ipmilan --ip=192.168.111.1 --ipport=6233 --username=admin --password=password --action=status --lanplus
 	if mocksExecuter.command == nil {
-		return false, errors.New("executedCommand is null")
+		return false, errors.New("The command from mocksExecuter is null")
 	}
-	expectedCommand, err := buildFenceAgentParams(far)
+	_, err := buildFenceAgentParams(far)
```
slintes, line 147: I don't think you need this, it's tested in the Funtionionality block already?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -65,17 +65,19 @@ var _ = Describe("FAR Controller", func() {
...
-				underTestFAR.ObjectMeta.Name = dummyNodeName
-				_, err := buildFenceAgentParams(underTestFAR)
-				Expect(err).To(HaveOccurred())
-			})
-			It("should succeed when FAR pod has been created", func() {
-				underTestFAR.ObjectMeta.Name = validNodeName
-				_, err := buildFenceAgentParams(underTestFAR)
-				Expect(err).NotTo(HaveOccurred())
+	Context("Functionality", func() {
+		 Context("buildFenceAgentParams", func() {
```
slintes, line 69: wrong indentation (gofmt?)

controllers/fenceagentsremediation_controller_test.go
razo7: Correct, it is redundant 👍🏻

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+// cliCommandsEquality creates the command for CLI and compares it with the production command
+func cliCommandsEquality(far *v1alpha1.FenceAgentsRemediation) (bool, error) {
+	//fence_ipmilan --ip=192.168.111.1 --ipport=6233 --username=admin --password=password --action=status --lanplus
+	if mocksExecuter.command == nil {
+		return false, errors.New("The command from mocksExecuter is null")
+	}
+
+	// hardcode expected command
+	expectedCommand := []string{" ", "--lanplus", "--password=password", "--username=admin", "--action=reboot", "--ip=192.168.111.1", "--ipport=6233"}
```
slintes, line 149: one last comment 😉 
would it make sense to set a better agent in line 65 than `" "` and test for it here? 🙂

razo7: It wouldn't hurt 😄

slintes (approved):
I think adding this test resulted in much better code in the end. Also the test itself evolved nice over time. Good work!

PR #12, opened 2023-02-16
GitHub CI Tests

razo7:
- Add pre-submit, and post-submit jobs.
- Add `test-no-verify` target.

### DIFF
--- a/.github/workflows/post-submit.yaml
+++ b/.github/workflows/post-submit.yaml
@@ -1,0 +1,33 @@
+name: Publish Container Images
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v*'
+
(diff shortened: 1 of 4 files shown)


razo7:
Merge #11 first in order to test the pre-submit job

slintes:
> Merge #11 first in order to test the pre-submit job

I think it makes more sense to merge this one first, so the new unit test is actually executed on its PR?

/lgtm
/hold

razo7:
> > Merge #11 first in order to test the pre-submit job
> 
> I think it makes more sense to merge this one first, so the new unit test is actually executed on its PR?
> 
> /lgtm /hold

It does make sense. 
/unhold

PR #14, opened 2023-02-16
Fix url-install-tool Target

razo7:
Missing OS and ARCH variables for downloading and installing opm and operator-sdk.


clobrano:
Tested :+1: 
/lgtm

PR #16, opened 2023-02-21
Align FAR CRDs to NHC

razo7:
FAR's CRDs didn't follow the motivation/desgin of the other remediators (e.g. [MDRTemplate Spec](https://github.com/medik8s/machine-deletion-remediation/blob/main/api/v1alpha1/machinedeletionremediationtemplate_types.go#L35)) - remediator should have template field under Spec. This PR should align FAR CRDs to NHC/MHC expectations

- Move the Spec fields from FARTemplate to FAR and let the FARTemplate to only have a template field under spec (of `FenceAgentsRemediationSpec` type).
- Add Cluster Role for Role aggregation. Since that will give NHC/MHC access to the template and remediation CRs with a special label.
- Update Reconcile according to this change.


razo7:
Github CI has failed and https://github.com/medik8s/fence-agents-remediation/pull/17 should resolve it.

controllers/fenceagentsremediation_controller.go
```diff
@@ -70,69 +69,54 @@ type FenceAgentsRemediationReconciler struct {
...
-func (r *FenceAgentsRemediationReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&v1alpha1.FenceAgentsRemediation{}).
-		Complete(r)
+	return emptyResult, nil
 }
 
-func (r *FenceAgentsRemediationReconciler) getFAPod(namespace string) (*corev1.Pod, error) {
+// getFenceAgentsPod fetches the FAR pod based on FAR's label and namespace
+func (r *FenceAgentsRemediationReconciler) getFenceAgentsPod(namespace string) (*corev1.Pod, error) {
```
slintes, line 119: something for a follow up, not this PR: you don't need to search the pod, you can know its namespace and name, see here for how to get the namespace, same should work for the name: https://github.com/medik8s/node-healthcheck-operator/blob/main/config/manager/manager.yaml#L53

slintes (commented):
lgtm in general, two nits inside, will let Michael do a review as well

controllers/fenceagentsremediation_controller.go
razo7: Good point, and it sounds like a better coding

api/v1alpha1/fenceagentsremediation_types.go
```diff
@@ -47,9 +59,9 @@ type FenceAgentsRemediation struct {
 	Status FenceAgentsRemediationStatus `json:"status,omitempty"`
 }
 
-//+kubebuilder:object:root=true
-
+// +kubebuilder:object:root=true
 // FenceAgentsRemediationList contains a list of FenceAgentsRemediation
+// +operator-sdk:csv:customresourcedefinitions:resources={{"FenceAgentsRemediation","v1alpha1","fenceagentsremediations"}}
```
razo7, line 64: IDK why but I assumed that this line would add `FenceAgentsRemediation` to the CSV as `FenceAgentsRemediationTemplate` has been added by [similar line](https://github.com/medik8s/fence-agents-remediation/pull/16/files#diff-9f945e8a1bb4d6890949b218fb8e67f95bb5f04b75cbea808da1291121af2dbaR49) to the [CSV](https://github.com/medik8s/fence-agents-remediation/pull/16/files#diff-370992a193637c0aec12471688d0a6310eb7ce759bcd5aaa18d4e0d522108afdR43-R46).

mshitrit: IIUC it's not the case because `FenceAgentsRemediation` is already part of the [CSV](https://github.com/medik8s/fence-agents-remediation/pull/16/files#diff-370992a193637c0aec12471688d0a6310eb7ce759bcd5aaa18d4e0d522108afdR22-R24)

mshitrit:
I would have felt a lot better if we had some unit tests to verify this kind of a change, but I guess in this case it's an egg and chicken scenario :slightly_smiling_face: 
/lgtm

api/v1alpha1/fenceagentsremediation_types.go
razo7: `FenceAgentsRemediation` is indeed part of the CSV.
My doubt was about the `resources` field which is missing for `FenceAgentsRemediation` in comparison to `FenceAgentsRemediationTemplate` in [these lines](https://github.com/medik8s/fence-agents-remediation/blob/main/bundle/manifests/fence-agents-remediation.clusterserviceversion.yaml#L59-L103).
I have compared it with [SNR CSV](https://github.com/medik8s/self-node-remediation/blob/main/bundle/manifests/self-node-remediation.clusterserviceversion.yaml#L52-L99), and I see `resources` field per CRD, unlike for FAR...

slintes: I think that tag needs to go to the FenceAgentsRemediation struct, not the list

razo7: Yes, it made the trick 😄 
thanks @slintes

PR #19, opened 2023-02-28
Align Makefile - Goimports instead of go fmt

razo7:
In the process of moving FAR to Medik8s organization we align the Makefile - **sixth** PR.

Use Goimports instead of `go fmt` since Goimports equals to running `go fmt` and fixing imports.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,8 @@ CONTROLLER_GEN_VERSION ?= v0.8.0
 # See https://pkg.go.dev/sigs.k8s.io/controller-runtime/tools/setup-envtest?tab=versions for the last version
 ENVTEST_VERSION ?= v0.0.0-20221022092956-090611b34874
 # See https://pkg.go.dev/golang.org/x/tools/cmd/goimports?tab=versions for the last version
+GOIMPORTS_VERSION ?= v0.6.0
+# See github.com/operator-framework/operator-registry/releases for the last version
 OPM_VERSION ?= v1.26.2
 # See github.com/operator-framework/operator-sdk/releases for the last version
 OPERATOR_SDK_VERSION ?= v1.26.0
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -143,12 +145,12 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 .PHONY: fmt
-fmt: ## Run go fmt against code.
-	go fmt ./...
+fmt: goimports ## Run go goimports against code - goimports = go fmt + fixing imports.
+	$(GOIMPORTS) -w  ./main.go ./api ./controllers ./test
```
slintes, line 149: `./pkg` is missing. Also there is no `./test` in this repo (yet), not sure if goimports comlains about it?

Copy / paste is ok, but always try to understand what the copied code is doing, and if it works in your own project 🙂

Makefile
```diff
@@ -143,12 +145,12 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
...
 
 .PHONY: vet
-vet: ## Run go vet against code.
-	go vet ./...
+vet: ## Run go vet against code - report likely mistakes in packages.
+	go vet  ./api/... ./controllers/... ./test/...
```
slintes, line 153: pkg and main.go is missing

slintes (changes requested):

Makefile
```diff
@@ -144,11 +146,11 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
...
 .PHONY: fmt
-fmt: ## Run go fmt against code.
-	go fmt ./...
+fmt: goimports ## Run go goimports against code - goimports = go fmt + fixing imports.
+	$(GOIMPORTS) -w  ./main.go ./pkg ./api ./controllers
 
 .PHONY: vet
-vet: ## Run go vet against code.
+vet: ## Run go vet against code - report likely mistakes in packages.
 	go vet ./...
```
slintes, line 154: ooooohhhh.... this is new! Well, for me at least 😎. `./...` used to also match the `vendor` dir, that's why we started listing our files/dirs manually. Great to see that it doesn't anymore. Well, since Go 1.9 already?! 🙈 Just  to be sure, `goimports` is still trying to modify files in `vendor`? 
https://go.dev/doc/go1.9#vendor-dotdotdot

slintes:
/lgtm

One question about goimports inside, feel free to unhold if `goimports` doesn't work with `./...`
/hold

Makefile
razo7: `goimports` v0.6.0 didn't  work with `./...`, so unless we add `./vendor` then goimports won't modify the files under vendor.

PR #20, opened 2023-03-06
FAR E2E Test - Check Node Reboot by Boot Time and Patch CredentialsRequest for AWS

razo7:
The PR adds the support of running a reboot of OCP node on AWS environment, and the E2E test that checks FAR's code.

Until now the E2E test checked if FAR's CR has been created and if the FA CLI command has been executed correctly by reviewing FAR's pod/container logs (both have been done in #32). Now we check the node's boot time (Kubelet ready condition status transition time) to verify that the FA has been doing a successful reboot.

But for running the **fence_aws** fence agent with a `reboot` action outside of AWS we use the **--skip-race-check** flag, and we patch a CredentialsRequest in OCP to add missing AWS permissions (e.g., _ec2:StartInstances_, and _ec2:StopInstances_).

[ECOPROJECT-1274](https://issues.redhat.com//browse/ECOPROJECT-1274)

### DIFF
--- a/pkg/utils/nodes.go
+++ b/pkg/utils/nodes.go
@@ -8,6 +8,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const WorkerLabelName = "node-role.kubernetes.io/worker"
+
 // getNodeWithName returns a node with a name nodeName, or an error if it can't be found
 func getNodeWithName(r client.Reader, nodeName string) (*corev1.Node, error) {
 	node := &corev1.Node{}
--- a/pkg/utils/pods.go
+++ b/pkg/utils/pods.go
@@ -10,10 +10,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// GetFenceAgentsRemediationPod fetches the FAR pod based on FAR's label and namespace
+// GetFenceAgentsRemediationPod fetches the first running pod that matches to FAR's label and namespace
+// The pod should be on running state since when we taint and reboot a node which had FAR, then that pod will be restarted on a different node,
+// This results with an old pod which is about to die and new pod is already running and we want to return the running pod
 func GetFenceAgentsRemediationPod(r client.Reader) (*corev1.Pod, error) {
-	var podNamespace string
-	pods := &corev1.PodList{}
+	podList := &corev1.PodList{}
 	selector := labels.NewSelector()
 	requirement, _ := labels.NewRequirement("app", selection.Equals, []string{"fence-agents-remediation-operator"})
 	selector = selector.Add(*requirement)
@@ -21,12 +22,17 @@ func GetFenceAgentsRemediationPod(r client.Reader) (*corev1.Pod, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed fetching FAR namespace - %w", err)
 	}
-	err = r.List(context.Background(), pods, &client.ListOptions{LabelSelector: selector, Namespace: podNamespace})
+	err = r.List(context.Background(), podList, &client.ListOptions{LabelSelector: selector, Namespace: podNamespace})
 	if err != nil {
 		return nil, fmt.Errorf("failed fetching FAR pod - %w", err)
 	}
-	if len(pods.Items) == 0 {
-		return nil, fmt.Errorf("no Fence Agent pods were found")
+	if len(podList.Items) == 0 {
+		return nil, fmt.Errorf("no FAR pods were found")
 	}
-	return &pods.Items[0], nil
+	for _, pod := range podList.Items {
+		if pod.Status.Phase == corev1.PodRunning {
+			return &pod, nil
+		}
+	}
+	return nil, fmt.Errorf("no running FAR pods were found")
 }
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -21,17 +21,20 @@ import (
 )
 
 const (
-	fenceAgentDummyName  = "echo"
-	fenceAgentAWS        = "fence_aws"
-	fenceAgentIPMI       = "fence_ipmilan"
-	fenceAgentAction     = "status"
-	nodeIndex            = 0
-	succeesStatusMessage = "ON"
-	containerName        = "manager"
+	fenceAgentAWS            = "fence_aws"
+	fenceAgentIPMI           = "fence_ipmilan"
+	fenceAgentAction         = "reboot"
+	nodeIdentifierPrefixAWS  = "--plug"
+	nodeIdentifierPrefixIPMI = "--ipport"
+	nodeIndex                = 3
+	succeesRebootMessage     = "\"Success: Rebooted"
+	containerName            = "manager"
 
 	// eventually parameters
-	timeoutLogs  = 1 * time.Minute
-	pollInterval = 10 * time.Second
+	//TODO: try to minimize timeout
+	timeoutLogs   = 6 * time.Minute
+	timeoutReboot = 10 * time.Minute
+	pollInterval  = 10 * time.Second
 )
 
 var _ = Describe("FAR E2e", func() {
@@ -49,15 +52,22 @@ var _ = Describe("FAR E2e", func() {
 		fmt.Printf("\ncluster name: %s and PlatformType: %s \n", string(clusterPlatform.Name), string(clusterPlatform.Status.PlatformStatus.Type))
 	})
 
-	Context("fence agent", func() {
+	Context("fence agent - fence_aws or fence_ipmilan", func() {
+		var (
+			nodeBootTimeBefore   time.Time
+			errBoot              error
+			testNodeName         string
+			nodeIdentifierPrefix string
+			testNodeID           string
+		)
 		BeforeEach(func() {
 			nodes := &corev1.NodeList{}
 			Expect(k8sClient.List(context.Background(), nodes, &client.ListOptions{})).ToNot(HaveOccurred())
 			if len(nodes.Items) <= 1 {
 				Fail("there is one or less available nodes in the cluster")
 			}
-			//TODO: Randomize the node selection
-			// run FA on the first node - a master node
+			//TODO: Randomize the node selection & verify valid index
+			// run FA on the fourth node - a worker node
 			nodeObj := nodes.Items[nodeIndex]
 			testNodeName := nodeObj.Name
 			log.Info("Testing Node", "Node name", testNodeName)
@@ -65,9 +75,11 @@ var _ = Describe("FAR E2e", func() {
 			switch clusterPlatform.Status.PlatformStatus.Type {
 			case configv1.AWSPlatformType:
 				fenceAgent = fenceAgentAWS
+				nodeIdentifierPrefix = nodeIdentifierPrefixAWS
 				By("running fence_aws")
 			case configv1.BareMetalPlatformType:
 				fenceAgent = fenceAgentIPMI
+				nodeIdentifierPrefix = nodeIdentifierPrefixIPMI
 				By("running fence_ipmilan")
 			default:
 				Skip("FAR haven't been tested on this kind of cluster (non AWS or BareMetal)")
@@ -81,12 +93,17 @@ var _ = Describe("FAR E2e", func() {
 			if err != nil {
 				Fail("can't get node information")
 			}
+			nodeName := v1alpha1.NodeName(testNodeName)
+			parameterName := v1alpha1.ParameterName(nodeIdentifierPrefix)
+			testNodeID = testNodeParam[parameterName][nodeName]
+			log.Info("Testing Node", "Node name", testNodeName, "Node ID", testNodeID)
+
+			// save the node's boot time prior to the fence agent call
+			nodeBootTimeBefore, errBoot = getNodeBootTime(testNodeName)
+			Expect(errBoot).ToNot(HaveOccurred(), "failed to get boot time of the node")
 
 			far = createFAR(testNodeName, fenceAgent, testShareParam, testNodeParam)
-		})
-
-		AfterEach(func() {
-			deleteFAR(far)
+			DeferCleanup(deleteFAR, far)
 		})
 
 		When("running FAR to reboot node ", func() {
@@ -96,7 +113,10 @@ var _ = Describe("FAR E2e", func() {
 				Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(far), testFarCR)).To(Succeed(), "failed to get FAR CR")
 
 				By("checking the command has been executed successfully")
-				checkFarLogs(succeesStatusMessage)
+				checkFarLogs(succeesRebootMessage)
+
+				By("checking the node's boot time after running the FA")
+				wasNodeRebooted(testNodeName, nodeBootTimeBefore)
 			})
 		})
 	})
@@ -197,7 +217,7 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
 			fmt.Printf("can't get nodes' information - AWS instance ID\n")
 			return nil, err
 		}
-		nodeIdentifier = v1alpha1.ParameterName("--plug")
+		nodeIdentifier = v1alpha1.ParameterName(nodeIdentifierPrefixAWS)
 
 	} else if clusterPlatformType == configv1.BareMetalPlatformType {
 		nodeListParam, err = e2eUtils.GetBMHNodeInfoList(machineClient)
@@ -205,10 +225,19 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
 			fmt.Printf("can't get nodes' information - ports\n")
 			return nil, err
 		}
-		nodeIdentifier = v1alpha1.ParameterName("--ipport")
+		nodeIdentifier = v1alpha1.ParameterName(nodeIdentifierPrefixIPMI)
 	}
 	testNodeParam = map[v1alpha1.ParameterName]map[v1alpha1.NodeName]string{nodeIdentifier: nodeListParam}
 	return testNodeParam, nil
+}
+
+// getNodeBootTime returns the bootime of node nodeName if possible, otherwise it returns an error
+func getNodeBootTime(nodeName string) (time.Time, error) {
+	bootTime, err := e2eUtils.GetBootTime(clientSet, nodeName, testNsName, log)
+	if bootTime != nil && err == nil {
+		return *bootTime, nil
+	}
+	return time.Time{}, err
 }
 
 // checkFarLogs gets the FAR pod and checks whether it's logs have logString
@@ -232,4 +261,21 @@ func checkFarLogs(logString string) {
 		}
 		return logs
 	}, timeoutLogs, pollInterval).Should(ContainSubstring(logString))
+}
+
+// wasNodeRebooted waits until there is a newer boot time than before, a reboot occurred, otherwise it falls with an error
+func wasNodeRebooted(nodeName string, nodeBootTimeBefore time.Time) {
+	log.Info("boot time", "node", nodeName, "old", nodeBootTimeBefore)
+	var nodeBootTimeAfter time.Time
+	Eventually(func() (time.Time, error) {
+		var errBootAfter error
+		nodeBootTimeAfter, errBootAfter = getNodeBootTime(nodeName)
+		if errBootAfter != nil {
+			log.Error(errBootAfter, "Can't get boot time of the node")
+		}
+		return nodeBootTimeAfter, errBootAfter
+	}, timeoutReboot, pollInterval).Should(
+		BeTemporally(">", nodeBootTimeBefore), "Timeout for node reboot has passed, even though FAR CR has been created")
+
+	log.Info("successful reboot", "node", nodeName, "offset between last boot", nodeBootTimeAfter.Sub(nodeBootTimeBefore), "new boot time", nodeBootTimeAfter)
 }
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -33,8 +33,9 @@ const (
 
 	//TODO: try to minimize timeout
 	// eventually parameters
-	timeoutLogs   = 6 * time.Minute
-	timeoutReboot = 10 * time.Minute // fencing with fence_aws should be completed within 6 minutes
+	timeoutGetPod = 1 * time.Minute
+	timeoutLogs   = 3 * time.Minute
+	timeoutReboot = 6 * time.Minute // fencing with fence_aws should be completed within 6 minutes
 	pollInterval  = 10 * time.Second
 )
 
@@ -259,7 +260,7 @@ func getFarPod() *corev1.Pod {
 			return nil
 		}
 		return pod
-	}, timeoutLogs, pollInterval).ShouldNot(BeNil(), "can't find the pod after timeout")
+	}, timeoutGetPod, pollInterval).ShouldNot(BeNil(), "can't find the pod after timeout")
 	return pod
 }
 
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -33,7 +33,7 @@ const (
 	// eventually parameters
 	//TODO: try to minimize timeout
 	timeoutLogs   = 6 * time.Minute
-	timeoutReboot = 10 * time.Minute
+	timeoutReboot = 10 * time.Minute // fencing with fence_aws should be completed within 6 minutes
 	pollInterval  = 10 * time.Second
 )
 
@@ -151,22 +151,25 @@ func deleteFAR(far *v1alpha1.FenceAgentsRemediation) {
 func buildSharedParameters(clusterPlatform *configv1.Infrastructure, action string) (map[v1alpha1.ParameterName]string, error) {
 	const (
 		//AWS
-		secretAWS    = "aws-cloud-credentials"
-		secretKeyAWS = "aws_access_key_id"
-		secretValAWS = "aws_secret_access_key"
+		// secretDefaultAWS    = "aws-cloud-credentials"
+		secretAWSName      = "aws-cloud-fencing-credentials-secret"
+		secretAWSNamespace = "openshift-operators"
+		secretKeyAWS       = "aws_access_key_id"
+		secretValAWS       = "aws_secret_access_key"
 
 		// BareMetal
 		//TODO: secret BM should be based on node name - > oc get bmh -n openshift-machine-api BM_NAME -o jsonpath='{.spec.bmc.credentialsName}'
-		secretBMHExample = "ostest-master-0-bmc-secret"
-		secretKeyBM      = "username"
-		secretValBM      = "password"
+		secretBMHName      = "ostest-master-0-bmc-secret"
+		secretBMHNamespace = "openshift-machine-api"
+		secretKeyBM        = "username"
+		secretValBM        = "password"
 	)
 	var testShareParam map[v1alpha1.ParameterName]string
 
 	// oc get Infrastructure.config.openshift.io/cluster -o jsonpath='{.status.platformStatus.type}'
 	clusterPlatformType := clusterPlatform.Status.PlatformStatus.Type
 	if clusterPlatformType == configv1.AWSPlatformType {
-		accessKey, secretKey, err := e2eUtils.GetCredentials(clientSet, secretAWS, secretKeyAWS, secretValAWS)
+		accessKey, secretKey, err := e2eUtils.GetCredentials(clientSet, secretAWSName, secretAWSNamespace, secretKeyAWS, secretValAWS)
 		if err != nil {
 			fmt.Printf("can't get AWS credentials\n")
 			return nil, err
@@ -176,17 +179,18 @@ func buildSharedParameters(clusterPlatform *configv1.Infrastructure, action stri
 		regionAWS := string(clusterPlatform.Status.PlatformStatus.AWS.Region)
 
 		testShareParam = map[v1alpha1.ParameterName]string{
-			"--access-key": accessKey,
-			"--secret-key": secretKey,
-			"--region":     regionAWS,
-			"--action":     action,
+			"--access-key":      accessKey,
+			"--secret-key":      secretKey,
+			"--region":          regionAWS,
+			"--action":          action,
+			"--skip-race-check": "",
 			// "--verbose":    "", // for verbose result
 		}
 	} else if clusterPlatformType == configv1.BareMetalPlatformType {
 		// TODO : get ip from GetCredientals
 		// oc get bmh -n openshift-machine-api ostest-master-0 -o jsonpath='{.spec.bmc.address}'
 		// then parse ip
-		username, password, err := e2eUtils.GetCredentials(clientSet, secretBMHExample, secretKeyBM, secretValBM)
+		username, password, err := e2eUtils.GetCredentials(clientSet, secretBMHName, secretBMHNamespace, secretKeyBM, secretValBM)
 		if err != nil {
 			fmt.Printf("can't get BMH credentials\n")
 			return nil, err
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -246,10 +246,12 @@ func getNodeBootTime(nodeName string) (time.Time, error) {
 	return time.Time{}, err
 }
 
-// checkFarLogs gets the FAR pod and checks whether it's logs have logString
-func checkFarLogs(logString string) {
-	var pod *corev1.Pod
-	var err error
+// getFarPod gets the FAR pod by it's label
+func getFarPod() *corev1.Pod {
+	var (
+		pod *corev1.Pod
+		err error
+	)
 	EventuallyWithOffset(1, func() *corev1.Pod {
 		pod, err = utils.GetFenceAgentsRemediationPod(k8sClient)
 		if err != nil {
@@ -258,10 +260,22 @@ func checkFarLogs(logString string) {
 		}
 		return pod
 	}, timeoutLogs, pollInterval).ShouldNot(BeNil(), "can't find the pod after timeout")
+	return pod
+}
 
+// checkFarLogs gets the FAR pod and checks whether it's logs have logString
+func checkFarLogs(logString string) {
+	pod := getFarPod()
 	EventuallyWithOffset(1, func() string {
 		logs, err := e2eUtils.GetLogs(clientSet, pod, containerName)
 		if err != nil {
+			if apiErrors.IsNotFound(err) {
+				// If FAR pod was running in testNodeName, then after reboot it was recreated in another node, and with a new name.
+				// Thus the "old" pod's name prior to this eventually won't link to a running pod, since it was already evicted by the reboot
+				log.Error(err, "failed to get logs. FAR pod might have been recreated due to rebooting the node it was resided. Might try again")
+				pod = getFarPod()
+				return ""
+			}
 			log.Error(err, "failed to get logs. Might try again")
 			return ""
 		}
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -31,8 +31,8 @@ const (
 	succeesRebootMessage     = "\"Success: Rebooted"
 	containerName            = "manager"
 
-	// eventually parameters
 	//TODO: try to minimize timeout
+	// eventually parameters
 	timeoutLogs   = 6 * time.Minute
 	timeoutReboot = 10 * time.Minute // fencing with fence_aws should be completed within 6 minutes
 	pollInterval  = 10 * time.Second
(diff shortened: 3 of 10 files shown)


test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,171 @@
...
+	if len(pods.Items) == 0 {
+		podNotFoundErr := &apiErrors.StatusError{ErrStatus: metav1.Status{
+			Status: metav1.StatusFailure,
+			Code:   http.StatusNotFound,
+			Reason: metav1.StatusReasonNotFound,
+		}}
+		return nil, podNotFoundErr
+	}
+	return &pods.Items[0], nil
+}
```
razo7, line 171: I have copy paste the function from the [controller](https://github.com/medik8s/fence-agents-remediation/blob/main/controllers/fenceagentsremediation_controller.go#L112-L137), since I don't have the reconciler object in far_e2e_test.go.
I can call this function from  far_e2e_test.go but it would require removing the reconciler which is needed for listing the objects in the controller

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	"github.com/medik8s/fence-agents-remediation/api/v1alpha1"
+	farController "github.com/medik8s/fence-agents-remediation/controllers"
+	"github.com/medik8s/fence-agents-remediation/pkg/cli"
+	farUtils "github.com/medik8s/fence-agents-remediation/test/e2e/utils"
+)
+
+const (
+	testNamespace  = "fence-agents-remediation"
+	validNodeName  = "worker-0"
```
mshitrit, line 27: I don't think the node name can't be hard coded, we can't relay on the same name in all clusters

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,171 @@
...
+		var testStart time.Time
+		oldNodes := &corev1.NodeList{}
+		req, _ := labels.NewRequirement(hostNameLabel, selection.Exists, []string{validNodeName})
+		selector := labels.NewSelector().Add(*req)
+		BeforeEach(func() {
+			Expect(k8sClient.List(context.Background(),
+				oldNodes, &client.ListOptions{LabelSelector: selector})).ToNot(HaveOccurred())
+			testNode = &oldNodes.Items[0]
+
+			// save boot time
```
mshitrit, line 67: I think this comment is a bit confusing.
IIUC this is not the Node's boot time, just the time in the beginning of the test

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,171 @@
...
+			testStart = time.Now()
+			far = createFAR(validNodeName, fenceAgentIPMI, testShareParam, testNodeParam)
```
mshitrit, line 69: Nit: I think it could be a nice opportunity to use [DeferCleanup()](https://onsi.github.io/ginkgo/#cleaning-up-our-cleanup-code-defercleanup)

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+		AfterEach(func() {
+			deleteFAR(far)
+		})
+
+		When("running FAR to reboot node ", func() {
+			It("should execute the fence agent cli command", func() {
+				By("checking the CR has been created")
+				farCR := &v1alpha1.FenceAgentsRemediation{}
+				ExpectWithOffset(2*offsetExpect, k8sClient.Get(context.Background(),
```
mshitrit, line 80: >2*offsetExpect

Can you explain why is this offset needed ? 
In any case I think it's confusing to use it like (a fixed number is clearer IMO)

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,171 @@
...
+					client.ObjectKey{Name: validNodeName, Namespace: testNamespace}, farCR)).ToNot(HaveOccurred())
+
+				By("checking the command has been executed successfully")
+				checkFarLogs(testNode, cli.SuccessCommandLog)
+				log.Info("We have found that the command has been executed successfully", "Node name", validNodeName)
+			})
+			It("should reboot node", func() {
```
mshitrit, line 87: I think both `It` can be merged into one

test/e2e/far_e2e_test.go
razo7: Sounds like a good idea.
Since I am using an old Ginkgo version ([1.16.5](https://github.com/medik8s/fence-agents-remediation/blob/main/Makefile#L18)), then I can't use `DeferCleanup()`. I will give it a try on a follow up PR after this PR is merged and I will update all the of FAR tools (including Ginkgo).

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,172 @@
...
+	log.Error(err, "failed to get boot time")
+	return time.Time{}, err
+}
+
+// checkFarLogs get the FAR pod and check whether its logs has logString
+func checkFarLogs(node *corev1.Node, logString string) {
+	By("checking logs")
+	pod, err := getFenceAgentsPod(testNamespace)
+	if err != nil {
+		fmt.Print(err)
```
mshitrit, line 136: probably better to log it instead

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,171 @@
...
+				Eventually(func() (time.Time, error) {
+					bootTime, err := farUtils.GetBootTime(clientSet, validNodeName, testNamespace, log)
+					if bootTime != nil && err == nil {
+						log.Info("got boot time", "time", *bootTime)
+						return *bootTime, nil
+					}
+					log.Error(err, "failed to get boot time")
+					return time.Time{}, err
+				}, timeout, pollInterval).Should(
+					BeTemporally(">", testStart),
```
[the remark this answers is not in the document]
razo7: I agree

test/e2e/far_e2e_test.go
razo7: It is no longer hard codded

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+			far                *v1alpha1.FenceAgentsRemediation
+			cond               corev1.NodeCondition
+			errBoot            error
+		)
+		testNode := &corev1.Node{}
+		nodes := &corev1.NodeList{}
+		BeforeEach(func() {
+			Expect(k8sClient.List(context.Background(), nodes, &client.ListOptions{})).ToNot(HaveOccurred())
+			if len(nodes.Items) <= 1 {
+				Skip("there is one or less available nodes in the cluster")
```
mshitrit, line 64: Are we expecting to test on such clusters (one or less nodes) ? otherwise probably better to fail the test

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+// getKubeletReadyCondition return the  condition of Node kubelet, otherwise return an error
+func getKubeletReadyCondition(nodeName string) (corev1.NodeCondition, error) {
+	emptyCondition := corev1.NodeCondition{}
+	node, err := clientSet.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
+	if err != nil {
+		return emptyCondition, err
+	}
+	for _, condition := range node.Status.Conditions {
+		if condition.Type == "Ready" {
```
mshitrit, line 137: Nit: I suggest replace "Ready" with corev1.NodeReady

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+			if len(nodes.Items) <= 1 {
+				Skip("there is one or less available nodes in the cluster")
+			}
+			//TODO: Randomize the node selection
+			// Use FA on the first node - master-0
+			testNode = &nodes.Items[0]
+			log.Info("Testing Node", "Node name", testNode.Name)
+
+			// save the node's boot time prior to the fence agent call
+			if cond, errBoot = getKubeletReadyCondition(testNode.Name); errBoot != nil {
```
mshitrit, line 72: I don't think that getting the last transition of the Ready condition is necessarily equivalent to the node's boot time.
[Here](https://github.com/medik8s/self-node-remediation/blob/main/e2e/self_node_remediation_test.go#L412) is how we get the boot time in SNR

mshitrit:
I think that as long as we are using hard coded [node names](https://github.com/medik8s/fence-agents-remediation/pull/20/files#diff-b9729ac4458c7637438e3bf558999b35adedcaf4d13af1e3eb752db501fd4997R44-R49) or [ip](https://github.com/medik8s/fence-agents-remediation/pull/20/files#diff-b9729ac4458c7637438e3bf558999b35adedcaf4d13af1e3eb752db501fd4997R39) this test will not be able to run successfully in our CI

test/e2e/far_e2e_test.go
razo7: > I don't think that getting the last transition of the Ready condition is necessarily equivalent to the node's boot time.

What makes you think this way?

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+		AfterEach(func() {
+			deleteFAR(far)
+		})
+
+		When("running FAR to reboot node ", func() {
+			It("should execute the fence agent cli command", func() {
+				By("checking the CR has been created")
+				farCR := &v1alpha1.FenceAgentsRemediation{}
+				ExpectWithOffset(offsetExpect, k8sClient.Get(context.Background(),
```
slintes, line 87: why with offset here?

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+					client.ObjectKey{Name: testNode.Name, Namespace: testNamespace}, farCR)).ToNot(HaveOccurred())
```
slintes, line 88: nit: IMHO a bit nicer: client.ObjectKeyFromObject(farCR)

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+	options := client.ListOptions{
+		LabelSelector: podLabelsSelector,
+		Namespace:     namespace,
+	}
+	if err := k8sClient.List(context.Background(), pods, &options); err != nil {
+		log.Error(err, "can't find the pod by it's labels")
+		return nil
+	}
+	if len(pods.Items) == 0 {
+		podNotFoundErr := &apiErrors.StatusError{ErrStatus: metav1.Status{
```
slintes, line 177: constructing an API error here just for logging it looks wrong to me.
Maybe just use log.error(errors.New("..."), ...)?

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+			Status: metav1.StatusFailure,
+			Code:   http.StatusNotFound,
+			Reason: metav1.StatusReasonNotFound,
+		}}
+		log.Error(podNotFoundErr, "Zero containers for the pod")
```
slintes, line 182: we looked for pods, not containers 🙂

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+		return nil
+	}
+	return &pods.Items[0]
+}
+
+// wasNodeRebooted wait for kublet status to switch from flase to true
+func wasNodeRebooted(nodeName string, lastReadyTime metav1.Time) {
+	kubletPhase := "was ready"
```
slintes, line 190: As posted earlier I don't think we can use the ready condition at all.
However, just as an exercise, I think this code is way too complicated.
Can't the test be reduced to `return cond.Status == corev1.ConditionTrue && cond.LastTransitionTime.After(lastReadyTime.Time)`? 🙂

slintes (changes requested):

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+					wasNodeRebooted(testNode.Name, nodeBootTimeBefore)
+				} else {
+					Skip("we couldn't get the boot time of the node prior to FAR CR, thus we can't verify if it was rebooted")
+				}
+			})
+		})
+	})
+})
+
+// createFAR assign the input to FenceAgentsRemediation object, create CR it with offset, and return the CR object
```
clobrano, line 104: Nit: s/assign/assigns, same for "create" and "return"

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+			return condition, nil
+		}
+	}
+	return emptyCondition, fmt.Errorf("Node %s is not ready", nodeName)
+}
+
+// checkFarLogs get the FAR pod and check whether its logs has logString
+func checkFarLogs(logString string) {
+	By("checking logs")
+	var pod *corev1.Pod
```
clobrano, line 147: There is already a `By` section before calling this function, maybe we can get rid of this one.

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+				By("checking the CR has been created")
+				farCR := &v1alpha1.FenceAgentsRemediation{}
+				ExpectWithOffset(offsetExpect, k8sClient.Get(context.Background(),
+					client.ObjectKey{Name: testNode.Name, Namespace: testNamespace}, farCR)).ToNot(HaveOccurred())
+
+				By("checking the command has been executed successfully")
+				checkFarLogs(cli.SuccessCommandLog)
+
+				By("checking the node's boot time after running the FA")
+				if !nodeBootTimeBefore.IsZero() {
```
clobrano, line 94: I wonder if it would be better to check this value at the beginning of the section, so to avoid all the previous steps

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+			"worker-1": "6234",
+			"worker-2": "6235",
+		},
+	} // get ports
+	Context("fence agent - fence_ipmilan", func() {
+		var (
+			nodeBootTimeBefore metav1.Time
+			far                *v1alpha1.FenceAgentsRemediation
+			cond               corev1.NodeCondition
+			errBoot            error
```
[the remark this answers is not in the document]
razo7: Yes, it is redundant

test/e2e/far_e2e_test.go
razo7: I agree but the comment around `getKubeletReadyCondition`.
If the reboot is faster than waiting around 40s until an unresponsive node is marked as not ready, then why this test will fail? IIUC my test will wait more time which is safer than finish testing too early.
It might means that my test will unnecessary last longer. ATM I don't see this as a problem, but when the E2E tests will be more complicated that testing this might be time consuming :thinking

test/e2e/far_e2e_test.go
clobrano: I agree with that, but IIUC the value of `nodeBootTimeBefore` is set before anything else in `BeforeEach`, and if zero all the test has done is useless.

test/e2e/far_e2e_test.go
slintes: I expected that you test that the transition time to ready status is after start test time. When the node never gets unready that will fail.

test/e2e/far_e2e_test.go
```diff
@@ -21,8 +21,8 @@ import (
 )
 
 const (
-	// testNamespace  = "fence-agents-remediation"
-	testNamespace  = "openshift-operators"
+	testNamespace = "fence-agents-remediation"
+	// testNamespace  = "openshift-operators"
```
clobrano, line 25: Is there a reason to keep this commented code?

razo7: Not really 😄

test/e2e/far_e2e_test.go
razo7: On a second though, I will just test this value in the _BeforeEach_ and fail if it is zero.

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+const (
+	// testNamespace  = "fence-agents-remediation"
+	testNamespace  = "openshift-operators"
+	fenceAgentIPMI = "fence_ipmilan"
+
+	// eventually parameters
+	timeout      = 2 * time.Minute
+	pollInterval = 10 * time.Second
+	offsetExpect = 1
```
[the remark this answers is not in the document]
slintes: I can't follow. What's the improvement here? Why would writing `offsetOfOneForExpect` instead of `1` be better? 🤔 

When we go further down that road we end up with
```
const (
    ONE = 1
    TWO = 2
    THREE = 3

    offsetOfOneForExpect = ONE
)

func add(a, b int) (int, error) {
    if a == ONE && b == 2 {
       return THREE, nil
    }
    return 0, fmt.Errorf("running out of  \"MACROS\" for integers!")
}
```
😉

test/e2e/far_e2e_test.go
razo7: > When the node never gets unready that will fail.

Is that bad? I think it works as I would expect. 
Is there a scenario where a node has been remediated and in the process of remediation the node never gets to _Not Ready_ status?

slintes: > Is that bad? I think it works as I would expect.

You want to know if FAR works, right? You want to do that by checking that the Node rebooted, right? You want to do that by looking at the Ready condition, right? So when that condition never changes, how do you that the node rebooted and far did something?

> Is there a scenario where a node has been remediated and in the process of remediation the node never gets to Not Ready status?

When remediation is faster than 40s: yes

razo7: > When remediation is faster than 40s: yes

Now, I got it. Thanks 👍🏻

razo7: I am reverting to verify reboot by looking for boot time rather than using the getKubeletReadyCondition

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,185 @@
...
+	Context("fence agent - fence_ipmilan", func() {
+		var (
+			far                *v1alpha1.FenceAgentsRemediation
+			nodeBootTimeBefore time.Time
+			errBoot            error
+			testNodeName       string
+		)
+		nodes := &corev1.NodeList{}
+		BeforeEach(func() {
+			Expect(k8sClient.List(context.Background(), nodes, &client.ListOptions{})).ToNot(HaveOccurred())
```
slintes, line 59: leaving this up to you, but I think running remediation tests on worker nodes is more stable in CI...

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,185 @@
...
+			if len(nodes.Items) <= 1 {
+				Fail("there is one or less available nodes in the cluster")
+			}
+			//TODO: Randomize the node selection
+			// Use FA on the first node - master-0
```
slintes, line 64: "master-0" is cluster specific

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,185 @@
...
+			nodeObj := nodes.Items[0]
+			testNodeName = nodeObj.Name
+			log.Info("Testing Node", "Node name", testNodeName)
+
+			// save the node's boot time prior to the fence agent call
+			nodeBootTimeBefore, errBoot = getNodeBootTime(testNodeName)
+			if errBoot != nil {
+				Fail("Can't get boot time of the node")
```
slintes, line 72: nit, a bit nicer: `Expect(errBoot).ToNot(HaveOccurred(), "failed to get boot time of the node")`

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,185 @@
...
+		var errBootAfter error
+		nodeBootTimeAfter, errBootAfter = getNodeBootTime(nodeName)
+		if errBootAfter != nil {
+			log.Error(errBootAfter, "Can't get boot time of the node")
+		}
+		return nodeBootTimeAfter, errBootAfter
+	}, timeoutReboot, pollInterval).Should(
+		BeTemporally(">", nodeBootTimeBefore), "The node didn't finish a reboot after FAR CR has been created and timeout time has passed")
+
+	log.Info("successful reboot", "node", nodeName, "offset beteween boot times", nodeBootTimeAfter.Sub(nodeBootTimeBefore), "new boot time", nodeBootTimeAfter)
```
slintes, line 146: typo "beteween"

slintes (commented):
looks much better now, some nits inside

test/e2e/far_e2e_test.go
razo7: No longer true after #42

pkg/cli/cliexecuter.go
```diff
@@ -24,6 +24,11 @@ type executer struct {
 	clientSet *kubernetes.Clientset
 }
 
+const (
+	failCommandLog    = "Command failed to be executed"
+	SuccessCommandLog = "Command has been executed successfully"
+)
```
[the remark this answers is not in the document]
razo7: I do no longer apply this kind of change..

test/e2e/far_e2e_test.go
razo7: This PR takes longer time than expected, and Ginkgo v2 was already updated - #50

[DeferCleanup()](https://onsi.github.io/ginkgo/#cleaning-up-our-cleanup-code-defercleanup) was added to E2E test.

config/ocp_aws/install/fence_aws_credentials_request.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - kms:GenerateDataKey
+      - kms:GenerateDataKeyWithoutPlainText
+      - kms:DescribeKey
+      effect: Allow
+      resource: '*'
+    - action:
+      - kms:RevokeGrant
+      - kms:CreateGrant
+      - kms:ListGrants
+      effect: Allow
```
beekhof, line 41: Do we need these?

razo7: That's a very good question... 

I have begun with the credentials of [`openshift-machine-api-aws` CredentialRequest](https://github.com/openshift/machine-api-operator/blob/931f6f67c1c72597fc56104cd45a71c11f19293d/install/0000_30_machine-api-operator_00_credentials-request.yaml#L20-L61), plus  `ec2:StartInstances`, and `ec2:StopInstances`, and try to minimize the amount of actions. There might be more actions to remove, but this is where I have stopped.

I do remember that I tried only with `ec2` and `iam` actions, and the reboot failed (due to timeout of waiting for the node to turn off, probably not a permission problem...) without `kms` actions.
But with `kms` actions it was always working. Not on CI, but on Cluster Bot.

P.S. I tried using  the [power_timeout](https://manpages.ubuntu.com/manpages/focal/man8/fence_aws.8.html) flag to extend the timeout, but it didn't work as IIUC the [**fence_aws** FA](https://github.com/ClusterLabs/fence-agents/blob/main/agents/aws/fence_aws.py) doesn't use this flag (and only set it to [60 seconds by default](https://github.com/ClusterLabs/fence-agents/blob/main/agents/aws/fence_aws.py#L182)).

config/ocp_aws/install/fence_aws_credentials_request.yaml
```diff
@@ -0,0 +1,45 @@
+# similar to https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_00_credentials-request.yaml#L1-L6
+apiVersion: cloudcredential.openshift.io/v1
+kind: CredentialsRequest
```
slintes, line 3: I'm afraid this won't work, it won't be installed. OLM supports a limited set of resource types only. See switch cases at [1] and few CRD types at [2].

[1] : https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator.go#L2070
[2]: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/supportedresources.go#L15

razo7: Do you mean that the [`openshift-aws-cloud-fencing`](https://github.com/medik8s/fence-agents-remediation/pull/20/files/3cc361ad0c09cc88329d5a22856a96f0f2327d03#diff-6c27ee83cb33364a46c46291cde6b29cd14061ee251d05d7c7d8884d8390e9f6R5) CR won't be created on OpenShift CI?
Looking on the logs from E2E it looks like it was indeed created ([lines 103-107](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/medik8s_fence-agents-remediation/20/pull-ci-medik8s-fence-agents-remediation-main-4.12-openshift-e2e/1666095122816700416#1:build-log.txt%3A107)), but I would wait for a [fixed `fence_aws`](https://bugzilla.redhat.com/show_bug.cgi?id=2183158) FA to verify my understanding in practice :D 

>Downloading  mirror.openshift.com/pub/openshift-v4/clients/ocp/4.13.0/openshift-client-linux-4.13.0.tar.gz
mkdir -p /go/src/github.com/medik8s/fence-agents-remediation/bin/oc/4.13.0/oc-extracted 2>/dev/null || true
tar -xzf /go/src/github.com/medik8s/fence-agents-remediation/bin/oc/4.13.0/oc-tar -C /go/src/github.com/medik8s/fence-agents-remediation/bin/oc/4.13.0/oc-extracted
/go/src/github.com/medik8s/fence-agents-remediation/bin/kustomize/v4@v4.5.7/kustomize build config/ocp_aws | /go/src/github.com/medik8s/fence-agents-remediation/bin/oc/4.13.0/oc-extracted/oc apply -f -
credentialsrequest.cloudcredential.openshift.io/openshift-aws-cloud-fencing created

slintes: oh ok, you create the CredentialsRequest yourself. I thought you want to put it into the bundle and let OLM install it. All good then 🙂

test/e2e/far_e2e_test.go
```diff
@@ -49,25 +52,34 @@ var _ = Describe("FAR E2e", func() {
...
 			nodes := &corev1.NodeList{}
 			Expect(k8sClient.List(context.Background(), nodes, &client.ListOptions{})).ToNot(HaveOccurred())
 			if len(nodes.Items) <= 1 {
 				Fail("there is one or less available nodes in the cluster")
 			}
-			//TODO: Randomize the node selection
-			// run FA on the first node - a master node
+			//TODO: Randomize the node selection & verify valid index
+			// run FA on the fourth node - a worker node
 			nodeObj := nodes.Items[nodeIndex]
```
mshitrit, line 71: Assuming the 4th node would be a worker node (or even that there will be at least 4 nodes) seems a bit risky.
I suggest either  modify the initial fetching of nodes to retrieve worker nodes  (on `k8sClient.List(...  &client.ListOptions{}..` or just filter out the results

razo7: It is small and important change :D

test/e2e/utils/cluster.go
```diff
@@ -34,11 +36,11 @@ func GetClusterInfo(config configclient.Interface) (*configv1.Infrastructure, er
 }
 
 // GetCredentials searches for AWS or BMH secret, and then returns it decoded
-func GetCredentials(clientSet *kubernetes.Clientset, secretName, secretKey, secretVal string) (string, string, error) {
+func GetCredentials(clientSet *kubernetes.Clientset, secretName, secretNamespace, secretKey, secretVal string) (string, string, error) {
```
slintes, line 39: misleading function name and description: this is not specific to AWS or BMH, it's just getting a secret's key and value based on function parameters

test/e2e/far_e2e_test.go
```diff
@@ -131,22 +155,25 @@ func deleteFAR(far *v1alpha1.FenceAgentsRemediation) {
 func buildSharedParameters(clusterPlatform *configv1.Infrastructure, action string) (map[v1alpha1.ParameterName]string, error) {
 	const (
 		//AWS
-		secretAWS    = "aws-cloud-credentials"
-		secretKeyAWS = "aws_access_key_id"
-		secretValAWS = "aws_secret_access_key"
+		// secretDefaultAWS    = "aws-cloud-credentials"
```
slintes, line 158: can this be removed?

slintes (commented):
nice work! Overall looks good to me, just nits

test/e2e/far_e2e_test.go
razo7: Sure, it is just a small leftover :)

slintes:
/hold

checking the actual logs...

pkg/utils/nodes.go
```diff
@@ -8,6 +8,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const WorkerLabelName = "node-role.kubernetes.io/worker"
```
mshitrit, line 11: Nit: might want to add this to common repo, IIRC it is also used in SNR & NHC

mshitrit:
/lgtm
/hold
Giving others final chance to review as well

pkg/utils/pods.go
```diff
@@ -10,8 +10,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// GetFenceAgentsRemediationPod fetches the FAR pod based on FAR's label and namespace
+// GetFenceAgentsRemediationPod fetches the first pod that match to FAR's label and namespace
```
clobrano, line 13: typo s/match/matches

clobrano (approved):
A couple of nits
/lgtm
giving chance to the other to give feedback
/hold
feel free to unhold

pkg/utils/pods.go
razo7: In a follow up :)

test/e2e/utils/cluster.go
```diff
@@ -33,16 +35,16 @@ func GetClusterInfo(config configclient.Interface) (*configv1.Infrastructure, er
 	return clusterInfra, nil
 }
 
-// GetCredentials searches for AWS or BMH secret, and then returns it decoded
-func GetCredentials(clientSet *kubernetes.Clientset, secretName, secretKey, secretVal string) (string, string, error) {
+// GetSecretData searches for the platform's secret, and then returns it's decoded two data values
```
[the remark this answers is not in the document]
razo7: That's not always the case, but I can express some example here :)
In a follow up :)

pkg/utils/nodes.go
slintes: I'm wondering if this is somewhere in client-go / k8s api server code...

slintes:
/hold

when someone requested significant changes or we discussed important issues with the PR, please wait until he had a chance to give another review...

pkg/utils/pods.go
```diff
@@ -10,8 +10,15 @@ import (
...
 func GetFenceAgentsRemediationPod(r client.Reader) (*corev1.Pod, error) {
+	podlsit, err := GetFARPodList(r)
```
slintes, line 15: typo: podList

pkg/utils/pods.go
```diff
@@ -10,8 +10,15 @@ import (
...
+	return &podlsit.Items[0], err
```
slintes, line 16: what about returning the pod which is in `Running` phase?

test/e2e/far_e2e_test.go
```diff
@@ -197,39 +224,86 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
...
 }
 
-// checkFarLogs gets the FAR pod and checks whether it's logs have logString
-func checkFarLogs(logString string) {
-	var pod *corev1.Pod
-	var err error
-	EventuallyWithOffset(1, func() *corev1.Pod {
-		pod, err = utils.GetFenceAgentsRemediationPod(k8sClient)
+// getNodeBootTime returns the bootime of node nodeName if possible, otherwise it returns an error
+func getNodeBootTime(nodeName string) (time.Time, error) {
```
slintes, line 242: what's the value of this, why not using `e2eUtils.GetBootTime()` directly :confused:

test/e2e/far_e2e_test.go
```diff
@@ -197,39 +224,86 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
...
+	bootTime, err := e2eUtils.GetBootTime(clientSet, nodeName, testNsName, log)
+	if bootTime != nil && err == nil {
+		return *bootTime, nil
+	}
+	return time.Time{}, err
+}
+
+// getFarPod gets a podList with all the pods that match by FAR's label
+func getFarPod() *corev1.PodList {
```
slintes, line 251: misleading func name and var name below, should be either pods or podList

test/e2e/far_e2e_test.go
```diff
@@ -197,39 +224,86 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
...
+	var (
+		pod *corev1.PodList
+		err error
+	)
+	EventuallyWithOffset(1, func() *corev1.PodList {
+		pod, err = utils.GetFARPodList(k8sClient)
```
slintes, line 257: why do you use the list here, wouldn't GetFenceAgentsRemediationPod() be the better choice? Do we need this function at all? :confused:

test/e2e/far_e2e_test.go
```diff
@@ -197,39 +224,86 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
...
+	var (
+		logs string
+		err  error
+	)
 	EventuallyWithOffset(1, func() string {
-		logs, err := e2eUtils.GetLogs(clientSet, pod, containerName)
-		if err != nil {
-			log.Error(err, "failed to get logs. Might try again")
-			return ""
+		podList := getFarPod()
```
slintes, line 274: what about using GetFenceAgentsRemediationPod() directly here?

slintes (changes requested):

pkg/utils/nodes.go
razo7: I didn't find it :)

pkg/utils/pods.go
razo7: Sounds better :)

test/e2e/far_e2e_test.go
razo7: It is quite old PR and I can't recall why it was needed. Anyway I have removed as I see your point 👍🏻

test/e2e/far_e2e_test.go
razo7: No, it is redundant with the new update to GetFenceAgentsRemediationPod() function

pkg/utils/nodes.go
mshitrit: I already have an open PR in common since it's a very small change I thought it would be ok to added that label there as well: https://github.com/medik8s/common/pull/3

pkg/utils/pods.go
```diff
@@ -10,23 +10,33 @@ import (
...
 
-// GetFenceAgentsRemediationPod fetches the FAR pod based on FAR's label and namespace
+// GetFenceAgentsRemediationPod fetches the first running pod that matches to FAR's label and namespace
+// The pod should be on running state since when we taint and reboot a node which had FAR, then that pod will be restarted on a different node,
+// This results with an old pod which is about to die and new pod is already running and we want to return the running pod
 func GetFenceAgentsRemediationPod(r client.Reader) (*corev1.Pod, error) {
-	var podNamespace string
-	pods := &corev1.PodList{}
+	var (
+		podNamespace string
```
slintes, line 18: unused (overwritten by `:=` )

pkg/utils/pods.go
```diff
@@ -10,23 +10,33 @@ import (
...
+		pod          *corev1.Pod
```
slintes, line 19: unused (will always be nil at the return in the last line)

pkg/utils/pods.go
```diff
@@ -10,23 +10,33 @@ import (
...
+	err = r.List(context.Background(), podList, &client.ListOptions{LabelSelector: selector, Namespace: podNamespace})
 	if err != nil {
 		return nil, fmt.Errorf("failed fetching FAR pod - %w", err)
 	}
-	if len(pods.Items) == 0 {
+	if len(podList.Items) == 0 {
 		return nil, fmt.Errorf("no Fence Agent pods were found")
 	}
-	return &pods.Items[0], nil
+	for _, pod := range podList.Items {
```
slintes, line 36: please use https://github.com/razo7/fence-agents-remediation/blob/78507c8120631d24803cca00af54922029d3ca38/vendor/k8s.io/api/core/v1/types.go#L2624

test/e2e/far_e2e_test.go
```diff
@@ -197,39 +223,61 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
...
 	return testNodeParam, nil
 }
 
 // checkFarLogs gets the FAR pod and checks whether it's logs have logString
 func checkFarLogs(logString string) {
-	var pod *corev1.Pod
-	var err error
-	EventuallyWithOffset(1, func() *corev1.Pod {
+	var (
+		logs string
```
slintes, line 243: nit: what's the value of declaring these vars here? I would just use `:=` in the Eventually block.

slintes (changes requested):

PR #28, opened 2023-03-23
Add bundle-run Makefile target

clobrano:
Add Makefile's target to run bundle via operator-sdk.

Signed-off-by: Carlo Lobrano <c.lobrano@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,6 +180,11 @@ test: test-no-verify verify-unchanged ## Generate and format code, run tests, ge
 test-no-verify: manifests generate go-verify fmt vet envtest ginkgo # Generate and format code, and run tests
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(ENVTEST_DIR)/$(ENVTEST_VERSION) -p path)"  $(GINKGO) -v -r --keepGoing -requireSuite -coverprofile cover.out
 
+.PHONY: bundle-run
+export BUNDLE_RUN_NAMESPACE ?= openshift-operators
+bundle-run: operator-sdk ## Run bundle image. Default NS is "openshift-operators", redefine BUNDLE_RUN_NAMESPACE to override it.
+	$(OPERATOR_SDK) -n $(BUNDLE_RUN_NAMESPACE) run bundle $(BUNDLE_IMG)
+
 ##@ Build
 
 .PHONY: build


mshitrit:
Do we plan to use this for anything other than to deploy operators on our dev clusters ?

clobrano:
> Do we plan to use this for anything other than to deploy operators on our dev clusters ?

And maybe to simplify testing for others

mshitrit:
> maybe to simplify testing for others

I am not sure that code that is intended purely for dev purposes belongs here, my first intuition is that this would belong in a more internal space.

clobrano:
> I am not sure that code that is intended purely for dev purposes belongs here, my first intuition is that this would belong in a more internal space.

That's a good point and I don't have strong opinion about it

mshitrit:
> > I am not sure that code that is intended purely for dev purposes belongs here, my first intuition is that this would belong in a more internal space.
> 
> That's a good point and I don't have strong opinion about it

@slintes , @razo7 WDYT, do you have any preference ?

mshitrit:
> What's the difference to all the other targets, isn't the complete Makefile for devs? thinking

I guess that depends on how you define `dev` , from my perspective most of the targets have to do with testing/deploying the operator which are necessary to the CI.
IMO this different than a  target which is used just to deploy on my local dev cluster.

Definitely not something I feel strongly about though, and IIUC other objections, so we can probably merge ahead.

Makefile
```diff
@@ -317,6 +317,11 @@ bundle-build: bundle bundle-update ## Build the bundle image.
 bundle-push: ## Push the bundle image.
 	$(MAKE) docker-push IMG=$(BUNDLE_IMG)
 
+# Run bundle image
+.PHONY: bundle-run
+bundle-run: operator-sdk
+	$(OPERATOR_SDK) -n openshift-operators run bundle $(BUNDLE_IMG)
```
razo7, line 323: > > > I am not sure that code that is intended purely for dev purposes belongs here, my first intuition is that this would belong in a more internal space.
> > 
> > 
> > That's a good point and I don't have strong opinion about it
> 
> @slintes , @razo7 WDYT, do you have any preference ?

I join some of us who I believe agree that this is kind of gray area, if this target is needed or in the right place. I don't mind having it. Maybe just the the change the bundle name to include something about that specific namespace, `-n openshift-operators` or even remove it from the target and set is as a variable that we can export?

clobrano: A variable we can export seems a good improvement

slintes:
> > What's the difference to all the other targets, isn't the complete Makefile for devs? thinking
> 
> I guess that depends on how you define `dev` , from my perspective most of the targets have to do with testing/deploying the operator which are necessary to the CI. IMO this different than a target which is used just to deploy on my local dev cluster.

ok, let me rephrase 😉 
"IMHO the Makefile helps developers and CI to build and deploy the operator" 😎 

> A variable we can export seems a good improvement

👍🏼 (but `openshift-operators` is a good default value IMHO, as long as most people use it on OpenShift)

> but would the "Development" section fit better?

I have no preferance 🤷🏼‍♂️

razo7:
> I added this target to the "Bundle" section just because it refers to bundle, but would the "Development" section fit better?
https://github.com/medik8s/fence-agents-remediation/blob/main/Makefile#L138

I don't mind to use the current location

PR #29, opened 2023-03-23
Dummy e2e Test

razo7:
Basic and dummy E2E test which only creates CR, verify it was created and then deletes it. The FAR CR is not running any FA, and only echo an empty string -> `echo " "`
This way we can catch installation/running errors, such as #26, because deployment would fail until we add functionality of E2E test in #20.

### DIFF
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -14,7 +14,7 @@ import (
 	"github.com/medik8s/fence-agents-remediation/api/v1alpha1"
 )
 
-const fenceAgentDummy = "echo"
+const fenceAgentDummyName = "echo"
 
 var _ = Describe("FAR E2e", func() {
 	Context("fence agent - dummy", func() {
@@ -24,7 +24,7 @@ var _ = Describe("FAR E2e", func() {
 		BeforeEach(func() {
 			testShareParam := map[v1alpha1.ParameterName]string{}
 			testNodeParam := map[v1alpha1.ParameterName]map[v1alpha1.NodeName]string{}
-			far = createFAR(testNodeName, fenceAgentDummy, testShareParam, testNodeParam)
+			far = createFAR(testNodeName, fenceAgentDummyName, testShareParam, testNodeParam)
 		})
 
 		AfterEach(func() {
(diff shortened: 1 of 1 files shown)


test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,64 @@
...
+	. "github.com/onsi/gomega"
+
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/medik8s/fence-agents-remediation/api/v1alpha1"
+)
+
+const fenceAgentDummy = "echo"
```
clobrano, line 17: (very) nit: I think the name of this variable could be more descriptive. Is it the Agent name?

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,64 @@
...
+		})
+
+		It("should check whether the CR has been created", func() {
+			testFarCR := &v1alpha1.FenceAgentsRemediation{}
+			Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(far), testFarCR)).To(Succeed(), "failed to get FAR CR")
+		})
+	})
+})
+
+// createFAR assigns the input to FenceAgentsRemediation object, creates CR it with offset, and returns the CR object
```
clobrano, line 41: is there a maybe a typo here? "creates CR it with offset"

test/e2e/far_e2e_test.go
```diff
@@ -0,0 +1,64 @@
...
+func createFAR(nodeName string, agent string, sharedparameters map[v1alpha1.ParameterName]string, nodeparameters map[v1alpha1.ParameterName]map[v1alpha1.NodeName]string) *v1alpha1.FenceAgentsRemediation {
```
clobrano, line 42: Use camel case naming format might be helpful
nodeParameters, sharedParameters

clobrano (changes requested):

test/e2e/far_e2e_test.go
mshitrit: Nit: IIUC the offset refers to the error reporting and not the CR creation

test/e2e/far_e2e_test.go
mshitrit: Nit: IMO var value could be better (i.e  for example dummy-fence-agent)

mshitrit:
/lgtm
/hold 
left 2 nits, feel free to fix/unhold

test/e2e/far_e2e_test.go
razo7: `echo` is just an example to `dummy-fence-agent`  and I doubt would run without a failure. In the end a pod will try to run a cli command `dummy-fence-agent`, and I have used `echo` since it was most likely to be executed on any pod (and Linux based OS).

test/e2e/far_e2e_test.go
razo7: Correct I will fix it in a follow up

PR #42, opened 2023-05-01
Use Downward API for Namespace

razo7:
[Downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/) simplifies the code by getting the pod's namespace without hard code the value. It stores the pod's namespace in environment variable - `DEPLOYMENT_NAMESPACE` as it [was suggested](https://github.com/medik8s/fence-agents-remediation/pull/16#discussion_r1113279956) earlier in #16.

Fix [ECOPROJECT-1234](https://issues.redhat.com//browse/ECOPROJECT-1234)

### DIFF
--- a/pkg/utils/namespaces.go
+++ b/pkg/utils/namespaces.go
@@ -1,0 +1,20 @@
+package utils
+
+import (
+	"fmt"
+	"os"
+)
+
+// GetDeploymentNamespace returns the Namespace this operator is deployed on.
+func GetDeploymentNamespace() (string, error) {
+	// deployNamespaceEnvVar is the constant for env variable DEPLOYMENT_NAMESPACE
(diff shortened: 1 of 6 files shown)


slintes (changes requested):

pkg/utils/namespaces.go
```diff
@@ -0,0 +1,21 @@
+package utils
+
+import (
+	"fmt"
+	"os"
+)
+
+// deployNamespaceEnv is a constant for env variable DEPLOYMENT_NAMESPACE
+// which specifies the Namespace to watch.
```
slintes, line 9: missed it in the ealier review: this isn't the namespace to watch (it's watching all namespaces)

PR #46, opened 2023-05-08
Use Kubernetes Recommended Label for commonLabels

razo7:
Changing FAR's label from ` app: fence-agents-remediation-operator` to `app.kubernetes.io/name: fence-agents-remediation-operator`, since it is one of [Kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) for commonLabels.

### DIFF
--- a/bundle/manifests/fence-agents-remediation-controller-manager-metrics-service_v1_service.yaml
+++ b/bundle/manifests/fence-agents-remediation-controller-manager-metrics-service_v1_service.yaml
@@ -3,7 +3,7 @@ kind: Service
 metadata:
   creationTimestamp: null
   labels:
-    app: fence-agents-remediation-operator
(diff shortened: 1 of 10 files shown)


razo7:
Another commonLabel that we might also consider of using is `app.kubernetes.io/version`.

clobrano:
/lgtm

Giving a chance to get more feedbacks, feel free to unhold
/hold

beekhof:
Merging.  Updates from a pre-GA version that lacks resource deletion is not a priority

slintes:
Just for the record, the update issue with modified labels (if it still exists in OLM) can easily be worked around by changing the deployment name. Just in case we want to make Tech Preview users happy 🙂

PR #48, opened 2023-05-10
Validate Node Name

razo7:
- Validate FAR CR name to a node name in the cluster.
- Add three new verbs to check whether a specific node name match the nodes' names in the cluster.
- Add and update unit-tests


pkg/utils/nodes.go
```diff
@@ -0,0 +1,49 @@
...
+
+var logger = ctrl.Log.WithName("utils-nodes")
+
+// similar to getNodes - https://github.com/medik8s/node-maintenance-operator/blob/main/test/e2e/node_maintenance_test.go#L258-L282
+// getNodeNames returns a list of node names or an error it can't get the node names
+func getNodeNames(r client.Reader) ([]string, error) {
+	nodes := make([]string, 0)
+	nodesList := &corev1.NodeList{}
+	err := r.List(context.TODO(), nodesList, &client.ListOptions{})
+	//wait one second
```
slintes, line 21: why this?

slintes (commented):
I noticed the WIP, however a quick first review 😉

controllers/fenceagentsremediation_controller.go
```diff
@@ -50,6 +50,7 @@ func (r *FenceAgentsRemediationReconciler) SetupWithManager(mgr ctrl.Manager) er
 
 //+kubebuilder:rbac:groups=core,resources=pods/exec,verbs=create
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;update;delete;deletecollection
+//+kubebuilder:rbac:groups="",resources=nodes,verbs=list;watch
```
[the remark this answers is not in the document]
razo7: I tried with and without `get` and saw that it is working, thus I though of adding less verbs for rbac.

pkg/utils/nodes.go
razo7: Simply redundant

pkg/utils/nodes.go
```diff
@@ -0,0 +1,49 @@
...
+	return nodes, nil
+}
+
+// isNodeNameValid returns an error if nodeName doesn't match the node names from getNodeNames, otherwise a nil
+func IsNodeNameValid(r client.Reader, nodeName string) error {
+	clusterNodes, err := getNodeNames(r)
+	if err != nil {
+		return err
+	}
+	for _, node := range clusterNodes {
```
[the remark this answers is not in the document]
razo7: Have one function that go over the list and search for a specific given name or to instead of loop over the list of node names I should search differently?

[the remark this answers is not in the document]
slintes: No. Do not get a node list at all. Just do a `get` for a single node with the given name. When it exists -> valid, if not -> not valid 🙂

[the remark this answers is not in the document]
razo7: More reasons for getting the `get` the verb ;)

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
slintes: > I though of adding less verbs for rbac

And in general that's a good approach!
But a `list` without `get` doesn't make sense, you have access to all resources already.
Thinking more about it: the other way around it can make sense though: a `get` without `list` / `watch`. That's a difference IMHO. Even more when you also restrict the `get` with a resource name (e.g. you are only allowed to read a secret with a specific name)

pkg/utils/nodes.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// isNodeNameValid returns an error if nodeName doesn't match any node name int the cluster, otherwise a nil
+func IsNodeNameValid(r client.Reader, nodeName string) error {
+	logger := ctrl.Log.WithName("utils-nodes")
+	node := &corev1.Node{}
+	objNodeName := types.NamespacedName{Name: nodeName}
+	err := r.Get(context.TODO(), objNodeName, node)
+	if err != nil {
+		errMsg := fmt.Sprintf("node %s is not part of the cluster's node names", nodeName)
```
slintes, line 20: You need to check the error type. Only NotFound means that the node doesn't exist. There can be other temporary errors.

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
razo7: Having the three verbs is working, so I tried to minimize it:

1. `get;watch` failed and couldn't move on in the reconcile (to run the FA)
`W0511 11:22:40.347954 1 reflector.go:324] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:250: failed to list *v1.Node: nodes is forbidden: User "system:serviceaccount:openshift-operators:fence-agents-remediation-controller-manager" cannot list resource "nodes" in API group "" at the cluster scope
20`
`E0511 11:22:40.347984 1 reflector.go:138] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:250: Failed to watch *v1.Node: failed to list *v1.Node: nodes is forbidden: User "system:serviceaccount:openshift-operators:fence-agents-remediation-controller-manager" cannot list resource "nodes" in API group "" at the cluster scope`

2. `get;list` failed but I could move on in the reconcile (to run the FA) 
`E0511 11:44:07.909025 1 reflector.go:138] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:250: Failed to watch *v1.Node: unknown (get nodes)`

To conclude, I tend to just stick with `get;list;watch` verbs.

pkg/utils/nodes.go
```diff
@@ -0,0 +1,24 @@
...
+// isNodeNameValid returns an error if nodeName doesn't match any node name int the cluster, otherwise a nil
+func IsNodeNameValid(r client.Reader, nodeName string) error {
+	node := &corev1.Node{}
+	objNodeName := types.NamespacedName{Name: nodeName}
+	err := r.Get(context.Background(), objNodeName, node)
+	if err != nil {
+		// Build node not found error
+		msg := fmt.Sprintf("Failed: node %s is not part of the cluster's node names", nodeName)
+		err = buildFailureApiMessage(err, msg)
+		return fmt.Errorf("CR's name doesn't match any node name in the cluster - %w", err)
```
slintes, line 21: 1. Logging is weird:

- 2 lines above: "node %s is not part of the cluster's node names" -> not true, you just failed to get it for SOME reason
- here: "CR's name doesn't match any node name in the cluster" -> yes, but maybe just because you failed to get the node for a reason which isn't notFound

2. Don't you need to differentiate error handling based on the error type? I think you have 3 cases here:
- node found  = name valid
- node not found = name invalid
- any other error = retry

Does that make sense?

With that in mind, the func signature might need an update as well (atm the `is` prefix suggests that the return value is a boolean, bit it's an error)

pkg/utils/pods.go
```diff
@@ -31,13 +29,9 @@ func GetFenceAgentsRemediationPod(r client.Reader) (*corev1.Pod, error) {
...
 	if len(pods.Items) == 0 {
-		podNotFoundErr := &apiErrors.StatusError{ErrStatus: metav1.Status{
-			Status: metav1.StatusFailure,
-			Code:   http.StatusNotFound,
-			Reason: metav1.StatusReasonNotFound,
-		}}
-		logger.Error(podNotFoundErr, "No Fence Agent pods were found")
-		return nil, podNotFoundErr
+		// Build pod not found error
+		err = buildFailureApiMessage(err, "Failed: Not found a pod that match the label")
```
slintes, line 33: `err` is always nil here, what is `buildFailureApiMessage` expected to do?

slintes (changes requested):

pkg/utils/error.go
```diff
@@ -0,0 +1,37 @@
...
+	"fmt"
+	"net/http"
+
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// apiError
+
+func buildFailureApiMessage(err error, msg string) error {
```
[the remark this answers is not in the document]
razo7: > IMHO it's not a good fit here

Do you have in mind a better choice?

pkg/utils/nodes.go
razo7: I agree on 1and I hope now it is better.

> Don't you need to differentiate error handling based on the error type?
I am trying to do so using `buildApiError` function

>any other error = retry

I am not fully understand why to retry getting the node?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -84,6 +86,25 @@ var _ = Describe("FAR Controller", func() {
...
+				Expect(k8sClient.Create(context.Background(), node)).NotTo(HaveOccurred())
+			})
+			AfterEach(func() {
+				Expect(k8sClient.Delete(context.Background(), node)).NotTo(HaveOccurred())
+			})
+			When("FAR's name doesn't match an existing node name", func() {
+				It("should fail", func() {
+					Expect(farUtils.IsNodeNameValid(k8sClient, dummyNodeName)).To(HaveOccurred())
+				})
+				It("should succeed", func() {
```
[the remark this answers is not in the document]
razo7: Can you please elaborate? Why don't match?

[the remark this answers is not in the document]
slintes: you have 1 `When` with 2 `It`s, and when you read them it doesn't make sense... either it should fail or succeed, both doesn't make sense 🤷🏼 
- When FAR's name doesn't match an existing node name It should fail
- When FAR's name doesn't match an existing node name It should succeed

pkg/utils/nodes.go
slintes: > > Don't you need to differentiate error handling based on the error type?

>  I am trying to do so using `buildApiError` function
 
I don't see that? 🤔 You just return another error, no matter which error occurred in first place

> > any other error = retry
 
> I am not fully understand why to retry getting the node?

Because NotFound is the only error which says: there is no node with that name.
Every other error is: I had an issue while getting the node.

razo7: > You just return another error, no matter which error occurred in first place

I create new error with apiErrors based on the given error type (see the use of [IsNotFound](https://github.com/medik8s/fence-agents-remediation/pull/48/files#diff-cf316a9ab7edc81ea1ea26252702d68f8286e9693eb9a284073535a9b0e21301R15), and [IsForbidden](https://github.com/medik8s/fence-agents-remediation/pull/48/files#diff-cf316a9ab7edc81ea1ea26252702d68f8286e9693eb9a284073535a9b0e21301R23) functions). 
There might be a better way of doing that by updating the current error, but the **_[buildApiError](https://github.com/medik8s/fence-agents-remediation/pull/48/files#diff-cf316a9ab7edc81ea1ea26252702d68f8286e9693eb9a284073535a9b0e21301R12)_** function does consider the input error type.

razo7: > Every other error is: I had an issue while getting the node.

Agreed, I just don't understand why we need to retry getting the node. It sounds like you are suggesting to perform a recursion of calling CheckNodeName if it gets an error, when it tries to find a node with a matching `nodeName`.

ATM if [err](https://github.com/medik8s/fence-agents-remediation/pull/48/files#diff-44072254fdc9157bd09396bfdafca285a425236352804c1252445a87e47716ebR17) is not of type _notFound_ and not _forbidden_, then the buildApiError will use the [default error](https://github.com/medik8s/fence-agents-remediation/pull/48/files#diff-cf316a9ab7edc81ea1ea26252702d68f8286e9693eb9a284073535a9b0e21301R33) for the "retry" case you presented.

controllers/fenceagentsremediation_controller_test.go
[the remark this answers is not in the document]
razo7: oho, now I get it. 
I just followed a similar pattern from the [previous UTs](https://github.com/medik8s/fence-agents-remediation/pull/48/files#diff-fac0cf56dda42683c1e24b46c87cfa4cca46c444145ee69de9e264db50f10289R76-R87) with `should fail` and `should succeed`, and with one **when**.
I thought they were ok, thus I didn't understand you. I will update them as well.

pkg/utils/error.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"fmt"
+	"net/http"
+
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// buildApiError returns new error with updated status
```
clobrano, line 11: Could you make it clearer? I couldn't tell what the function does only reading this header

pkg/utils/nodes.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CheckNodeName returns an error if nodeName doesn't match any node name int the cluster, otherwise a nil
```
clobrano, line 12: typo: s/int/in

pkg/utils/nodes.go
slintes: > I just don't understand why we need to retry getting the node.

Because you don't know if the node exists or not. So you don't know if the CR is valid or not.

> It sounds like you are suggesting to perform a recursion of calling CheckNodeName if it gets an error,

No, not recursion. Stop reconcile by returning the error, which results in requeue with backoff.

razo7: > Does this mean that, in Reconcile, you return with an error either way?

Correct

> Wouldn't be okay to requeue, instead?

> Stop reconcile by returning the error, which results in requeue with backoff.

IIUC both of you are suggesting to requeue with a backoff (e.g. 10 seconds) when the error is not of type _notFound_ and not _forbidden_. I added it now.
But I still wonder why waiting a backoff would be better. It sounds like you suspect that the `notFound` error is temporary, and after the backoff it will be resolved. 
Could be, but if not than we should add in reconcile a flag/counter to stop future backoffs.
WDYT?

pkg/utils/nodes.go
```diff
@@ -0,0 +1,24 @@
...
+// CheckNodeName returns an error if nodeName doesn't match any node name in the cluster, otherwise a nil
+func CheckNodeName(r client.Reader, nodeName string) error {
+	node := &corev1.Node{}
+	objNodeName := types.NamespacedName{Name: nodeName}
+	err := r.Get(context.Background(), objNodeName, node)
+	if err != nil {
+		// no match between CR name and the cluster nodes
+		msg := fmt.Sprintf("node %s is invalid", nodeName)
+		err = buildApiError(err, msg)
+		return fmt.Errorf("CR's name doesn't match any node name in the cluster - %w", err)
```
mshitrit, line 21: I think it would simply things to replace this by wrapping the error.
Something like:
```
"github.com/pkg/errors"
...
errors.Wrapf(err, "node %s is invalid", nodeName)
```

Later on in `Reconcile` when you need to check whether node was found or not you can use the error cause to do that:
```
if apiErrors.IsNotFound(errors.Cause(err)) {
	return emptyResult, err
}
```

controllers/fenceagentsremediation_controller.go
```diff
@@ -79,7 +84,16 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
 	}
-	// TODO: Validate FAR CR name to nodeName. Run isNodeNameValid
+	// Validate FAR CR name to match a nodeName from the cluster
+	r.Log.Info("Check FAR CR's name")
+	if err := farUtils.CheckNodeName(r.Client, req.Name); err != nil {
+		if apiErrors.IsNotFound(err) {
+			return emptyResult, err
+		}
+		r.Log.Error(err, "Failed to validate FAR CR's name- retry")
+		return ctrl.Result{RequeueAfter: apiFailureRetryTimeout}, nil
```
mshitrit, line 94: Why not just return the error here ?

pkg/utils/nodes.go
slintes: > But I still wonder why waiting a backoff would be better. 

No hardcoded backoff, just return an error. `controller-runtime` will requeue with an increasing backoff automatically. This ensures to retry but not overload a potentailly already broken api server.

pkg/utils/nodes.go
slintes: > Later on in Reconcile...

That check is better placed here IMHO, because isNotFound is a valid result for this function and not an error.

razo7: >  isNotFound is a valid result for this function and not an error.

> node found = name valid
node not found = name invalid
any other error = retry

When `node not found = name invalid`, then we should not return an error _CheckNodeName_ function and in reconcile or are you referring that _CheckNodeName_ function should not return an error but in reconcile we should raise an error (since we didn't found a match between CR's name and cluster node names)?

I am still not sure what is your suggestion. I went with the second option.

slintes: > the func signature might need an update as well (atm the is prefix suggests that the return value is a boolean, bit it's an error)

you removed the `is` prefix... what about returning a bool instead? That would make things easier...

```
func Reconcile() {
  ...
  isValid, err := isNodeNameValid(name)
  if err != nil {
    return err
  } else if !isValid {
    // log, event, whatever, and return without error, because a requeue does not make sense
    return result, nil
  {
  // remediate...
}

func isNodeNameValid(name) (bool, error) {
  err := getNode(name)
  if isNotFound(err) {
    return false, nil
  } else if err != nil {
    return false, err // bool value doesn't matter
  }
  return true, nil
}
```

controllers/fenceagentsremediation_controller.go
```diff
@@ -79,7 +80,17 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "failed to get FAR CR")
 		return emptyResult, err
 	}
-	// TODO: Validate FAR CR name to nodeName. Run isNodeNameValid
+	// Validate FAR CR name to match a nodeName from the cluster
+	r.Log.Info("Check FAR CR's name")
+	if exist, err := farUtils.IsNodeNameValid(r.Client, req.Name); !exist {
```
slintes, line 85: - nit:  `exists` doesn't match function name, `valid` or `isValid` would be better
- you rely on the bool value being false in case of an error. The bool value shouldn't matter though IMHO. Better first check for the error, then the bool.

pkg/utils/nodes.go
```diff
@@ -0,0 +1,28 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// IsNodeNameValid returns an error if nodeName doesn't match any node name int the cluster, otherwise a nil
+func IsNodeNameValid(r client.Reader, nodeName string) (bool, error) {
+	node := &corev1.Node{}
+	nodeExist := true
```
slintes, line 16: what do you need this for? IMHO it would be easier to read when using true or false directly
```
if isNotFound(err) {
  return false, nil
} else if err != nil {
  return false, err
}
return true, nil
```
or if you want to keep error handling nested:
```
if err != nil {
  if isNotFound(err) {
    return false, nil
  }
  return false, err
}
return true, nil
```

slintes (changes requested):

pkg/utils/nodes.go
```diff
@@ -0,0 +1,28 @@
...
+	objNodeName := types.NamespacedName{Name: nodeName}
+	err := r.Get(context.Background(), objNodeName, node)
+	if err != nil {
+		nodeExist = false
+		if !apiErrors.IsNotFound(err) {
+			// no match between CR name and the cluster nodes and the error is not of type notFound
+			err = buildApiError(err, "unknown error")
```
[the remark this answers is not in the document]
razo7: Now, when I got a better understanding there is no value.. The func and file was removed :)

controllers/fenceagentsremediation_controller.go
```diff
@@ -79,7 +80,16 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "failed to get FAR CR")
 		return emptyResult, err
 	}
-	// TODO: Validate FAR CR name to nodeName. Run isNodeNameValid
+	// Validate FAR CR name to match a nodeName from the cluster
+	r.Log.Info("Check FAR CR's name")
+	valid, err := farUtils.IsNodeNameValid(r.Client, req.Name)
```
mshitrit, line 85: Nit: I  prefer `utils` over `farUtils`.
in any case this is the name of the pkg and it's clear we are in the context of far.

pkg/utils/nodes.go
```diff
@@ -0,0 +1,26 @@
...
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// IsNodeNameValid returns an error if nodeName doesn't match any node name int the cluster, otherwise a nil
+func IsNodeNameValid(r client.Reader, nodeName string) (bool, error) {
```
mshitrit, line 13: use Client interface  (i.e client.Client)

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -69,39 +72,53 @@ var _ = Describe("FAR Controller", func() {
...
 
 	Context("Functionality", func() {
-		Context("buildFenceAgentParams", func() {
-			When("FAR's name isn't a node name", func() {
+		Context("buildFenceAgentParams - check CR name", func() {
+			When("FAR's name doesn't match a node name", func() {
 				It("should fail", func() {
 					underTestFAR.ObjectMeta.Name = dummyNodeName
 					_, err := buildFenceAgentParams(underTestFAR)
 					Expect(err).To(HaveOccurred())
```
mshitrit, line 80: Nit: maybe check the error message as well instead of just existence of error

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -69,39 +72,53 @@ var _ = Describe("FAR Controller", func() {
 	underTestFAR = newFenceAgentsRemediation(validNodeName, fenceAgentIPMI, testShareParam, testNodeParam)
 
 	Context("Functionality", func() {
-		Context("buildFenceAgentParams", func() {
-			When("FAR's name isn't a node name", func() {
+		Context("buildFenceAgentParams - check CR name", func() {
```
clobrano, line 75: Nit: using function names in Ginkgo's node description makes it harder to understand what the test is verifying.
For example, in this context, it might be just "Check CR name against node"

razo7: Ok, do you have a suggestion for better naming?
I see a small value of testing the 2 functions in UT, but it does catch errors when I do small changes in them (or their dependencies)

controllers/fenceagentsremediation_controller.go
```diff
@@ -30,7 +30,11 @@ import (
 
 	"github.com/medik8s/fence-agents-remediation/api/v1alpha1"
 	"github.com/medik8s/fence-agents-remediation/pkg/cli"
-	farUtils "github.com/medik8s/fence-agents-remediation/pkg/utils"
+	utils "github.com/medik8s/fence-agents-remediation/pkg/utils"
```
slintes, line 33: redundant alias

note to self: add a check for this to sort-imports :)

controllers/fenceagentsremediation_controller_test.go
clobrano: For example, in this context, it might be just "Check CR name against node"

controllers/fenceagentsremediation_controller.go
```diff
@@ -79,10 +84,19 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
 	}
-	// TODO: Validate FAR CR name to nodeName. Run isNodeNameValid
+	// Validate FAR CR name to match a nodeName from the cluster
+	r.Log.Info("Check FAR CR's name")
+	valid, err := utils.IsNodeNameValid(r.Client, req.Name)
+	if err != nil {
+		return emptyResult, err
+	}
+	if !valid {
+		r.Log.Info("consider recreating the CR - invalid CR's name to the cluster node name", "FAR CR's Name", req.Name)
```
slintes, line 94: nit: I would skip the recommendation on how to fix the issue, you don't know why the issue happened. Also, the error message isn't clear to me... maybe "didn't find node matching CR's name" ? :thinking:

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -31,6 +31,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/medik8s/fence-agents-remediation/api/v1alpha1"
+	utils "github.com/medik8s/fence-agents-remediation/pkg/utils"
```
slintes, line 34: redundant alias

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -69,39 +71,54 @@ var _ = Describe("FAR Controller", func() {
...
 					underTestFAR.ObjectMeta.Name = dummyNodeName
 					_, err := buildFenceAgentParams(underTestFAR)
 					Expect(err).To(HaveOccurred())
+					Expect(err).To(Equal(errors.New(errorBuildingFAParams)))
 				})
+			})
+			When("FAR's name does match a node name", func() {
 				It("should succeed", func() {
 					underTestFAR.ObjectMeta.Name = validNodeName
 					_, err := buildFenceAgentParams(underTestFAR)
```
slintes, line 86: just fyi, this is a bit shorter and nicer:
`Expect(buildFenceAgentParams(underTestFAR)).To(Succeed())`

pkg/utils/pods.go
```diff
@@ -3,18 +3,15 @@ package utils
...
-	apiErrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // GetFenceAgentsRemediationPod fetches the FAR pod based on FAR's label and namespace
-func GetFenceAgentsRemediationPod(r client.Reader) (*corev1.Pod, error) {
+func GetFenceAgentsRemediationPod(r client.Client) (*corev1.Pod, error) {
```
slintes, line 14: nit: when you read only, Reader is absolutely fine to use

pkg/utils/nodes.go
```diff
@@ -0,0 +1,26 @@
...
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// IsNodeNameValid returns an error if nodeName doesn't match any node name int the cluster, otherwise a nil
+func IsNodeNameValid(r client.Client, nodeName string) (bool, error) {
```
slintes, line 13: nit: when you read only, Reader is absolutely fine to use

test/e2e/far_e2e_test.go
```diff
@@ -16,8 +16,8 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 
 	"github.com/medik8s/fence-agents-remediation/api/v1alpha1"
-	farUtils "github.com/medik8s/fence-agents-remediation/pkg/utils"
-	farE2eUtils "github.com/medik8s/fence-agents-remediation/test/e2e/utils"
+	utils "github.com/medik8s/fence-agents-remediation/pkg/utils"
```
slintes, line 19: redundant alias

slintes (commented):
overall looks good now, some nits only inside

controllers/fenceagentsremediation_controller_test.go
razo7: I see, then I tend to let the context just be the function name (`buildFenceAgentParams`), and just remove the ` - check CR name`

controllers/fenceagentsremediation_controller_test.go
razo7: I tried that, and it failed since on success the function returns a string list.

Error :

>    [FAILED] Expected an error-type.  Got:
      <[]string | len:6, cap:8>: [
          "--username=admin",
          ...

controllers/fenceagentsremediation_controller.go
razo7: Gotcha, and I agree 👍🏻

clobrano:
/lgtm
/hold

giving others the chance to check if their reviews are addressed

controllers/fenceagentsremediation_controller_test.go
razo7: Indeed, checking behavior rather than a function (by it's name) seems more right.
Something to think off for next PRs and writing UTs

razo7:
I believe 2/3 acks is good enough for this PR :slightly_smiling_face: 
In any case @mshitrit feel free to reply here, and I will address it with a new PR :D

/unhold

PR #49, opened 2023-05-11
Set Environment Variable for Tests

razo7:
A follow up PR for #42 which add an environment variable DEPLOYMENT_NAMESPACE to the `manager` container, but we need to set in the UT and E2E tests, otherwise they have raised an error (and now they will fail).

### DIFF
--- a/pkg/utils/pods.go
+++ b/pkg/utils/pods.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -83,6 +83,8 @@ var _ = BeforeSuite(func() {
 	k8sClient, err = ctrl.New(config, ctrl.Options{Scheme: scheme.Scheme})
 	Expect(err).NotTo(HaveOccurred())
 
+	os.Setenv("DEPLOYMENT_NAMESPACE", operatorInstalledNamespcae)
```
slintes, line 86: Why is this needed in the e2e test? 🤔

pkg/utils/pods.go
```diff
@@ -24,6 +24,7 @@ func GetFenceAgentsRemediationPod(r client.Reader) (*corev1.Pod, error) {
 	podNamespace, err := GetDeploymentNamespace()
 	if err != nil {
 		logger.Error(err, "failed fetching FAR namespace")
+		return nil, err
```
clobrano, line 27: This does not seem just related to the test.
Maybe a specific commit (if not a PR) to highlight this change might be beneficial

test/e2e/e2e_suite_test.go
razo7: Function _GetFenceAgentsRemediationPod_ calls function _GetDeploymentNamespace_  which is looking for the namespace value  from **DEPLOYMENT_NAMESPACE** envioronment variable. 
Without this line I receive an error (see below) and the e2e test fails with [timeout](https://github.com/medik8s/fence-agents-remediation/blob/main/test/e2e/far_e2e_test.go#L239-L246).

Error:

> failed to get pod. Might try again	{"error": "failed fetching FAR namespace - DEPLOYMENT_NAMESPACE must be set"}

slintes: nevermind, found it, it's indirectly needed by the test code as well, and not just controller code

razo7: So lgtm ? :)

PR #47, opened 2023-05-10
Apply PSA Changes

razo7:
Following [NMO's PR](https://github.com/medik8s/node-maintenance-operator/pull/61) to comply with Kubernetes v1.25 PSA changes we drop all the capabilities for the `manager` container.

We also add the `seccompProfile.type: RuntimeDefault` by default, unlike for NMO, since it has been tested and deployed from OCP 4.11+.

Related docs:
- https://kubernetes.io/docs/concepts/security/pod-security-standards/
- https://kubernetes.io/docs/concepts/security/pod-security-admission/
- https://master.sdk.operatorframework.io/docs/best-practices/pod-security-standards/
- https://kubernetes.io/docs/setup/best-practices/enforcing-pod-security-standards/#adopt-a-multi-mode-strategy
- https://connect.redhat.com/en/blog/important-openshift-changes-pod-security-standards

PR #51, opened 2023-05-29
[WIP] RBAC Proxy TLS

razo7:
New deprecation warnings from **_kube-rbac-proxy_** container, after updating kube-rbac-proxy version to v0.14.1 in #47.
According to https://github.com/brancz/kube-rbac-proxy/issues/196 TLS will be mandatory from V1 release.

In order to comply with it we add a self-signed TLS to the kube-rbac-proxy container.

*Since the secret is exposed and not encrypted, then we won't merge it ATM and we only want to test it now.


slintes:
Using a hardcoded, public cert and key is a bad idea. Let's discuss a better solution...

/close

PR #52, opened 2023-05-30
Reduce Container Size

razo7:
Drastically minimize container size (~1GB -> ~210MB):

1. Cleaning temporary files after installing external packages -  by using`clean all` (see it also from [container best practices-> Clean temporary files](https://docs.openshift.com/container-platform/4.13/openshift_images/create-images.html))
2. Build the container in a [multi-stage](https://docs.docker.com/build/building/multi-stage/) manner to exclude unneeded packages Golang, source files and .git from the final image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,6 @@
 # Build the manager binary
 FROM quay.io/centos/centos:stream8 AS builder
-RUN dnf install -y dnf-plugins-core \
-    && dnf config-manager --set-enabled ha \
-    && dnf install -y golang git \
+RUN dnf install -y golang git \
     && dnf clean all -y
 
 # Ensure correct Go version
@@ -12,7 +10,7 @@ RUN go install golang.org/dl/go${GO_VERSION}@latest \
     && /bin/cp -f ~/go/bin/go${GO_VERSION} /usr/bin/go \
     && go version
 
-WORKDIR /
+WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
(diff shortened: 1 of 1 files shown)


slintes:
what about using a multi stage container build, so the resulting image doesn't contain Go, source files and .git?

razo7:
> what about using a multi stage container build, so the resulting image doesn't contain Go, source files and .git?

That's definitely something I am keen of doing for not only removing the above redundant stuff, but to decrease the enormous size of the image (around 1 GB prior to this PR cleaning). 

I haven't done it until now, since I have a problem of using a smaller base image and still be able to find and download the **fence-agents-all package**. But thinking again about it I think we can still add another base image for the purpose you have mentioned and deal with the fence-agents-all package problem afterwards (in a different PR).

What is the problem? In short I wanted to use **UBI** base image (or it's variants, such as _ubi-minimal_ ),  after building the FAR binary and with UBI try to install fence-agents-all package. Since this package is part of HighAvailability add-on repo, then we need to enable it in UBI so it could find the desired package. Enabling is easy, but having/adding this repo is the problem.

slintes:
> I haven't done it until now, since I have a problem of using a smaller base image

you can use the same image in both build steps. The final one will just not contain Go and sources / `.git`. I assume that will save a lot of space already with very low effort. You can look for a smaller base image later.

razo7:
> The final one will just not contain Go and sources / .git. I assume that will save a lot of space already with very low effort.

You are right, it did save ~640 MB, and the resulting image is ~210 MB. Still not as minimal as other Medik8s operators, but in a much better place than around 1 GB prior to this PR.

PR #53, opened 2023-06-07
Support and Handle Taints

razo7:
In the process of supporting workload deletion we are supporting the Medik8s remediation taint (with FAR name value), and [_NoExecute_ effect](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions):

- Adding taint support - Specifically supporting the Medik8s remediation taint, and possibly in the future more taints.
- Support the taint removal with a Finalizer
- Add two RBAC verbs on node resource (`delete`, and `update`) - It is needed for adding and removing taint on a cluster node.
- Remove redundant dummy E2E test
- Addition/removal of taint or finalizer will only be handled after Node-CR validation 
- UT for valid and invalid Node-CR 

[ECOPROJECT-1321](https://issues.redhat.com//browse/ECOPROJECT-1321) and [ECOPROJECT-1322](https://issues.redhat.com//browse/ECOPROJECT-1322)

### DIFF
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -88,6 +88,18 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "Failed to get FenceAgentsRemediation CR")
 		return emptyResult, err
 	}
+	// Validate FAR CR name to match a nodeName from the cluster
+	r.Log.Info("Check FAR CR's name")
+	valid, err := utils.IsNodeNameValid(r.Client, req.Name)
+	if err != nil {
+		r.Log.Error(err, "Unexpected error when validating CR's name with nodes' names", "CR's Name", req.Name)
+		return emptyResult, err
+	}
+	if !valid {
+		r.Log.Error(err, "Didn't find a node matching the CR's name", "CR's Name", req.Name)
+		return emptyResult, nil
+	}
+
 	// Add finalizer when the CR is created
 	if !controllerutil.ContainsFinalizer(far, v1alpha1.FARFinalizer) && far.ObjectMeta.DeletionTimestamp.IsZero() {
 		controllerutil.AddFinalizer(far, v1alpha1.FARFinalizer)
@@ -110,17 +122,6 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		return emptyResult, nil
 	}
 
-	// Validate FAR CR name to match a nodeName from the cluster
-	r.Log.Info("Check FAR CR's name")
-	valid, err := utils.IsNodeNameValid(r.Client, req.Name)
-	if err != nil {
-		r.Log.Error(err, "Unexpected error when validating CR's name with nodes' names", "CR's Name", req.Name)
-		return emptyResult, err
-	}
-	if !valid {
-		r.Log.Error(err, "Didn't find a node matching the CR's name", "CR's Name", req.Name)
-		return emptyResult, nil
-	}
 	// Fetch the FAR's pod
 	r.Log.Info("Fetch FAR's pod")
 	pod, err := utils.GetFenceAgentsRemediationPod(r.Client)
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/medik8s/fence-agents-remediation/api/v1alpha1"
 	"github.com/medik8s/fence-agents-remediation/pkg/cli"
@@ -87,6 +88,29 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "Failed to get FenceAgentsRemediation CR")
 		return emptyResult, err
 	}
+	// Add finalizer when object is created
+	if far.ObjectMeta.DeletionTimestamp.IsZero() {
+		controllerutil.AddFinalizer(far, v1alpha1.FARFinalizer)
+		if err := r.Client.Update(context.Background(), far); err != nil {
+			return emptyResult, fmt.Errorf("failed to add finalizer to the CR - %w", err)
+		}
+	} else if controllerutil.ContainsFinalizer(far, v1alpha1.FARFinalizer) {
+		r.Log.Info("CR's deletion timestamp is not zero, and FAR finalizer exists", "CR Name", req.Name)
+		// The object is being deleted
+
+		// remove node's taints
+		if err := utils.RemoveTaint(r.Client, far.Name); err != nil && !apiErrors.IsNotFound(err) {
+			return emptyResult, err
+		}
+		// remove finalizer
+		controllerutil.RemoveFinalizer(far, v1alpha1.FARFinalizer)
+		if err := r.Client.Update(context.Background(), far); err != nil {
+			return emptyResult, fmt.Errorf("failed to remove finalizer from CR - %w", err)
+		}
+		r.Log.Info("Finalizer was removed", "CR Name", req.Name)
+		return emptyResult, nil
+	}
+
 	// Validate FAR CR name to match a nodeName from the cluster
 	r.Log.Info("Check FAR CR's name")
 	valid, err := utils.IsNodeNameValid(r.Client, req.Name)
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -57,7 +57,7 @@ func (r *FenceAgentsRemediationReconciler) SetupWithManager(mgr ctrl.Manager) er
 
 //+kubebuilder:rbac:groups=core,resources=pods/exec,verbs=create
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;update;delete;deletecollection
-//+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch;patch;delete
+//+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch;update;delete
 //+kubebuilder:rbac:groups=fence-agents-remediation.medik8s.io,resources=fenceagentsremediations,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=fence-agents-remediation.medik8s.io,resources=fenceagentsremediations/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=fence-agents-remediation.medik8s.io,resources=fenceagentsremediations/finalizers,verbs=update
@@ -88,16 +88,15 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "Failed to get FenceAgentsRemediation CR")
 		return emptyResult, err
 	}
-	// Add finalizer when object is created
-	if far.ObjectMeta.DeletionTimestamp.IsZero() {
+	// Add finalizer when the CR is created
+	if !controllerutil.ContainsFinalizer(far, v1alpha1.FARFinalizer) && far.ObjectMeta.DeletionTimestamp.IsZero() {
 		controllerutil.AddFinalizer(far, v1alpha1.FARFinalizer)
 		if err := r.Client.Update(context.Background(), far); err != nil {
 			return emptyResult, fmt.Errorf("failed to add finalizer to the CR - %w", err)
 		}
-	} else if controllerutil.ContainsFinalizer(far, v1alpha1.FARFinalizer) {
+	} else if controllerutil.ContainsFinalizer(far, v1alpha1.FARFinalizer) && !far.ObjectMeta.DeletionTimestamp.IsZero() {
+		// Delete CR only when a finalizer and DeletionTimestamp are set
 		r.Log.Info("CR's deletion timestamp is not zero, and FAR finalizer exists", "CR Name", req.Name)
-		// The object is being deleted
-
 		// remove node's taints
 		if err := utils.RemoveTaint(r.Client, far.Name); err != nil && !apiErrors.IsNotFound(err) {
 			return emptyResult, err
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -57,7 +57,7 @@ func (r *FenceAgentsRemediationReconciler) SetupWithManager(mgr ctrl.Manager) er
 
 //+kubebuilder:rbac:groups=core,resources=pods/exec,verbs=create
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;update;delete;deletecollection
-//+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch;patch;delete
 //+kubebuilder:rbac:groups=fence-agents-remediation.medik8s.io,resources=fenceagentsremediations,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=fence-agents-remediation.medik8s.io,resources=fenceagentsremediations/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=fence-agents-remediation.medik8s.io,resources=fenceagentsremediations/finalizers,verbs=update
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -114,6 +114,11 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "Invalid sharedParameters/nodeParameters from CR", "CR's Name", req.Name)
 		return emptyResult, err
 	}
+	// Add medik8s remediation taint
+	r.Log.Info("Add Medik8s remediation taint", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
+	if err := utils.AppendTaint(r.Client, far.Name); err != nil {
+		return emptyResult, err
+	}
 	cmd := append([]string{far.Spec.Agent}, faParams...)
 	// The Fence Agent is excutable and the parameters structure are valid, but we don't check their values
 	r.Log.Info("Execute the fence agent", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
--- a/controllers/fenceagentsremediation_controller_test.go
+++ b/controllers/fenceagentsremediation_controller_test.go
@@ -29,14 +29,17 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/medik8s/fence-agents-remediation/api/v1alpha1"
 	"github.com/medik8s/fence-agents-remediation/pkg/utils"
 )
 
 const (
-	dummyNodeName  = "dummy-node"
-	validNodeName  = "worker-0"
+	dummyNode      = "dummy-node"
+	node01         = "worker-0"
+	node02         = "worker-1"
 	fenceAgentIPMI = "fence_ipmilan"
 )
 
@@ -68,13 +71,13 @@ var _ = Describe("FAR Controller", func() {
 			"worker-2": "6235",
 		},
 	}
-	underTestFAR = newFenceAgentsRemediation(validNodeName, fenceAgentIPMI, testShareParam, testNodeParam)
+	underTestFAR = getFenceAgentsRemediation(node01, fenceAgentIPMI, testShareParam, testNodeParam)
 
 	Context("Functionality", func() {
 		Context("buildFenceAgentParams", func() {
 			When("FAR CR's name doesn't match a node name", func() {
 				It("should fail", func() {
-					underTestFAR.ObjectMeta.Name = dummyNodeName
+					underTestFAR.ObjectMeta.Name = dummyNode
 					_, err := buildFenceAgentParams(underTestFAR)
 					Expect(err).To(HaveOccurred())
 					Expect(err).To(Equal(errors.New(errorMissingNodeParams)))
@@ -82,7 +85,7 @@ var _ = Describe("FAR Controller", func() {
 			})
 			When("FAR CR's name does match a node name", func() {
 				It("should succeed", func() {
-					underTestFAR.ObjectMeta.Name = validNodeName
+					underTestFAR.ObjectMeta.Name = node01
 					Expect(buildFenceAgentParams(underTestFAR)).Error().NotTo(HaveOccurred())
 				})
 			})
@@ -90,18 +93,18 @@ var _ = Describe("FAR Controller", func() {
 
 		Context("IsNodeNameValid", func() {
 			BeforeEach(func() {
-				node = getNode(validNodeName)
+				node = getNode(node01)
 				DeferCleanup(k8sClient.Delete, context.Background(), node)
 				Expect(k8sClient.Create(context.Background(), node)).To(Succeed())
 			})
 			When("FAR CR's name doesn't match to an existing node name", func() {
 				It("should fail", func() {
-					Expect(utils.IsNodeNameValid(k8sClient, dummyNodeName)).To(BeFalse())
+					Expect(utils.IsNodeNameValid(k8sClient, dummyNode)).To(BeFalse())
 				})
 			})
 			When("FAR's name does match to an existing node name", func() {
 				It("should succeed", func() {
-					Expect(utils.IsNodeNameValid(k8sClient, validNodeName)).To(BeTrue())
+					Expect(utils.IsNodeNameValid(k8sClient, node01)).To(BeTrue())
 				})
 			})
 		})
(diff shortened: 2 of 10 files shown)


controllers/fenceagentsremediation_controller.go
```diff
@@ -48,13 +49,13 @@ type FenceAgentsRemediationReconciler struct {
 // SetupWithManager sets up the controller with the Manager.
 func (r *FenceAgentsRemediationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&v1alpha1.FenceAgentsRemediation{}).
+		For(&farV1alpha1.FenceAgentsRemediation{}).
```
mshitrit, line 52: Nit: what is the reason for this change ?

controllers/fenceagentsremediation_controller.go
```diff
@@ -27,8 +27,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
-	"github.com/medik8s/fence-agents-remediation/api/v1alpha1"
+	farV1alpha1 "github.com/medik8s/fence-agents-remediation/api/v1alpha1"
```
mshitrit, line 32: Nit: what is the reason for this change ?

controllers/fenceagentsremediation_controller.go
```diff
@@ -74,39 +75,70 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
 		if apiErrors.IsNotFound(err) {
-			// FAR is deleted, stop reconciling
-			r.Log.Info("FAR CR is deleted - nothing to do", "CR Name", req.Name, "CR Namespace", req.Namespace)
+			// Request object of FAR CR was not found, could have been deleted after reconcile request.
+			// Return and don't requeue
+			r.Log.Info("FenceAgentsRemediation CR was not found", "CR Name", req.Name, "CR Namespace", req.Namespace)
 			return emptyResult, nil
 		}
-		r.Log.Error(err, "failed to get FAR CR")
+		r.Log.Info("Failed to get FenceAgentsRemediation CR, requeuing.", "CR Name", req.Name)
```
mshitrit, line 86: I think an error log was appropriate here, why change to Info ?

controllers/fenceagentsremediation_controller.go
```diff
@@ -74,39 +75,70 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
 		return emptyResult, err
 	}
+	// Add finalizer when object is created
+	if far.ObjectMeta.DeletionTimestamp.IsZero() {
```
mshitrit, line 90: IIUC this will also be triggered on change of the CR as well, maybe just check if there is already a finalizer.

controllers/fenceagentsremediation_controller.go
```diff
@@ -74,39 +75,70 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
+
 	// Validate FAR CR name to match a nodeName from the cluster
 	r.Log.Info("Check FAR CR's name")
 	valid, err := utils.IsNodeNameValid(r.Client, req.Name)
 	if err != nil {
 		return emptyResult, err
 	}
 	if !valid {
-		r.Log.Info("didn't find a node matching the CR's name", "CR's Name", req.Name)
+		r.Log.Info("Didn't find a node matching the CR's name", "CR's Name", req.Name)
```
mshitrit, line 119: IMO this scenario might need something more visible than Info log

pkg/utils/taints.go
```diff
@@ -0,0 +1,121 @@
...
+			continue
+		}
+		newTaints = append(newTaints, taints[i])
+	}
+	return newTaints, deleted
+}
+
+// HandleTaint calls other taint-functions basesd on the taint name and acion.
+// If there is no associated node for FAR or taint name is invalid or taint action is invalid then it returns an error
+func HandleTaint(r client.Client, taintName, actionOnTaint string, far *farV1alpha1.FenceAgentsRemediation) error {
```
slintes, line 51: What's the value of this generic `handleXyz` function, compared to having the usual `addXyz` and `removeXyz` functions? 🤔

pkg/utils/taints.go
```diff
@@ -0,0 +1,121 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	farV1alpha1 "github.com/medik8s/fence-agents-remediation/api/v1alpha1"
+)
+
+var (
+	loggerTaint           = ctrl.Log.WithName("taints")
+	FARNodeNoExecuteTaint = &corev1.Taint{
+		Key:    farV1alpha1.Medik8sNoExecuteTaintKey,
+		Value:  farV1alpha1.FARNoExecuteTaintKeyValue,
```
mshitrit, line 20: I think the value can be removed, since we have a unique key and always using the same value

pkg/utils/taints.go
```diff
@@ -0,0 +1,121 @@
...
+// If there is no associated node for FAR or taint name is invalid or taint action is invalid then it returns an error
+func HandleTaint(r client.Client, taintName, actionOnTaint string, far *farV1alpha1.FenceAgentsRemediation) error {
+	var taintObj *corev1.Taint
+	// check if taint exist
+	switch taintName {
+	case farV1alpha1.FARNoExecuteTaint:
+		taintObj = FARNodeNoExecuteTaint
+	default:
+		return fmt.Errorf("invalid taintName - %s", taintName)
+	}
```
mshitrit, line 59: Maybe remove, and assume it's called for `FARNoExecuteTaint`

pkg/utils/taints.go
```diff
@@ -0,0 +1,121 @@
...
+
+	switch actionOnTaint {
+	case farV1alpha1.AppendTaintAction:
+		return appendTaint(r, taintObj, node)
+	case farV1alpha1.RemoveTaintAction:
+		return removeTaint(r, taintObj, node)
+	default:
+		return fmt.Errorf("invalid action for taint - %s", actionOnTaint)
+	}
```
mshitrit, line 73: In case there are only two valid options, I suggest to simplify by replacing with a bool

controllers/fenceagentsremediation_controller.go
razo7: Only my preferance. I though it is much more clear this way, but I can revert it

controllers/fenceagentsremediation_controller.go
razo7: Which kind of change? IIUC it won't be zero only when the CR was deleted (and it's `DeletionTimestamp` is greater than zero)

controllers/fenceagentsremediation_controller.go
razo7: Such as? Something in UI?

pkg/utils/taints.go
razo7: Wouldn't it be useful for identifing the remediatior by the tain't value?

pkg/utils/taints.go
razo7: Remove the switch and assume that it was called for FARNoExecuteTaint?
I though of having this check (with switch or if-else) to maybe support the out-of-service taint

pkg/utils/taints.go
razo7: In both `addXyz` and `removeXyz` functions I was thinking to check the the taintName (remediationTain/ another taint), and then fetch the node from FAR. Thus using `handleXyz` we can generalize it and check it prior to call the `addXyz` and `removeXyz` functions.
It enables us expose only one public function rather than two. But maybe it doesn't worth it :thinking

controllers/fenceagentsremediation_controller.go
razo7: Because we return the error on the next line. Otherwise, the log will have two errors (one from log and from return) for one error..

pkg/utils/taints.go
```diff
@@ -0,0 +1,120 @@
...
+		if taintToDelete.MatchTaint(&taints[i]) {
+			deleted = true
+			continue
+		}
+		newTaints = append(newTaints, taints[i])
+	}
+	return newTaints, deleted
+}
+
+// HandleTaint calls other taint-functions basesd on the taint name and acion.
```
clobrano, line 49: typo: `s/basesd/based`

pkg/utils/nodes.go
```diff
@@ -24,3 +26,15 @@ func IsNodeNameValid(r client.Reader, nodeName string) (bool, error) {
 	}
 	return true, nil
 }
+
+// getNodeFromFAR returns the unhealthy node reported in the given far CR (by it's name)
+func getNodeFromFAR(r client.Client, far *farV1alpha1.FenceAgentsRemediation) (*corev1.Node, error) {
```
clobrano, line 31: nit: I think that the function might be easier to understand if named like
`getNodeWithName(r client.Client, name string) (*corev1.Node, error)`

It is up to the caller give the right name (`far.Name` in this case)

WDYT?

pkg/utils/taints.go
```diff
@@ -0,0 +1,120 @@
...
+// appendTaint appends new taint to the taint list when it is not present, and returns error if it fails in the process
+func appendTaint(r client.Client, taintObj *corev1.Taint, node *corev1.Node) error {
+
+	if taintExists(node.Spec.Taints, taintObj) {
+		return nil
+	}
+
+	// add the taint to the taint list
+	patch := client.MergeFrom(node.DeepCopy())
+	taint := *taintObj
```
clobrano, line 84: sorry if it is obvious, but why this assignment?

pkg/utils/taints.go
slintes: To me it's irritating TBH, I think it adds more complexity than value.
You can have 2 functions which share common code by just putting that shared code into another func.

pkg/utils/taints.go
razo7: To store and declare the taint object in a different place (in the beginning), and in the function we just do assignment.

controllers/fenceagentsremediation_controller.go
mshitrit: Sorry I don't quite follow.
- Why returning the error should effect whether it should be logged or not ?
- And even if it does, you log it anyway, so why log an error as an info ?

controllers/fenceagentsremediation_controller.go
mshitrit: That is correct, however the DeletionTimeStamp will be equal to zero not only when the CR is created (which is the assumption in the code) but also when the CR is modified (without being deleted).

controllers/fenceagentsremediation_controller.go
mshitrit: At least an error log, maybe an event.

pkg/utils/taints.go
mshitrit: The key is already unique for the remediator, IIUC it would have made sense if the same remediator would like to apply the same taint twice with different distinguished variations.

pkg/utils/taints.go
mshitrit: I wouldn't pre-construct a more complicated code for an option we might eventually not implement, worst case we can always make the chance when adding that taint.

controllers/fenceagentsremediation_controller.go
razo7: > CR is modified (without being deleted)

And this can happen when the FAR's CR will have a status, right?
Anyway, there is no harm of restricting the if clause 👍🏻

pkg/utils/taints.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	// find node by name
+	node, err := getNodeWithName(r, nodeName)
+	if err != nil {
+		return nil, nil, err
+	}
+	return taintObj, node, nil
+}
+
+// appendTaint appends new taint to the taint list when it is not present, and returns error if it fails in the process
```
mshitrit, line 68: Nit: public method comment should be identical to method name (i.e AppendTaint, instead of appendTaint)

pkg/utils/taints.go
mshitrit: This is indeed confusing.

IIUC `getTaintAndNode` is returning  a singleton taint which shouldn't be modified, therefore enforcing this assignment.

I think this is wrong, I suggest instead that `getTaintAndNode` would return a copy of the taint which in turn could be modified.

pkg/utils/taints.go
razo7: > IIUC it would have made sense if the same remediator would like to apply the same taint twice with different distinguished variations.

Could be, but SNR doesn't have this scenario, and it's taint does have a value - [SNR's remediation taint](https://github.com/medik8s/self-node-remediation/blob/main/controllers/selfnoderemediation_controller.go#L59-L63).

I was thinking on a different case with NHC escalation.
When different remediators are running one after another, and the taint's value can act as an identifier for the remediator that it is running ATM. I assume that it can easily be found/described in NHC log, but maybe adding in the taint's value will be useful for other stuff (e.g. fetching metrics on the acting remediator with NHC)

pkg/utils/taints.go
razo7: I have removed the `handleXyz` function

pkg/utils/taints.go
slintes: > The key is already unique for the remediator

Looking at SNR, I´d say no, it isn't: https://github.com/medik8s/self-node-remediation/blob/main/controllers/selfnoderemediation_controller.go#L59

Using the value for the remediator makes sense to me

controllers/fenceagentsremediation_controller.go
razo7: > And this can happen when the FAR's CR will have a status, right?
Or are you referring to an update by adding/removing the finalizer?

pkg/utils/taints.go
slintes: - getTaintAndNode should not do two unrelated things in one function in first place IMHO
- getting a taint by name, which exist as pointer in a global var, is irritating IMHO... why not just a createNoExecuteTaint function, which builds and returns the taint by value and not pointer?

controllers/fenceagentsremediation_controller.go
razo7: I was thinking of avoiding logging of errors before reconcile returns an error in order to avoid noisy reconcile (duplicating error), but it is better to duplicate rather just depends on errors from reconcile.
Fixed in https://github.com/medik8s/fence-agents-remediation/pull/55/commits/f1cec90b2a1e1151e7e432259a52efd677fb2bed of #55

controllers/fenceagentsremediation_controller.go
slintes: it doesn't matter what's causing an update, CRs can be updated, that's it. An empty deletion timestamp just isn't an indicator for a just created resource.

controllers/fenceagentsremediation_controller.go
```diff
@@ -84,6 +85,29 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
+		if err := utils.RemoveTaint(r.Client, far.Name); err != nil && !apiErrors.IsNotFound(err) {
+			return emptyResult, err
+		}
+		// remove finalizer
+		controllerutil.RemoveFinalizer(far, v1alpha1.FARFinalizer)
+		if err := r.Client.Update(context.Background(), far); err != nil {
+			return emptyResult, fmt.Errorf("failed to remove finalizer from CR - %w", err)
+		}
+		r.Log.Info("Finalizer was removed", "CR Name", req.Name)
+		return emptyResult, nil
```
slintes, line 108: you need to handle the case with deletionTimestamp set, but no finalizer set

pkg/utils/nodes.go
```diff
@@ -5,15 +5,22 @@ import (
...
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// getNodeWithName returns a node with a name nodeName, or an error if it can't be found
+func getNodeWithName(r client.Reader, nodeName string) (*corev1.Node, error) {
+	node := &corev1.Node{}
+	key := client.ObjectKey{Name: nodeName}
+	if err := r.Get(context.TODO(), key, node); err != nil {
```
slintes, line 15: you can just return `r.Get(...)`

pkg/utils/taints.go
```diff
@@ -0,0 +1,110 @@
+package utils
+
+// Copy paste from SNR - https://github.com/medik8s/self-node-remediation/blob/main/pkg/utils/taints.go
```
slintes, line 3: not that much, is it? ;)

pkg/utils/taints.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+var (
+	loggerTaint = ctrl.Log.WithName("taints")
+)
+
+// taintExists checks if the given taint exists in list of taints. Returns true if exists false otherwise.
+func taintExists(taints []corev1.Taint, taintToFind *corev1.Taint) bool {
+	for _, taint := range taints {
+		if taint.MatchTaint(taintToFind) {
```
slintes, line 22: Heads up, this doesn't check for equal taint value!
Maybe it's a good idea to use unique taint keys per operator.

pkg/utils/taints.go
```diff
@@ -0,0 +1,110 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// deleteTaint removes all the taints that have the same key and effect to given taintToDelete.
+func deleteTaint(taints []corev1.Taint, taintToDelete *corev1.Taint) ([]corev1.Taint, bool) {
```
slintes, line 30: see comment above, this would potentially delete taints set by other medik8s operators!

pkg/utils/taints.go
```diff
@@ -0,0 +1,110 @@
...
+		return err
+	}
+
+	taint := createNoExecuteTaint()
+	// check if taint doesn't exist
+	if taintExists(node.Spec.Taints, &taint) {
+		return nil
+	}
+	// add the taint to the taint list
+	patch := client.MergeFrom(node.DeepCopy())
```
slintes, line 66: Using `patch` for updating taints is very risky unfortunately. The API server is updating taints as well while nodes get healthy/unhealthy. We saw this with SNR IIRC, when it was overwriting changes made by the API server, or the other way around. So please either use `Update` below, or `MergeFromWithOptions` here with the `OptimisticLock` option. Same for remove taint.

slintes (changes requested):
look much better now :+1:
however, some important comments inside

pkg/utils/taints.go
razo7: oops :(

pkg/utils/nodes.go
razo7: Do you mean?
```suggestion
     return node, r.Get(context.TODO(), key, node)
```

pkg/utils/taints.go
razo7: Or check for taint's value as well 👐🏻

razo7: >Maybe it's a good idea to use unique taint keys per operator.

I agree, and the `MatchTaint` func declaration supports that:
> // MatchTaint checks if the taint matches taintToMatch. Taints are unique by key:effect,
// if the two taints have same key:effect, regard as they match.

How about `"medik8s.io/far-remediation"` for FAR, and probably SNR should change it's key to include it's name, so maybe `"medik8s.io/snr-remediation"`.
CC: @mshitrit

razo7: Following this understanding and [previous review](https://github.com/medik8s/fence-agents-remediation/pull/53#discussion_r1226239684) do you still support the decision to have a value for the taint? I am less pro for having for the value now, but I don't have a strong preference.

controllers/fenceagentsremediation_controller.go
razo7: > An empty deletion timestamp just isn't an indicator for a just created resource.

How about empty finalizers?

controllers/fenceagentsremediation_controller.go
```diff
@@ -86,12 +86,12 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		return emptyResult, err
 	}
 	// Add finalizer when object is created
-	if far.ObjectMeta.DeletionTimestamp.IsZero() {
+	if ! controllerutil.ContainsFinalizer(far, v1alpha1.FARFinalizer) {
```
slintes, line 89: do you want to add the finalizer if the deletion timestamp is set? :)

pkg/utils/nodes.go
slintes: Ah sorry, somehow I thought the node is also a return value of Get()
But this should work as well.
Even though it looks a bit strange 🙈 
Leaving it up to you if you want to keep this or revert.

razo7: I will revert it 👍🏻

controllers/fenceagentsremediation_controller.go
razo7: After our discussion I see that I have neglected the scenario of `when the CR is deleted already before you reconcile for the first time` which can happen due to a deletion in a possible delay (e.g., network issues, node issues, and etc.).

controllers/fenceagentsremediation_controller.go
```diff
@@ -77,13 +78,35 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
 		if apiErrors.IsNotFound(err) {
-			// FAR is deleted, stop reconciling
+			// FAR's CR is deleted, stop reconciling
 			r.Log.Info("FAR CR is deleted - nothing to do", "CR Name", req.Name, "CR Namespace", req.Namespace)
 			return emptyResult, nil
 		}
 		r.Log.Error(err, "failed to get FAR CR")
 		return emptyResult, err
 	}
+	// Add finalizer when the CR is created
```
slintes, line 88: I think the node name validation needs to go before this?

slintes (commented):
Looks almost good to me now, just one comment inline.
Please add tests for the finalizer and taint 🙂

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -29,14 +29,17 @@ import (
...
 	"github.com/medik8s/fence-agents-remediation/api/v1alpha1"
 	"github.com/medik8s/fence-agents-remediation/pkg/utils"
 )
 
 const (
-	dummyNodeName  = "dummy-node"
-	validNodeName  = "worker-0"
+	dummyNode      = "dummy-node"
+	node01         = "worker-0"
+	node02         = "worker-1"
```
slintes, line 42: unused

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -110,29 +113,58 @@ var _ = Describe("FAR Controller", func() {
...
-		When("creating FAR CR", func() {
-			It("should build the exec command based on FAR CR", func() {
-				Eventually(func() (bool, error) {
-					res, err := cliCommandsEquality(underTestFAR)
-					return res, err
-				}, 1*time.Second, 500*time.Millisecond).Should(BeTrue())
+		When("creating valid FAR CR", func() {
+			BeforeEach(func() {
+				node = getNode(node01)
+				underTestFAR = getFenceAgentsRemediation(node01, fenceAgentIPMI, testShareParam, testNodeParam)
```
slintes, line 131: This is redundant, see outermost Describe.
And because it's the "default" `node`, maybe move `node = getNode(node01)` to the the outermost Describe as well?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -110,29 +113,58 @@ var _ = Describe("FAR Controller", func() {
...
+			})
+			It("should have finalizer and taint", func() {
+				farNamespacedName := client.ObjectKey{Name: node01, Namespace: defaultNamespace}
+				nodeNamespacedName := client.ObjectKey{Name: node01}
+				FARNoExecuteTaint := utils.CreateFARNoExecuteTaint()
```
slintes, line 136: nit: lowercase var name

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -110,29 +113,58 @@ var _ = Describe("FAR Controller", func() {
...
+				node = getNode(node01)
+				underTestFAR = getFenceAgentsRemediation(dummyNode, fenceAgentIPMI, testShareParam, testNodeParam)
+			})
+			It("should fail", func() {
+				By("Not finding a matching node to FAR CR's name")
+				nodeNamespacedName := client.ObjectKey{Name: dummyNode}
+				Expect(k8sClient.Get(context.Background(), nodeNamespacedName, node)).To(Not(Succeed()))
+				By("Not having finalizer and no taints were added")
+				farNamespacedName := client.ObjectKey{Name: dummyNode, Namespace: defaultNamespace}
+				FARNoExecuteTaint := utils.CreateFARNoExecuteTaint()
```
slintes, line 156: nit: lowercase

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -128,17 +129,15 @@ var _ = Describe("FAR Controller", func() {
...
 			It("should have finalizer and taint", func() {
 				farNamespacedName := client.ObjectKey{Name: node01, Namespace: defaultNamespace}
 				nodeNamespacedName := client.ObjectKey{Name: node01}
-				FARNoExecuteTaint := utils.CreateFARNoExecuteTaint()
 				Eventually(func() bool {
 					Expect(k8sClient.Get(context.Background(), nodeNamespacedName, node)).To(Succeed())
 					Expect(k8sClient.Get(context.Background(), farNamespacedName, underTestFAR)).To(Succeed())
 					res, _ := cliCommandsEquality(underTestFAR)
-					return controllerutil.ContainsFinalizer(underTestFAR, v1alpha1.FARFinalizer) && utils.TaintExists(node.Spec.Taints, &FARNoExecuteTaint) && res
+					return controllerutil.ContainsFinalizer(underTestFAR, v1alpha1.FARFinalizer) && utils.TaintExists(node.Spec.Taints, &farNoExecuteTaint) && res
```
slintes, line 140: nit: doing two tests (finalizer and taint) at the same time has the disadvantage that you don't know what's missing in case it fails. Better test everything on its own.

slintes (approved):
one nit inside, leaving it up to you if you want to address it.

/hold

controllers/fenceagentsremediation_controller_test.go
razo7: I agree, and I will update it in a follow up PR

PR #54, opened 2023-06-08
Support Only reboot Action

razo7:
Support only **`reboot`** action - we build the FA command with  _reboot_ action, when `--action` equals to reboot or this attribute is missing (since reboot is the [default action](https://github.com/ClusterLabs/fence-agents/blob/main/lib/fencing.py.py#L103)), otherwise we return an error and end reconcile

[ECOPROJECT-1323](https://issues.redhat.com//browse/ECOPROJECT-1323)

EDIT: Another valid action use case could be the **off** which ensure fencing, just without recovering the node. When then the admin/user will be responsible to turn the node on (or to recover the node) after a possible further investigation.

### DIFF
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -158,7 +158,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "Fence Agent response was a failure", "CR's Name", req.Name)
 		return emptyResult, err
 	}
-	if outputErr != "" || outputRes == "" {
+	if outputErr != "" || outputRes != SuccessFAResponse {
 		// response wasn't failure or sucesss message
 		err := fmt.Errorf("unknown fence agent response - expecting `%s` response, but we received `%s`", SuccessFAResponse, outputRes)
 		r.Log.Error(err, "Fence Agent response wasn't a success message", "CR's Name", req.Name)
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -170,19 +170,19 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 // buildFenceAgentParams collects the FAR's parameters for the node based on FAR CR, and if the CR is missing parameters
 // or the CR's name don't match nodeParamter name then return an error
 func buildFenceAgentParams(far *v1alpha1.FenceAgentsRemediation) ([]string, error) {
+	logger := ctrl.Log.WithName("build-fa-parameters")
 	if far.Spec.NodeParameters == nil || far.Spec.SharedParameters == nil {
-		return nil, errors.New(errorMissingParams)
+		err := errors.New(errorMissingParams)
+		logger.Error(err, "Missing parameters")
+		return nil, err
 	}
 	var fenceAgentParams []string
-	logger := ctrl.Log.WithName("build-fa-parameters")
 	// add shared parameters except the action parameter
 	for paramName, paramVal := range far.Spec.SharedParameters {
-		if paramName == parameterActionName {
-			if paramVal != parameterActionValue {
-				logger.Info("FAR doesn't support any other action than reboot", "action", paramVal)
-			}
-		} else {
+		if paramName != parameterActionName {
 			fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, paramVal)
+		} else if paramVal != parameterActionValue {
+			logger.Info("FAR doesn't support any other action than reboot", "action", paramVal)
 		}
 	}
 	// ensure the FA uses the reboot action
@@ -195,6 +195,7 @@ func buildFenceAgentParams(far *v1alpha1.FenceAgentsRemediation) ([]string, erro
 			fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, nodeVal)
 		} else {
 			err := errors.New(errorMissingNodeParams)
+			logger.Error(err, "Missing matching nodeParam and CR's name")
 			return nil, err
 		}
 	}
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -38,6 +38,8 @@ const (
 	errorMissingParams     = "nodeParameters or sharedParameters or both are missing, and they cannot be empty"
 	errorMissingNodeParams = "node parameter is required, and cannot be empty"
 	SuccessFAResponse      = "Success: Rebooted"
+	parameterActionName    = "--action"
+	parameterActionValue   = "reboot"
 )
 
 // FenceAgentsRemediationReconciler reconciles a FenceAgentsRemediation object
@@ -172,10 +174,21 @@ func buildFenceAgentParams(far *v1alpha1.FenceAgentsRemediation) ([]string, erro
 		return nil, errors.New(errorMissingParams)
 	}
 	var fenceAgentParams []string
+	logger := ctrl.Log.WithName("build-fa-parameters")
+	// add shared parameters except the action parameter
 	for paramName, paramVal := range far.Spec.SharedParameters {
-		fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, paramVal)
+		if paramName == parameterActionName {
+			if paramVal != parameterActionValue {
+				logger.Info("FAR doesn't support any other action than reboot", "action", paramVal)
+			}
+		} else {
+			fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, paramVal)
+		}
 	}
+	// ensure the FA uses the reboot action
+	fenceAgentParams = appendParamToSlice(fenceAgentParams, parameterActionName, parameterActionValue)
 
+	// append node parameters
 	nodeName := v1alpha1.NodeName(far.Name)
 	for paramName, nodeMap := range far.Spec.NodeParameters {
 		if nodeVal, isFound := nodeMap[nodeName]; isFound {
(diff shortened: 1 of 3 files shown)


razo7:
E2E test is expected to fail as we don't support `reboot` action in E2E. Hopefully with the merge of #20 the E2E test could be tested.
/hold

controllers/fenceagentsremediation_controller.go
```diff
@@ -48,6 +48,11 @@ func (r *FenceAgentsRemediationReconciler) SetupWithManager(mgr ctrl.Manager) er
 		Complete(r)
 }
 
+const (
+	methodNameFA      = "--action"
```
slintes, line 52: can we have more descriptive names please? e.g. `parameterNameAction` and `parameterValueActionReboot`?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -70,6 +77,17 @@ var _ = Describe("FAR Controller", func() {
 
 	Context("Functionality", func() {
 		Context("buildFenceAgentParams", func() {
+			When("FAR include different action than reboot", func() {
+				It("should succeed with a warning", func() {
+					invalidValTestFAR := newFenceAgentsRemediation(validNodeName, fenceAgentIPMI, invalidShareParam, testNodeParam)
+					_, invalidShareString, err := buildFenceAgentParams(invalidValTestFAR)
```
slintes, line 83: this does not even compile...

slintes (changes requested):

controllers/fenceagentsremediation_controller.go
```diff
@@ -38,6 +38,8 @@ const (
 	errorMissingParams     = "nodeParameters or sharedParameters or both are missing, and they cannot be empty"
 	errorMissingNodeParams = "node parameter is required, and cannot be empty"
 	SuccessFAResponse      = "Success: Rebooted"
```
mshitrit, line 40: I think this could be made private

razo7: ATM it can be. But it might make more sense keep it public and use it in https://github.com/medik8s/fence-agents-remediation/blob/main/test/e2e/far_e2e_test.go#L119, instead of creating another const  `succeesRebootMessage`

controllers/fenceagentsremediation_controller.go
```diff
@@ -168,10 +170,21 @@ func buildFenceAgentParams(far *v1alpha1.FenceAgentsRemediation) ([]string, erro
 		return nil, errors.New(errorMissingParams)
```
mshitrit, line 170: Nit: I suggest writing an error to the log before returning the error

controllers/fenceagentsremediation_controller.go
```diff
@@ -168,10 +170,21 @@ func buildFenceAgentParams(far *v1alpha1.FenceAgentsRemediation) ([]string, erro
...
+	// add shared parameters except the action parameter
 	for paramName, paramVal := range far.Spec.SharedParameters {
-		fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, paramVal)
+		if paramName == parameterActionName {
+			if paramVal != parameterActionValue {
+				logger.Info("FAR doesn't support any other action than reboot", "action", paramVal)
+			}
+		} else {
+			fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, paramVal)
+		}
```
mshitrit, line 182: Nit: I think this can be simplified a bit into:
```
if paramName != parameterActionName {
	fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, paramVal)
} else if paramVal != parameterActionValue {
	logger.Info("FAR doesn't support any other action than reboot", "action", paramVal)
}
```

controllers/fenceagentsremediation_controller.go
```diff
@@ -152,7 +154,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "Fence Agent response was a failure", "CR's Name", req.Name)
 		return emptyResult, err
 	}
-	if outputErr != "" || outputRes == "" {
+	if outputErr != "" || outputRes != SuccessFAResponse {
```
slintes, line 157: I think we discussed this before, can't find it though: are we sure this is the response by all fence agents?
Can't we assume that `err == nil && outputErr == ""` mean success, no matter what `outputRes` is?

controllers/fenceagentsremediation_controller.go
```diff
@@ -168,10 +170,21 @@ func buildFenceAgentParams(far *v1alpha1.FenceAgentsRemediation) ([]string, erro
...
+		if paramName == parameterActionName {
+			if paramVal != parameterActionValue {
+				logger.Info("FAR doesn't support any other action than reboot", "action", paramVal)
+			}
+		} else {
+			fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, paramVal)
+		}
 	}
+	// ensure the FA uses the reboot action
+	fenceAgentParams = appendParamToSlice(fenceAgentParams, parameterActionName, parameterActionValue)
```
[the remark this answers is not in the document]
slintes: - is the action name for triggering a reboot the same for all fence agents?
- is reboot the only action which makes sense for us? What about shutdown / switch off? That would ensure fencing, just without the chance to recover. Valid usecase IMHO, when users don't want to rejoin unhealthy nodes but investigate it manually instead 🤔 

> By doing this we make sure a reboot will occur regardless of the user input

agree, that's not a good idea

> I can think of using status maybe as a method to check the node's status after the FA execution return success

I do not understand? 🤔 

> so we can be sure that the node status is back to Ready.

That might never happen, the node can have a permanent issue which isn't fixed by a reboot

controllers/fenceagentsremediation_controller.go
razo7: We can, but I think it is more safe to look for this [success message](https://github.com/ClusterLabs/fence-agents/blob/main/lib/fencing.py.py#L1045) or the [return code zero](https://github.com/ClusterLabs/fence-agents/blob/main/lib/fencing.py.py#L1063).

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
razo7: >is the action name for triggering a reboot the same for all fence agents?

Yes, every FA which uses the [fence_action function](https://github.com/ClusterLabs/fence-agents/blob/main/lib/fencing.py.py#L939-L1064) so all the FA have **reboot** as the action name to trigger **off** and then **on** for a node.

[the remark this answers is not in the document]
clobrano: the logic needed to exclude any action except reboot is a bit unclear to me. Just as idea, would it be possible to validate the spec with a webhook instead? I guess this way it would be clear since the beginning what's accepted and what's not.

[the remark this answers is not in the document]
razo7: > What about shutdown / switch off? That would ensure fencing, just without the chance to recover. Valid usecase IMHO, when users don't want to rejoin unhealthy nodes but investigate it manually instead

Good point. I prefer to only enable `reboot` for now, and for next release support the `off` action which is indeed a valid use case.

controllers/fenceagentsremediation_controller.go
razo7: But we do is it in https://github.com/medik8s/fence-agents-remediation/pull/54/files#diff-b9729ac4458c7637438e3bf558999b35adedcaf4d13af1e3eb752db501fd4997R119

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
razo7: > would it be possible to validate the spec with a webhook instead?

I am unsure if it is possible, since the action is not an independent field in spec, and it is part of the [SharedParameters](https://github.com/medik8s/fence-agents-remediation/blob/main/api/v1alpha1/fenceagentsremediation_types.go#L42-L44) . But it does worth thinking to maybe separate it, and then use webhook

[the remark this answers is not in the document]
clobrano: I think you could check it if and only if it is set

controllers/fenceagentsremediation_controller.go
slintes: A non zero exit code will result in a non nil outputErr 🙂 
But also nice to see that all agents share the relevant code with the success message, thx 👍🏼

razo7: YW it was a good question

controllers/fenceagentsremediation_controller.go
```diff
@@ -182,10 +182,15 @@ func buildFenceAgentParams(far *v1alpha1.FenceAgentsRemediation) ([]string, erro
 		if paramName != parameterActionName {
 			fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, paramVal)
 		} else if paramVal != parameterActionValue {
-			logger.Info("FAR doesn't support any other action than reboot", "action", paramVal)
+			// --action attribute was selected but it is differemt than reboot
+			err := errors.New("FAR doesn't support any other action than reboot")
+			logger.Error(err, "can't build CR with this action attribute", "action", paramVal)
+			return nil, err
```
mshitrit, line 188: I think it's fine returning nil here since I don't see how this `error` will be resolved.

razo7: So you are suggesting to not return error and avoid exponential back off errors by reporting an error only once from the above line (187)?

mshitrit: Yep, IMO it's enough to just log it.

mshitrit: Correction: I wasn't aware that we aren't in reconcile context, so returning the error here is find, but I wouldn't return it in the Reconcile method

slintes (approved):

PR #55, opened 2023-06-08
Fence Agent Output Check

razo7:
1. Don't log (and expose) FA command - It might include privileged credentials that an attacker can manipulate and abuse.
2. Check fence agent CLI output - Error and expected success output.
3. Small improvement to UT check - https://github.com/medik8s/fence-agents-remediation/pull/48#discussion_r1222721600
4. Add log.error in Reconcile for better identification of error's source/root.

[ECOPROJECT-1329](https://issues.redhat.com//browse/ECOPROJECT-1329)

### DIFF
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -34,8 +34,9 @@ import (
 )
 
 const (
-	errorBuildingFAParams = "node parameter is required, and cannot be empty"
-	SuccessFAResponse = "Success: Rebooted"
+	errorMissingParams     = "nodeParameters or sharedParameters or both are missing, and they cannot be empty"
+	errorMissingNodeParams = "node parameter is required, and cannot be empty"
+	SuccessFAResponse      = "Success: Rebooted"
 )
 
 // FenceAgentsRemediationReconciler reconciles a FenceAgentsRemediation object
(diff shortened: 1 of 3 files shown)


pkg/cli/cliexecuter.go
```diff
@@ -44,7 +44,7 @@ func NewExecuter(config *restclient.Config) (Executer, error) {
 }
 
 // Execute builds and runs a Post request on contianer for SPDY (shell) executor
-func (e executer) Execute(pod *corev1.Pod, command []string) (stdout string, stderr string, err error) {
+func (e executer) Execute(pod *corev1.Pod, nodeParams []string, command []string) (stdout string, stderr string, err error) {
```
clobrano, line 47: Is the `nodeParams` parameter included in the function solely for debugging purposes, right?
I think it is better to avoid using arguments that the function doesn't actually require. Instead, you could consider logging the error along with the `nodeParams` content in the caller function.
WDYT?

controllers/fenceagentsremediation_controller.go
```diff
@@ -103,22 +107,31 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
 	if err != nil {
 		return emptyResult, err
 	}
 	cmd := append([]string{far.Spec.Agent}, faParams...)
 	// The Fence Agent is excutable and the parameters are valid but we don't know about their values
-	if _, _, err := r.Executor.Execute(pod, cmd); err != nil {
+	outputRes, outputErr, err := r.Executor.Execute(pod, nodeParams, cmd)
+	// Check if not error returned and a successful message was returned
+	if err != nil {
 		//TODO: better seperation between errors from wrong shared parameters values and wrong node parameters values
```
clobrano, line 119: not belonging to this change, but there is a typo here `s/seperation/separation`

pkg/cli/cliexecuter.go
razo7: > Is the nodeParams parameter included in the function solely for debugging purposes, right?

Correct, I do want that parameter for differentiating between log outputs of different FAR CRs (and most importantly success log output that I can look for in FAR e2e tests, see function **[buildExpectedLogOutput](https://github.com/medik8s/fence-agents-remediation/pull/20/files#diff-b9729ac4458c7637438e3bf558999b35adedcaf4d13af1e3eb752db501fd4997R266-R277)**).

razo7: > I think it is better to avoid using arguments that the function doesn't actually require.

👍🏻 

> Instead, you could consider logging the error along with the nodeParams content in the caller function.

The caller function is [`reconcile`](https://github.com/medik8s/fence-agents-remediation/pull/55/files#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R116), so IIUC you are suggesting to  log _nodeParams_  in reconcile after we call [execute](https://github.com/medik8s/fence-agents-remediation/pull/55/files#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R116)? Sounds better, but I believe we need to do more changes for that. ATM in reconcile we print on failure result or on an error, and I would like to have the nodeParams on success situation as well.

Therefore I see three options (could be more):

1. Don't log in `pkg/cli/cliexecuter.go`-  just log on caller function. This why we don't share the `nodeParams` parameter, and we don't log twice.
2. Don't share the `nodeParams` parameter to `pkg/cli/cliexecuter.go`, and we use the parameter for logging in reconcile. We end up with two logs stating the same thing, just in a different manner.
3. Keep it as it is now, and don't add more changes. One log output about CLI execution, but with the minimal cost of sharing a parameter.

controllers/fenceagentsremediation_controller.go
```diff
@@ -103,22 +107,31 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
 	cmd := append([]string{far.Spec.Agent}, faParams...)
 	// The Fence Agent is excutable and the parameters are valid but we don't know about their values
-	if _, _, err := r.Executor.Execute(pod, cmd); err != nil {
-		//TODO: better seperation between errors from wrong shared parameters values and wrong node parameters values
+	outputRes, outputErr, err := r.Executor.Execute(pod, nodeParams, cmd)
+	// Check if not error returned and a successful message was returned
+	if err != nil {
+		//TODO: better separation between errors from wrong shared parameters values and wrong node parameters values
+		if outputErr != "" {
+			err = fmt.Errorf(outputErr+"- %w", err)
```
mshitrit, line 121: Nit: since formatting is already used I suggest dropping  the "+" 
```
fmt.Errorf("%s- %w", outputErr,  err)
```

controllers/fenceagentsremediation_controller.go
```diff
@@ -103,22 +107,31 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
+		}
 		return emptyResult, err
 	}
+	if outputRes != successFAMessages {
+		errMes := fmt.Sprintf("unknown fence agent response - expecting a \"%s\" response", successFAMessages)
```
mshitrit, line 126: Nit: might be useful to report the actual response that was received (i.e `outputRes`)

pkg/cli/cliexecuter.go
razo7: Thinking more about it. Would it make more sense to not modify _buildFenceAgentParams_ function, and share the node's name rather than `nodeParams`?. Both of them are node identifiers....

razo7: Since the idea of adding this parameter is for differentiating between log outputs of different FAR CRs and ATM there is only one E2E test, so I will dismiss this change, and handle it in the future.

clobrano: Sorry for replying so late :facepalm: 

> Thinking more about it. Would it make more sense to not modify buildFenceAgentParams function, and share the node's name rather than nodeParams?. Both of them are node identifiers

Do you only want to log a Node identifier? If that so, is it enough the new log just above calling `.Executer`?

controllers/fenceagentsremediation_controller.go
```diff
@@ -77,47 +79,68 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
 	cmd := append([]string{far.Spec.Agent}, faParams...)
-	// The Fence Agent is excutable and the parameters are valid but we don't know about their values
-	if _, _, err := r.Executor.Execute(pod, cmd); err != nil {
-		//TODO: better seperation between errors from wrong shared parameters values and wrong node parameters values
+	// The Fence Agent is excutable and the parameters structure are valid, but we don't check their values
+	r.Log.Info("Execute the fence agent", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
+	outputRes, outputErr, err := r.Executor.Execute(pod, cmd)
+	if err != nil {
+		// response was a failure message
+		if outputErr != "" {
```
slintes, line 123: this is already logged in cliexecutor

controllers/fenceagentsremediation_controller.go
```diff
@@ -77,47 +79,68 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
+			err = fmt.Errorf("%s- %w", outputErr, err)
+		}
+		r.Log.Error(err, "Fence Agent response was a failure", "CR's Name", req.Name)
+		return emptyResult, err
+	}
+	if outputRes != SuccessFAResponse {
```
slintes, line 129: this looks risky to me. Are you sure that this will be the output by all fence-agent variations? I would expect that the command succeeded when err == nil && outputErr == ""

pkg/cli/cliexecuter.go
clobrano: I understand. In any case, since the [log before calling `Executer.Execute`](https://github.com/medik8s/fence-agents-remediation/pull/55/files#diff-f2169f37331a57117047ad17f0b9bac545f2606f32eadce37a69fedb2b92cc61) has everything `Executer.Execute` has (except `stdout`), consider logging there.

controllers/fenceagentsremediation_controller.go
razo7: Good catch as it would break the current E2E test since it uses status (and status has a different success message).
This logic suits better to https://github.com/medik8s/fence-agents-remediation/pull/54.

> Are you sure that this will be the output by all fence-agent variations?

I can verify it but I have got the impression that any FA from ClusterLabs has [specific actions](https://github.com/ClusterLabs/fence-agents/blob/main/doc/FenceAgentAPI.md#agent-operations-and-return-values) that it supports, and thus the output should be formatted the same. But I could be wrong, as I only tested on two FA so far.

> I would expect that the command succeeded when err == nil && outputErr == ""
 
I will use this term for now, but if my above impression/assumption is correct than it makes sense to restrict the the term that we check. In the future when we might want to support new and not FA from ClusterLabs, then we could ease this restriction...

pkg/cli/cliexecuter.go
razo7: But what value will it add? Since I want to easily search (by reviewing in log and in E2E) for a substring from log that has stdout and nodeIdentifier together.

clobrano: you could log *after*, when you also have `stdout`

razo7: > you could log after, when you also have stdout

Sure, I am aware of that. I just don't know which option is better from above. I only understand that 3 is not recommended.

clobrano: I'd say option 1 looks good

slintes:
lgtm from my side

clobrano:
/lgtm from my side too
giving to the others time to resolve any other open discussion though
/hold

PR #59, opened 2023-06-21
Improving Unit Test Logic and Time

razo7:
1. Separate logical unit-test and order- Test _taint_ and _finalizer_ existence in eventually following an expect rather than one eventually with and a logical OR/AND of there existence. This way we better check the order of their creation. Related to [older comment](https://github.com/medik8s/fence-agents-remediation/pull/53#discussion_r1234149421).
2. Reduce eventually timeout and polling interval- the values were too high

### DIFF
--- a/controllers/fenceagentsremediation_controller_test.go
+++ b/controllers/fenceagentsremediation_controller_test.go
@@ -110,6 +110,8 @@ var _ = Describe("FAR Controller", func() {
 		})
 	})
 	Context("Reconcile", func() {
+		farNamespacedName := client.ObjectKey{Name: node01, Namespace: defaultNamespace}
+		nodeKey := client.ObjectKey{Name: node01}
 		farNoExecuteTaint := utils.CreateFARNoExecuteTaint()
 		//Scenarios
 		BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


controllers/fenceagentsremediation_controller_test.go
```diff
@@ -133,29 +133,37 @@ var _ = Describe("FAR Controller", func() {
 			It("should have finalizer and taint", func() {
 				farNamespacedName := client.ObjectKey{Name: node01, Namespace: defaultNamespace}
 				nodeNamespacedName := client.ObjectKey{Name: node01}
```
mshitrit, line 135: This is a bit confusing, since namespace isn't relevant  in the context of nodes

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -133,29 +133,37 @@ var _ = Describe("FAR Controller", func() {
...
+				By("Searching for remediation taint")
 				Eventually(func() bool {
 					Expect(k8sClient.Get(context.Background(), nodeNamespacedName, node)).To(Succeed())
```
mshitrit, line 138: I think that checking for the node existence, which is created in the test code (and not in production code) is redundant.
IMO it's the equivalent of:
```
It("some test", func() {
  someVar := true
  Expect(someVar).To(BeTrue())
})
```

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -133,29 +133,37 @@ var _ = Describe("FAR Controller", func() {
...
 					Expect(k8sClient.Get(context.Background(), farNamespacedName, underTestFAR)).To(Succeed())
 					res, _ := cliCommandsEquality(underTestFAR)
-					return controllerutil.ContainsFinalizer(underTestFAR, v1alpha1.FARFinalizer) && utils.TaintExists(node.Spec.Taints, &farNoExecuteTaint) && res
-				}, 1*time.Second, 500*time.Millisecond).Should(BeTrue(), "finalizer and taint should be added, and command format is correct")
+					return utils.TaintExists(node.Spec.Taints, &farNoExecuteTaint) && res
+				}, 20*time.Millisecond, 10*time.Millisecond).Should(BeTrue(), "taint should be added, and command format is correct")
```
mshitrit, line 142: Nit: retry of every 10ms over a period of 20ms seems like cutting it really close, which in part defeating the use of `Eventually`, maybe use a longer period ? (first param)

controllers/fenceagentsremediation_controller_test.go
slintes: I usually use something like `nodeKey`: it represents the type (ObjectKey) and it doesn't matter if it has a namespace or not 🙂

controllers/fenceagentsremediation_controller_test.go
razo7: I will change to 100ms

controllers/fenceagentsremediation_controller_test.go
razo7: On the same principle I can ignore checking the CR creation, right?

razo7: Even though that I don't agree that I can ignore this line, because without this `k8sClient.Get` **node** variable won't be updated with the node from reconcile with taints that we check on line 141...

mshitrit: Good point, definitely need to check the taint on the node.

PR #56, opened 2023-06-12
Update Tools and API Versions

razo7:
Bump FAR to the latest tool and API versions (e.g. K8s, OCP, Controller-gen, Kustomize, and Operator-SDK).

In addition it also includes more stuff:

- Fix for writing/deleting the `envtest` target directory
- Fetching the machine client (for E2E) using OpenShift client-go

[ECOPROJECT-1403](https://issues.redhat.com//browse/ECOPROJECT-1403)

PR #61, opened 2023-06-29
Validate utils Package and Taint Functionality

razo7:
- Add one unit test for taint functionality on control-plane node with adding and removing FAR NoExecute taint
- Use worker label from Commons rather than declaring it in FAR's repoistory
- Run `make go-verify`
- Create new suite test and UT file for utils package, then move out the 3 UT for Utils functionality to utils_test, and update makefile to test the new directory.

### DIFF
--- a/controllers/fenceagentsremediation_controller_test.go
+++ b/controllers/fenceagentsremediation_controller_test.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	medik8sLabels "github.com/medik8s/common/pkg/labels"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
@@ -72,6 +73,10 @@ var _ = Describe("FAR Controller", func() {
 
 	// default FenceAgentsRemediation CR
 	underTestFAR := getFenceAgentsRemediation(node01, fenceAgentIPMI, testShareParam, testNodeParam)
+	nodeKey := client.ObjectKey{Name: node01}
+	farNamespacedName := client.ObjectKey{Name: node01, Namespace: defaultNamespace}
+	controlPlaneRoleTaint := getControlPlaneRoleTaint()
+	farNoExecuteTaint := utils.CreateFARNoExecuteTaint()
 
 	Context("Functionality", func() {
 		Context("buildFenceAgentParams", func() {
@@ -93,7 +98,7 @@ var _ = Describe("FAR Controller", func() {
 
 		Context("IsNodeNameValid", func() {
 			BeforeEach(func() {
-				node = getNode(node01)
+				node = getNode("", node01)
 				DeferCleanup(k8sClient.Delete, context.Background(), node)
 				Expect(k8sClient.Create(context.Background(), node)).To(Succeed())
 			})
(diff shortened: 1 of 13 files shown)


razo7:
Tests will fail since the PR uses [common repo](https://github.com/medik8s/fence-agents-remediation/pull/61/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R7) which depends on [newer Kubernetes API](https://github.com/medik8s/common/blob/main/go.mod), thus #56 should be merged, and then after rebasing the PR could be tested.

mshitrit:
I think at the moment go.mod/go.sum/vendor aren't in sync.
This can be resolved by `go mod tidy/vendor`.
However once done it will uncover a more serious[ issue](https://github.com/medik8s/self-node-remediation/pull/131#issuecomment-1616603651) which I've also encountered in a separate PR.

pkg/utils/nodes.go
```diff
@@ -33,3 +34,29 @@ func IsNodeNameValid(r client.Reader, nodeName string) (bool, error) {
...
+			},
+		}
+	} else {
+		return &corev1.Node{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: nodeName,
+			},
+		}
+	}
+}
```
razo7, line 62: I think it makes more sense to move this function which is used for unit-testing to another file. Maybe under [`test`](https://github.com/medik8s/fence-agents-remediation/tree/main/test)?

razo7: Or even having two local function for getting the node object? One in `fenceagentsremediation_controller_test.go` , and one in `utils_test.go`

pkg/utils/utils_suite_test.go
```diff
@@ -0,0 +1,99 @@
+/*
```
slintes, line 1: filename: you can skip "utils", it's redundant (there should be one suite per package)

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,78 @@
+package utils
```
slintes, line 1: nit: I would name this file nodes_test.go,  so you can have separate test files when you add new util files later on.

pkg/utils/nodes.go
slintes: Good point. I would put it into pkg/utils/nodes_test.go

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -132,7 +114,7 @@ var _ = Describe("FAR Controller", func() {
 
 		When("creating valid FAR CR", func() {
 			BeforeEach(func() {
-				node = getNode(node01)
+				node = utils.GetNode("", node01)
```
mshitrit, line 117: IIUC node is always initialized to the same value.
If this is indeed the case I suggest:
- initialize node value on var declaration
- move the node creation to a share BeforeEach block

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+var _ = Describe("Utils", func() {
+	var node *corev1.Node
+	nodeKey := client.ObjectKey{Name: node01}
+	controlPlaneRoleTaint := getControlPlaneRoleTaint()
+	farNoExecuteTaint := CreateFARNoExecuteTaint()
+	Context("FAR CR and Node Names Validity test", func() {
+		BeforeEach(func() {
+			node = GetNode("", node01)
+			DeferCleanup(k8sClient.Delete, context.Background(), node)
```
mshitrit, line 27: Nit: defer should be after the Create

pkg/utils/utils_test.go
razo7: But nodes_test.go wouldn't be interfered as the test file for nodes.go, while this file shares a test for the taint file as well? ATM it has tests for two utils files, `taint.go`, and `nodes.go`
Going with your direction maybe it would make more sense to move `Context("Taint functioninality test",...` logic to `taint_test.go`, and have `nodes_test.go` for the other tests that utils.go has. 
WDYT?

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,78 @@
...
+			node = GetNode("control-plane", node01)
+			Expect(k8sClient.Create(context.Background(), node)).To(Succeed())
+			DeferCleanup(k8sClient.Delete, context.Background(), node)
+		})
+		When("Control-plane node only has 1 the control-plane-role taint", func() {
+			It("should add and delete medik8s NoSchedule taint and keep other existing taints", func() {
+				By("having one control-plane-role taint")
+				taintedNode := &corev1.Node{}
+				Expect(k8sClient.Get(context.Background(), nodeKey, taintedNode)).To(Succeed())
+				Expect(TaintExists(taintedNode.Spec.Taints, &controlPlaneRoleTaint)).To(BeTrue())
```
mshitrit, line 53: IMO this is redundant since we added  this taint in the test code

controllers/fenceagentsremediation_controller_test.go
razo7: > IIUC node is always initialized to the same value.

Correct, but if it is initialized once then the second test would fail as the instance might be deleted or already created while the second test tries to create it (an error due race condition). To overcome it we initialize an independent instance of the node (same spec and name, but different **resource version**) in a different BeforeEach block.

pkg/utils/nodes.go
razo7: But then `controllers/fenceagentsremediation_controller_test.go` could not find the `GetNode` function from a test code `pkg/utils/nodes_test.go`.

pkg/utils/utils_test.go
slintes: yes, ideally every xyz.go file should have its own xyz_test.go file. But I won't block on this, for these few methods / tests

razo7: Sounds good to me :)

razo7:
I didn't though that `/retest` will trigger all the jobs, since the PR is still considered as draft... But now I know :D

pkg/utils/nodes.go
```diff
@@ -33,3 +34,29 @@ func IsNodeNameValid(r client.Reader, nodeName string) (bool, error) {
 	}
 	return true, nil
 }
+
+// used for making new node object for test and have a unique resourceVersion
```
slintes, line 38: comments should start with the function name, in the 1st line of the comment 😉

pkg/utils/nodes.go
```diff
@@ -33,3 +34,29 @@ func IsNodeNameValid(r client.Reader, nodeName string) (bool, error) {
...
+// GetNode returns a node object with the name nodeName based on the nodeType input
+func GetNode(nodeType, nodeName string) *corev1.Node {
```
slintes, line 40: IMHO nodeRole wold be a better name than nodeType?

slintes (commented):
two nits, other than that lgtm

pkg/utils/nodes.go
razo7: I agree

PR #62, opened 2023-07-02
Two Remediations and Taint Addition for E2E

razo7:
- Reorganize FAR E2E test code for testing two FAR creation, one following another, to two different nodes.
- Test the addition of FAR taint prior to checking the pod's logs and the node's boot time

### DIFF
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -41,26 +41,50 @@ const (
 
 var _ = Describe("FAR E2e", func() {
 	var (
-		far             *v1alpha1.FenceAgentsRemediation
-		fenceAgent      string
-		clusterPlatform *configv1.Infrastructure
-		err             error
+		fenceAgent, nodeIdentifierPrefix string
+		testShareParam                   map[v1alpha1.ParameterName]string
+		testNodeParam                    map[v1alpha1.ParameterName]map[v1alpha1.NodeName]string
+		nodeIndex                        int
+		secondRun                        bool
 	)
 	BeforeEach(func() {
-		clusterPlatform, err = e2eUtils.GetClusterInfo(configClient)
+		// create FAR CR spec based on OCP platformn
+		clusterPlatform, err := e2eUtils.GetClusterInfo(configClient)
 		if err != nil {
 			Fail("can't identify the cluster platform")
 		}
 		fmt.Printf("\ncluster name: %s and PlatformType: %s \n", string(clusterPlatform.Name), string(clusterPlatform.Status.PlatformStatus.Type))
+
+		switch clusterPlatform.Status.PlatformStatus.Type {
+		case configv1.AWSPlatformType:
+			fenceAgent = fenceAgentAWS
+			nodeIdentifierPrefix = nodeIdentifierPrefixAWS
+			By("running fence_aws")
+		case configv1.BareMetalPlatformType:
+			fenceAgent = fenceAgentIPMI
+			nodeIdentifierPrefix = nodeIdentifierPrefixIPMI
+			By("running fence_ipmilan")
+		default:
+			Skip("FAR haven't been tested on this kind of cluster (non AWS or BareMetal)")
+		}
+
+		testShareParam, err = buildSharedParameters(clusterPlatform, fenceAgentAction)
+		if err != nil {
+			Fail("can't get shared information")
+		}
+		testNodeParam, err = buildNodeParameters(clusterPlatform.Status.PlatformStatus.Type)
+		if err != nil {
+			Fail("can't get node information")
+		}
+		// run FA on the first worker node
+		nodeIndex = 0
 	})
 
-	Context("fence agent - fence_aws or fence_ipmilan", func() {
+	Context("stress cluster", func() {
 		var (
-			nodeBootTimeBefore   time.Time
-			errBoot              error
-			testNodeName         string
-			nodeIdentifierPrefix string
-			testNodeID           string
+			nodeName, errString string
+			nodeBootTimeBefore  time.Time
+			err                 error
 		)
 		BeforeEach(func() {
 			nodes := &corev1.NodeList{}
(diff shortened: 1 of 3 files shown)


test/e2e/far_e2e_test.go
```diff
@@ -71,61 +95,57 @@ var _ = Describe("FAR E2e", func() {
...
-				nodeIdentifierPrefix = nodeIdentifierPrefixIPMI
-				By("running fence_ipmilan")
-			default:
-				Skip("FAR haven't been tested on this kind of cluster (non AWS or BareMetal)")
-			}
-
-			testShareParam, err := buildSharedParameters(clusterPlatform, fenceAgentAction)
-			if err != nil {
-				Fail("can't get shared information")
+			if secondRun {
```
clobrano, line 98: In general it would be good to keep tests independent from each other. You should be able to run 1 test without the others, or just invert the order.

mshitrit: I agree.
just to be clear: it's fine to have a test that checks creation of two remeditions  - just do it in a single test (i.e create both remediation in that test scope)  instead of relying on another test to create the first one.

test/e2e/far_e2e_test.go
```diff
@@ -41,26 +41,50 @@ const (
...
+			fenceAgent = fenceAgentIPMI
+			nodeIdentifierPrefix = nodeIdentifierPrefixIPMI
+			By("running fence_ipmilan")
+		default:
+			Skip("FAR haven't been tested on this kind of cluster (non AWS or BareMetal)")
+		}
+
+		testShareParam, err = buildSharedParameters(clusterPlatform, fenceAgentAction)
+		if err != nil {
+			Fail("can't get shared information")
```
slintes, line 73: Is there a reason for this `if err != nil {Fail(msg)}` pattern?
We usually use `Expect(err).ToNot(HaveOccured(), msg)`

test/e2e/far_e2e_test.go
```diff
@@ -71,61 +95,57 @@ var _ = Describe("FAR E2e", func() {
...
-
-				By("checking the node's boot time after running the FA")
-				wasNodeRebooted(testNodeName, nodeBootTimeBefore)
 			})
 		})
 	})
 })
 
+// getNodeName returns the node's name based on valid index, otherwise it returns an error
+func getNodeName(index int) (string, string) {
```
slintes, line 134: Are you sure that nodes are returned in same order always, so that a different index always results in a different node?

slintes: > otherwise it returns an error

I don't see an error, just 2 strings 🤷🏼‍♂️

test/e2e/far_e2e_test.go
```diff
@@ -277,3 +310,18 @@ func wasNodeRebooted(nodeName string, nodeBootTimeBefore time.Time) {
 
 	log.Info("successful reboot", "node", nodeName, "offset between last boot", nodeBootTimeAfter.Sub(nodeBootTimeBefore), "new boot time", nodeBootTimeAfter)
 }
+
+// remediateNode run three functions to verify whether the node was remediated
```
slintes, line 314: comments are great, but what's the value for saying that we run three functions? Wouldn't it be more useful to say what they are verifying?

test/e2e/far_e2e_test.go
```diff
@@ -277,3 +310,18 @@ func wasNodeRebooted(nodeName string, nodeBootTimeBefore time.Time) {
...
+func remediateNode(nodeName, logString string, nodeBootTimeBefore time.Time) {
```
slintes, line 315: misleading name, I don't think we remediate a node here?

test/e2e/far_e2e_test.go
```diff
@@ -277,3 +310,18 @@ func wasNodeRebooted(nodeName string, nodeBootTimeBefore time.Time) {
...
+	By("Adding FAR NoExecute taint")
```
slintes, line 316: misleading comment, as far as I can tell we do not add the taint, but check it was added?

test/e2e/far_e2e_test.go
```diff
@@ -277,3 +310,18 @@ func wasNodeRebooted(nodeName string, nodeBootTimeBefore time.Time) {
...
+	wasFarTaintAdded(nodeName)
+
+	By("Executing the FA command and receive success response")
```
slintes, line 319: same

test/e2e/far_e2e_test.go
```diff
@@ -277,3 +310,18 @@ func wasNodeRebooted(nodeName string, nodeBootTimeBefore time.Time) {
...
+	// TODO: When reboot is running only once and it is running on FAR node, then FAR pod will
```
slintes, line 320: what's the plan for this TODO? This sounds like this test will fail regularly?

test/e2e/far_e2e_test.go
slintes: IMHO two tests are fine, however the 2nd one should not depend on the 1st for indeed. Can be resolved by using a BeforeEach() for the 2nd test to switch nodes.

slintes (changes requested):

test/e2e/far_e2e_test.go
razo7: >When reboot is running only once

ATM reconcile can be triggered for some reasons, e.g. when far pod was evicted.
When it is evicted, e.g. from FAR taint, then it will run the reboot again and then a success message should appear. But when we will resolve this problem (with status), then we should skip the check for the success message when we taint the node which had the FAR pod.

test/e2e/far_e2e_test.go
```diff
@@ -238,6 +201,57 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
 	return testNodeParam, nil
 }
 
+// randomizeWorkerNode returns a worker node name which is different than the previous one
+// (on the first call it will allways return new node)
+func randomizeWorkerNode(nodes *corev1.NodeList) string {
+	nodeName := previousNodeName
+	for previousNodeName == nodeName {
+		// Generate a random seed based on the current time
+		rand.New(rand.NewSource(time.Now().UnixNano()))
```
slintes, line 210: when you create you own rand, why don't you use it? ;)

slintes (changes requested):

test/e2e/far_e2e_test.go
razo7: The idea is that in the test setup there are three worker nodes, and the rand function uses a seed variable (in this case current time in nano seconds) to randomly pick a number. I use `previousNodeName` to verify on the second test that the selected worker node is different than the previous one. 
Since there are only 3 worker nodes I want to avoid choosing the previous node, then I have added this logic so we won't do the E2E test for only one node.

slintes: I understand what you want to do. The point is: `rand.New()` returns a new rand. But you don't use it. You use the default rand.

razo7: oops, now it used that 😆

slintes (approved):

PR #63, opened 2023-07-05
Add Explicit Workload Deletion

razo7:
Explicit/Manuel workload deletion.

Deleting the pods and their volume attachments in order to signal the scheduler that this node's workloads should be rescheduled on a different node.
Includes e2e and unit tests which adds volume attachments and pod to the unhealthy node in order to verify if they have deleted as expected.

[ECOPROJECT-1473](https://issues.redhat.com//browse/ECOPROJECT-1473)

### DIFF
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -57,6 +57,7 @@ func (r *FenceAgentsRemediationReconciler) SetupWithManager(mgr ctrl.Manager) er
 		Complete(r)
 }
 
+//+kubebuilder:rbac:groups=storage.k8s.io,resources=volumeattachments,verbs=get;list;watch;delete
 //+kubebuilder:rbac:groups=core,resources=pods/exec,verbs=create
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;update;delete;deletecollection
 //+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch;update;delete
@@ -164,6 +165,14 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "Fence Agent response wasn't a success message", "CR's Name", req.Name)
 		return emptyResult, err
 	}
+
+	// Reboot was finished and now we remove workloads (pods and their VA)
+	r.Log.Info("Manual workload deletion", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
+	if err := utils.DeleteResources(ctx, r.Client, req.Name); err != nil {
+		r.Log.Error(err, "Manual workload deletion has failed", "CR's Name", req.Name)
+		return emptyResult, err
+	}
+
 	return emptyResult, nil
 }
 
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -133,7 +133,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 	r.Log.Info("Fetch FAR's pod")
 	pod, err := utils.GetFenceAgentsRemediationPod(r.Client)
 	if err != nil {
-		r.Log.Error(err, "Can't find FAR's pod by it's label", "CR's Name", req.Name)
+		r.Log.Error(err, "Can't find FAR's pod by its label", "CR's Name", req.Name)
 		return emptyResult, err
 	}
 	//TODO: Check that FA is excutable? run cli.IsExecuteable
@@ -197,7 +197,7 @@ func buildFenceAgentParams(far *v1alpha1.FenceAgentsRemediation) ([]string, erro
 			return nil, err
 		}
 	}
-	// if --action attribute was not selected, then it's default value is reboot
+	// if --action attribute was not selected, then its default value is reboot
 	// https://github.com/ClusterLabs/fence-agents/blob/main/lib/fencing.py.py#L103
 	// Therefore we can safely add the reboot action regardless if it was initially added into the CR
 	fenceAgentParams = appendParamToSlice(fenceAgentParams, parameterActionName, parameterActionValue)
--- a/controllers/fenceagentsremediation_controller_test.go
+++ b/controllers/fenceagentsremediation_controller_test.go
@@ -27,6 +27,8 @@ import (
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -38,13 +40,17 @@ import (
 
 const (
 	dummyNode      = "dummy-node"
-	node01         = "worker-0"
+	workerNode     = "worker-0"
 	fenceAgentIPMI = "fence_ipmilan"
+	farPodName     = "far-pod"
+	testPodName    = "far-pod-test-1"
+	vaName1        = "va-test-1"
+	vaName2        = "va-test-2"
 )
 
 var (
-	faPodLabels    = map[string]string{"app.kubernetes.io/name": "fence-agents-remediation-operator"}
-	fenceAgentsPod *corev1.Pod
+	faPodLabels = map[string]string{"app.kubernetes.io/name": "fence-agents-remediation-operator"}
+	log         = ctrl.Log.WithName("controllers-unit-test")
 )
 
 var _ = Describe("FAR Controller", func() {
@@ -77,13 +83,13 @@ var _ = Describe("FAR Controller", func() {
 	}
 
 	// default FenceAgentsRemediation CR
-	underTestFAR := getFenceAgentsRemediation(node01, fenceAgentIPMI, testShareParam, testNodeParam)
+	underTestFAR := getFenceAgentsRemediation(workerNode, fenceAgentIPMI, testShareParam, testNodeParam)
 
 	Context("Functionality", func() {
 		Context("buildFenceAgentParams", func() {
 			When("FAR include different action than reboot", func() {
 				It("should succeed with a warning", func() {
-					invalidValTestFAR := getFenceAgentsRemediation(node01, fenceAgentIPMI, invalidShareParam, testNodeParam)
+					invalidValTestFAR := getFenceAgentsRemediation(workerNode, fenceAgentIPMI, invalidShareParam, testNodeParam)
 					invalidShareString, err := buildFenceAgentParams(invalidValTestFAR)
 					Expect(err).NotTo(HaveOccurred())
 					validShareString, err := buildFenceAgentParams(underTestFAR)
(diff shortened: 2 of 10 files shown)


controllers/fenceagentsremediation_controller.go
```diff
@@ -158,6 +159,14 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "Fence Agent response wasn't a success message", "CR's Name", req.Name)
 		return emptyResult, err
 	}
+
+	// Reboot was finished and now we enforce Non-Graceful Node Shutdown
+	r.Log.Info("Manuel workload deletion", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
```
clobrano, line 164: typo s/Manuel/Manual

(also few lines below)

pkg/utils/resources.go
```diff
@@ -0,0 +1,75 @@
...
+	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	loggerResouces = ctrl.Log.WithName("resource")
```
clobrano, line 16: typo s/loggerResouces/loggerResources/

clobrano (commented):
only some typos from my point of view
/lgtm 
Giving others a chance to review as well, feel free to unhold
/hold

pkg/utils/resources.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+func DeleteResources(r client.Client, nodeName string) error {
+	//fence
```
slintes, line 20: The reboot is the fencing.
This is just improving rescheduling time.

pkg/utils/resources.go
slintes: I would call it just `logger` or even `log`. Why do we need the package name in the name?

slintes (changes requested):

controllers/fenceagentsremediation_controller.go
```diff
@@ -158,6 +159,14 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "Fence Agent response wasn't a success message", "CR's Name", req.Name)
 		return emptyResult, err
 	}
+
+	// Reboot was finished and now we enforce Non-Graceful Node Shutdown
```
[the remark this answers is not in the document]
mshitrit: IIUC that's in line 150 `outputRes, outputErr, err := r.Executor.Execute(pod, cmd)`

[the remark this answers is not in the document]
slintes: - a *non graceful* shutdown might have happened already, depending on node issue which made it unhealthy
- we *deal with* the *non graceful* shutdown by resource deletion (or out of service taint)
- the reboot command is more likely a graceful shutdown (https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown)
- last but not least: the reboot command was executed before this comment already, and we don't do any other shutdown after this comment 🤷🏼‍♂️

pkg/utils/resources.go
razo7: - This is just improving rescheduling time.
What about stateful pods (pods with PVC)? IIUC these kind of pods won't move/evict to another node

pkg/utils/resources.go
```diff
@@ -0,0 +1,75 @@
...
+			if err != nil {
+				loggerResouces.Error(err, "failed to delete volumeAttachment", "name", va.Name)
+				return err
+			}
+		}
+	}
+
+	loggerResouces.Info("done deleting node resources", "node name", nodeName)
+
+	return nil
```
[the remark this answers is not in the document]
razo7: SNR rights are still preserved somehow with [line](https://github.com/medik8s/fence-agents-remediation/pull/63/files#diff-58c65d0ac78159b5a891995b8185dda69a88d7259470696f9de752dc52f7257fR3).
Next release we can move this logic, and more (such as taint logic from last PRs) to the common repo.

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
razo7: I believe this comment can be improved and reflect on the pods removal rather than Node Shutdown

pkg/utils/resources.go
razo7: > Why do we need the package name in the name?

Just for easier identification of the log's source.

slintes: I mean the var name :)

slintes (commented):
looks good now, what about tests? :)

pkg/utils/resources.go
```diff
@@ -0,0 +1,74 @@
...
+	"k8s.io/apimachinery/pkg/fields"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	log = ctrl.Log.WithName("utils-resource")
+)
+
+func DeleteResources(r client.Client, nodeName string) error {
```
slintes, line 19: we need to pass the context from the reconciler as parameter here, and use it in all client calls

slintes (changes requested):
something went wrong while rebasing, at least one commit is missing!

pkg/utils/resources.go
```diff
@@ -0,0 +1,75 @@
...
+	if err := r.List(context.Background(), &namespaces); err != nil {
+		loggerResouces.Error(err, "failed to list namespaces", err)
+		return err
+	}
+
+	loggerResouces.Info("starting to delete node resources", "node name", nodeName)
+
+	pod := &corev1.Pod{}
+	for _, ns := range namespaces.Items {
+		deleteOptions.Namespace = ns.Name
```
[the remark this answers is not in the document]
slintes: at least one commit is lost, it was addressing this comment ^^
https://github.com/medik8s/fence-agents-remediation/commit/b1574e8321863d77235a40ced5cdad766ed0807b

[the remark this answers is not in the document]
razo7: > I understand the ListOptions docs like an empty namespace means all namespaces.

You are right but my old comment that addressed your suggestion introduced an error on pod deletion -> `the server does not allow this method on the requested resource`.
When the namespace of ListOptions is [empty](https://github.com/medik8s/fence-agents-remediation/pull/63/files#diff-58c65d0ac78159b5a891995b8185dda69a88d7259470696f9de752dc52f7257fR26), and we delete as follows:

```suggestion
if err := r.DeleteAllOf(ctx, pod, deleteOptions); err != nil {
		log.Error(err, "failed to delete pods of unhealthy node") //, "namespace", ns.Name)
		return err
	}
```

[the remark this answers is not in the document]
slintes: oh, I see...

https://github.com/kubernetes-sigs/controller-runtime/issues/1842

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -102,69 +108,86 @@ var _ = Describe("FAR Controller", func() {
...
-				node = utils.GetNode("", node01)
+				node = utils.GetNode("", workerNode)
+				// createVA(vaName1, workerNode)
+				// createVA(vaName2, workerNode)
 				underTestFAR = getFenceAgentsRemediation(dummyNode, fenceAgentIPMI, testShareParam, testNodeParam)
 			})
-			It("should not have a finalizer nor taint", func() {
+			It("should not have a finalizer nor taint, while the two VAs and one pod will remain", func() {
 				By("Not finding a matching node to FAR CR's name")
 				nodeKey.Name = dummyNode
```
clobrano, line 172: The intention could be clearer if writing something like
```
nodeKey.Name = underTestFAR.Name
```
?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -182,18 +205,74 @@ func getFenceAgentsRemediation(nodeName, agent string, sharedparameters map[v1al
...
-	fenceAgentsPod.Namespace = defaultNamespace
+// buildPod builds a dummy pod
+func buildPod(containerName, podName, nodeName string) *corev1.Pod {
+	pod := &corev1.Pod{}
+	pod.Name = podName
+	if podName == farPodName {
+		// only when we build FAR pod then we add it's label
+		pod.Labels = faPodLabels
+	} else {
+		// testedPod should be reside on unhealthy node
```
clobrano, line 216: typo: *should reside or *should be

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -182,18 +205,74 @@ func getFenceAgentsRemediation(nodeName, agent string, sharedparameters map[v1al
...
+		Name:  containerName,
 		Image: "foo",
 	}
-	fenceAgentsPod.Spec.Containers = []corev1.Container{container}
-	return fenceAgentsPod
+	pod.Spec.Containers = []corev1.Container{container}
+	return pod
+}
+
+// createRunningPod builds new pod format, create it, and set it's status as running
```
clobrano, line 228: typo: *its

also below and in `far_e2e_test.go`

test/e2e/far_e2e_test.go
```diff
@@ -31,13 +31,19 @@ const (
 	fenceAgentAction         = "reboot"
 	nodeIdentifierPrefixAWS  = "--plug"
 	nodeIdentifierPrefixIPMI = "--ipport"
+	succeesRebootMessage     = "\"Success: Rebooted"
```
clobrano, line 34: typo: s/succeesRebootMessage/successRebootMessage

test/e2e/far_e2e_test.go
```diff
@@ -213,6 +234,27 @@ func randomizeWorkerNode(nodes *corev1.NodeList) string {
...
+			Namespace: testNsName,
+		},
+		Spec: storagev1.VolumeAttachmentSpec{
+			Attacher: pv,
+			Source:   storagev1.VolumeAttachmentSource{},
+			NodeName: nodeName,
+		},
+	}
+
+	va.Spec.Source.PersistentVolumeName = &pv
```
slintes, line 252: nit: why don't you set this above?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -182,18 +205,74 @@ func getFenceAgentsRemediation(nodeName, agent string, sharedparameters map[v1al
...
+	}
+	foo := "foo"
+	va.Spec.Source.PersistentVolumeName = &foo
+	ExpectWithOffset(1, k8sClient.Create(context.Background(), va)).To(Succeed())
+	return va
+}
+
+// verifyResourceCleanup fetches all the resources that we have crated for the test
+// and if they are still exist at the end of the test, then we clean them up for next test
+func verifyResourceCleanup(va1, va2 *storagev1.VolumeAttachment, pod *corev1.Pod) {
```
clobrano, line 258: I understand this function verifies if the resources have been already deleted, but its actual job is to clean them up, and the name should reflect that.
What something simple like `deleteResources`?

The same goes for the other similar functions here and in e2e test

slintes (approved):
one nit, I leave it up to you if you want to address it

/hold

clobrano (changes requested):

controllers/fenceagentsremediation_controller_test.go
razo7: I used it's when I wanted to use its... I will add that (and other typo comments) in a follow up PR if you don't mind

clobrano (commented):
/lgtm

clobrano (approved):

PR #64, opened 2023-07-05
End Reconcile after Finalizer Addition

razo7:
The reconcile should end when a finalizer has been added. Without ending the reconcile a new reconcile will always be triggered and will result in a duplicate reboot to the node (until FAR would have a status to the CR).

[ECOPROJECT-1415](https://issues.redhat.com//browse/ECOPROJECT-1415)

### DIFF
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -106,6 +106,10 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		if err := r.Client.Update(context.Background(), far); err != nil {
 			return emptyResult, fmt.Errorf("failed to add finalizer to the CR - %w", err)
 		}
+		r.Log.Info("Finalizer was added", "CR Name", req.Name)
+		return emptyResult, nil
+		// TODO: should return Requeue: true when the CR has a status and not end reconcile with empty result when a finalizer has been added
+		//return ctrl.Result{Requeue: true}, nil
 	} else if controllerutil.ContainsFinalizer(far, v1alpha1.FARFinalizer) && !far.ObjectMeta.DeletionTimestamp.IsZero() {
 		// Delete CR only when a finalizer and DeletionTimestamp are set
 		r.Log.Info("CR's deletion timestamp is not zero, and FAR finalizer exists", "CR Name", req.Name)


controllers/fenceagentsremediation_controller.go
```diff
@@ -106,6 +106,10 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		if err := r.Client.Update(context.Background(), far); err != nil {
 			return emptyResult, fmt.Errorf("failed to add finalizer to the CR - %w", err)
 		}
+		r.Log.Info("Finalizer was added", "CR Name", req.Name)
+		return emptyResult, nil
```
slintes, line 110: I do not understand how this prevents a duplicate reboot. Can you explain please?

razo7: I thought the PR description would explain it, but I will try again.

Before this change on a valid CR (by name), then the first reconcile will finish with emptyResult after running all the reconcile logic (including the FA execution). But since on the last (and first) reconcile call we have been updating the CR by adding a finalizer, then it triggers a second reconcile.
On this second reconcile it will try to execute the FA, since the CR doesn't have a status or any if block that we sparse the CR from running again the FA (it probably will fail until it eventually succeeds to reboot the node).
The reconcile won't finish be requeued until the FA will return success response, and we will end up with duplicate reboot.

After this change when we add the finalizer then we finish the reconcile with emptyResult. Since the CR was updated then on the second reconcile run will add taint and exeucte FA for the first time. Therefore, the PR change would stop the FA execution from the first reconcile run.

I am aware that other stuff, such as remediating the node which host the FAR PR, could trigger the reconcile again, and we will end up with another reboot but this problem will be resolved later with ECOPROJECT-1411

slintes: Ah ok, I mixed 1411 and 1415, sorry.

By returning the empty result here you only implicitly requeue because you know that it will happen because of the CR modification. I would prefer an explicit requeue. Did you try that? Or will it be reconciled twice then? 🤔

razo7: > By returning the empty result here you only implicitly requeue because you know that it will happen because of the CR modification

Correct.

> Did you try that? Or will it be reconciled twice then?

Correct again, see https://github.com/medik8s/fence-agents-remediation/pull/64/files#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R111-R112. I tried with `Requeue: true` or `RequeueAfter: 1 * time.Second`, and it resulted in a second reconcile. So until there's a status this PR will add this workaround...

slintes: ack, thx

slintes (approved):

PR #68, opened 2023-07-24
Set Node Unhealthy in E2E by Stopping Kubelet

razo7:
The E2E test should emphasize a realistic scenario of only when a node is unhealthy FAR CR should be created. We achieve that by stopping kubelet on the desired node.

### DIFF
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -79,13 +79,12 @@ var _ = Describe("FAR E2e", func() {
 
 	Context("stress cluster", func() {
 		var (
-			nodes, filteredNodes *corev1.NodeList
-			nodeName             string
+			nodes, filteredNodes                *corev1.NodeList
+			nodeName                            string
 			va                                  *storagev1.VolumeAttachment
 			pod                                 *corev1.Pod
 			creationTimePod, nodeBootTimeBefore time.Time
-			err                  error
-
+			err                                 error
 		)
 		BeforeEach(func() {
 			nodes = &corev1.NodeList{}
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -98,9 +98,10 @@ var _ = Describe("FAR E2e", func() {
 				Fail("No worker nodes found in the cluster")
 			}
 
-			testNodeName = randomizeWorkerNode(nodes)
-			previousNodeName = testNodeName
-			nodeNameParam := v1alpha1.NodeName(testNodeName)
+			selectedNode := randomizeWorkerNode(nodes)
+			nodeName = selectedNode.Name
+			previousNodeName = nodeName
+			nodeNameParam := v1alpha1.NodeName(nodeName)
 			parameterName := v1alpha1.ParameterName(nodeIdentifierPrefix)
 			testNodeID := testNodeParam[parameterName][nodeNameParam]
 			log.Info("Testing Node", "Node name", testNodeName, "Node ID", testNodeID)
@@ -120,7 +121,10 @@ var _ = Describe("FAR E2e", func() {
 			va = createVA(testNodeName)
 			DeferCleanup(cleanupTestedResources, va, pod)
 
-			far = createFAR(testNodeName, fenceAgent, testShareParam, testNodeParam)
+			// set the node as "unhealthy" by disabling kubelet
+			makeNodeUnready(selectedNode)
+
+			far := createFAR(nodeName, fenceAgent, testShareParam, testNodeParam)
 			DeferCleanup(deleteFAR, far)
 		})
 		When("running FAR to reboot two nodes", func() {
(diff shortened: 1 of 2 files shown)


slintes (approved):

test/e2e/far_e2e_test.go
```diff
@@ -201,17 +206,19 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
 	return testNodeParam, nil
 }
 
-// randomizeWorkerNode returns a worker node name which is different than the previous one
+// randomizeWorkerNode returns a worker node that his name is different than the previous one
```
mshitrit, line 209: IIUC this returns a random node out of the given node list.

>that his name is different than the previous one

I don't understand this part

razo7: Since we randomly pick a worker node and there are only 3 available nodes then we can pick the same node, even though we use a different seed for the random function. Therefore, we store the name of the last unhealthy node so we won't try to create second CR for the same node. 
> previous one

Refers to he previous remediation, so on the first remediation previousNodeName is nil, and on the second remediation it stores the name of the previous unheathy node in order to not create a CR on that node again.

mshitrit: I see now.
I advise against such implementation.
- IMO it's bad practice to use global var in such way (not concurrent safe and confusing)
- I think the method is confusing
- we have no idea how many iteration will occur before the for condition is broker because we can always get randomly the same node, also seems like a risk of a deadlock.

Instead I suggest the following , let `randomizeWorkerNode` do just that - get a random node out of the list:
```
func randomizeWorkerNode(nodes *corev1.NodeList) *corev1.Node {
	// Generate a random seed based on the current time
	r := rand.New(rand.NewSource(time.Now().UnixNano()))
	// Randomly select a worker node
	return &nodes.Items[r.Intn(len(nodes.Items))]
}
```

And just pass to it an already filtered list (filter out the node you don't want it to get).

test/e2e/utils/command.go
```diff
@@ -24,10 +24,21 @@ import (
 	"github.com/medik8s/fence-agents-remediation/api/v1alpha1"
 )
 
+// StopKubelet runs cmd connand to stop kubelet for the node and returns an error only if it fails
```
clobrano, line 27: typo s/connand/command

test/e2e/utils/command.go
```diff
@@ -24,10 +24,21 @@ import (
...
+	cmd := "microdnf install util-linux -y && /usr/bin/nsenter -m/proc/1/ns/mnt /bin/systemctl stop kubelet"
+	_, err := runCommandInCluster(c, nodeName, testNsName, cmd, log)
+	if err != nil && strings.Contains(err.Error(), "connection refused") {
+		log.Info("ignoring expected error when stopping kubelet", "error", err.Error())
+		return nil
+	}
+	return err
+}
+
 // GetBootTime gets the boot time of the given node by running a pod on it executing uptime command
```
clobrano, line 38: nit: function description should not state how the function works, just what it is expected to do (also, not part of this PR, so no need to change it here)

test/e2e/far_e2e_test.go
```diff
@@ -99,12 +100,29 @@ var _ = Describe("FAR E2e", func() {
...
+		When("running FAR to reboot a node", func() {
+			It("should successfully remediate it", func() {
+				checkRemediation(nodeName, succeesRebootMessage, nodeBootTimeBefore)
+			})
+		})
+
 		When("running FAR to reboot two nodes", func() {
 			It("should successfully remediate the first node", func() {
 				checkRemediation(nodeName, succeesRebootMessage, nodeBootTimeBefore)
+				// run the second remediation on a different node
```
slintes, line 119: This looks like setup code for the next test, it shouldn't be here.
I think all this node selection stuff can and should be in the `BeforeEach()` block above...

slintes (changes requested):

test/e2e/far_e2e_test.go
razo7: I think it suits better an AfterEach, so I have used it.

test/e2e/far_e2e_test.go
```diff
@@ -99,10 +100,29 @@ var _ = Describe("FAR E2e", func() {
...
 			DeferCleanup(deleteFAR, far)
 		})
+		When("running FAR to reboot one node", func() {
+			It("should successfully remediate it", func() {
+				checkRemediation(nodeName, nodeBootTimeBefore)
+			})
+		})
+
 		When("running FAR to reboot two nodes", func() {
+			AfterEach(func() {
```
slintes, line 117: I don't understand what you're trying to do here...

Can you please put this node selection stuff into the existing BeforeEach(), as suggested earlier already?

slintes (changes requested):

test/e2e/far_e2e_test.go
```diff
@@ -75,33 +73,53 @@ var _ = Describe("FAR E2e", func() {
...
 			nodeBootTimeBefore, err = e2eUtils.GetBootTime(clientSet, nodeName, testNsName, log)
 			Expect(err).ToNot(HaveOccurred(), "failed to get boot time of the node")
+
+			// set the node as "unhealthy" by disabling kubelet
+			makeNodeUnready(selectedNode)
+
 			far := createFAR(nodeName, fenceAgent, testShareParam, testNodeParam)
 			DeferCleanup(deleteFAR, far)
 		})
+		When("running FAR to reboot one node", func() {
```
slintes, line 117: Probably caused by rebasing etc., but I think this `When()` block can be removed. We test 2 remediations in the `When()` block below, that's enough I think :)

slintes (commented):
one last thing... other than that lgtm

slintes (approved):

PR #69, opened 2023-07-27
Add Status Conditions for FAR CR

razo7:
Adding three status conditions (`Processing`, `FenceAgentActionSucceeded`, and `Succeeded`) to help FAR in two ways:

- Convey the status of processing the CR, if the fence agent action was succeeded (only once), and if the whole remediation was succeeded (node was tainted with FAR taint, fence agent action was succeeded, and the workloads have been deleted).
- Limit the available sections in the reconcile based on the conditions, e.g., reboot and resource deletion will finish successfully only once.
 
Moreover, each status condition includes a reason and a message based on ProcessingChangeReason that changed the condition value.
The PR also updates reconcile structure : 

- Fetch FAR CR, validate CR name, check NHC timeout annotation, and add finalizer. ProcessingChangeReason = RemediationStarted
- Try to add FAR taint, build FA command and execute it (until it succeeds).  ProcessingChangeReason = FenceAgentSucceeded
- Try to delete workloads. ProcessingChangeReason = RemediationFinished

[ECOPROJECT-1411](https://issues.redhat.com//browse/ECOPROJECT-1411) 
[ECOPROJECT-1484](https://issues.redhat.com//browse/ECOPROJECT-1484)


slintes (changes requested):
Please use conditions

> Some resources in the v1 API contain fields called phase, and associated message, reason, and other status fields. The pattern of using phase is deprecated. Newer API types should use conditions instead. Phase was essentially a state-machine enumeration field, that contradicted [system-design principles](https://git.k8s.io/design-proposals-archive/architecture/principles.md#control-logic) and hampered evolution, since [adding new enum values breaks backward compatibility](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md). Rather than encouraging clients to infer implicit properties from phases, we prefer to explicitly expose the individual conditions that clients need to monitor. Conditions also have the benefit that it is possible to create some conditions with uniform meaning across all resource types, while still exposing others that are unique to specific resource types. See [#7856](http://issues.k8s.io/7856) for more details and discussion.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

slintes (changes requested):

controllers/fenceagentsremediation_controller.go
```diff
@@ -145,37 +172,133 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
-	r.Log.Info("Manual workload deletion", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
-	if err := utils.DeleteResources(ctx, r.Client, req.Name); err != nil {
-		r.Log.Error(err, "Manual workload deletion has failed", "CR's Name", req.Name)
-		return emptyResult, err
+		if err := updateConditions(v1alpha1.RemediationFinished, &far.Status.Conditions, r.Log); err != nil {
+			return emptyResult, err
+		}
+		r.Log.Info("FenceAgentsRemediation CR has completed to remediate the node", "Node Name", req.Name)
+
+		return ctrl.Result{Requeue: true}, nil
```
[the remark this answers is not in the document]
razo7: Not really. Can be discarded

controllers/fenceagentsremediation_controller.go
```diff
@@ -145,37 +172,133 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
-		r.Log.Error(err, "Fence Agent response was a failure", "CR's Name", req.Name)
-		return emptyResult, err
-	}
-	if outputErr != "" || outputRes != SuccessFAResponse+"\n" {
-		// response wasn't failure or sucesss message
-		err := fmt.Errorf("unknown fence agent response - expecting `%s` response, but we received `%s`", SuccessFAResponse, outputRes)
-		r.Log.Error(err, "Fence Agent response wasn't a success message", "CR's Name", req.Name)
-		return emptyResult, err
+
+	if meta.IsStatusConditionPresentAndEqual(far.Status.Conditions, commonConditions.ProcessingType, metav1.ConditionTrue) &&
```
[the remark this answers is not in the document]
razo7: >a better check for the FASucceeded condition would be != True. 

👍🏻 
>I think the code which prepares command execution can be moved from above inside this if block? (getting the pod, building params..

I though of that, and I went with the current implementation in order to limit the code sections which are affected by the status conditions. Having said that do you still see a greater value of adding the suggested code under the if block?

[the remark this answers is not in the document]
razo7: >Then you can set the condition to False in case execution failed.

ATM there is no ProcessingChangeReason for this use case. But I might add something for that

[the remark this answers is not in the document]
slintes: > Having said that do you still see a greater value of adding the suggested code under the if block?

yes, why executing all that code when it's not used 🤷🏼‍♂️ 

> ATM there is no ProcessingChangeReason

tbh, I dislike this "one reason for updating all conditions" pattern anyway, and this is why...

api/v1alpha1/fenceagentsremediation_types.go
```diff
@@ -74,12 +74,10 @@ type FenceAgentsRemediationStatus struct {
 
 	// Represents the observations of a FenceAgentsRemediation's current state.
 	// Known .status.conditions.type are: "Processing", "FenceAgentActionSucceeded", and "Succeeded".
-	// +patchMergeKey=type
-	// +patchStrategy=merge
 	// +listType=map
 	// +listMapKey=type
 	// +operator-sdk:csv:customresourcedefinitions:type=status,displayName="conditions",xDescriptors="urn:alm:descriptor:io.kubernetes.conditions"
-	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
+	Conditions []metav1.Condition `json:"conditions,omitempty" protobuf:"bytes,1,rep,name=conditions"`
```
slintes, line 80: missed in the first review: why the `protobuf` tag?

slintes: That doc is primarily targeting core k8s types. So in general it's good to follow its recommendations, but not everything applies to CRDs as well 🙂

controllers/fenceagentsremediation_controller.go
```diff
@@ -156,34 +156,34 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
+		//TODO: Check that FA is excutable? run cli.IsExecuteable
+
+		// Build FA parameters
+		r.Log.Info("Combine fence agent parameters", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
+		faParams, err := buildFenceAgentParams(far)
+		if err != nil {
+			r.Log.Error(err, "Invalid sharedParameters/nodeParameters from CR - edit/recreate the CR", "CR's Name", req.Name)
+			return emptyResult, nil
+		}
+		// Add FAR (medik8s) remediation taint
```
slintes, line 180: should the taint be in this block as well? 🤔

controllers/fenceagentsremediation_controller.go
```diff
@@ -128,52 +155,159 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
+	return nil
+}
+
+// updateConditions updates the status conditions of a FenceAgentsRemediation object based on the provided ProcessingChangeReason.
+// return an error if an unknown ProcessingChangeReason is provided
+func updateConditions(reason v1alpha1.ProcessingChangeReason, currentConditions *[]metav1.Condition, log logr.Logger) error {
+
+	var (
+		processingConditionStatus, fenceAgentActionSucceededConditionStatus, succeededConditionStatus metav1.ConditionStatus
+		conditinMessage                                                                               string
```
slintes, line 241: typo: conditionMessage

api/v1alpha1/fenceagentsremediation_types.go
```diff
@@ -23,16 +23,39 @@ import (
...
-type NodeName string
-
 const (
 	// FARFinalizer is a finalizer for a FenceAgentsRemediation CR deletion
 	FARFinalizer string = "fence-agents-remediation.medik8s.io/far-finalizer"
 	// Taints
 	FARNoExecuteTaintKey = "medik8s.io/fence-agents-remediation"
+	// FenceAgentActionSucceededType is the condition type used to signal whether the Fence Agent action was succeeded successfully or not
+	FenceAgentActionSucceededType = "FenceAgentActionSucceeded"
+	// error status messages
```
slintes, line 33: these errors are used for tests only, correct? They shouldn't be in the api package then.

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -333,19 +351,37 @@ func testPodDeletion(podName string, resourceDeletionWasTriggered bool) {
...
 			return apierrors.IsNotFound(err)
 
-		}, 5*time.Second, 250*time.Millisecond).Should(BeFalse())
+		}, timeoutDeletion, pollInterval).Should(BeFalse())
 		log.Info("Pod exist", "pod", podName)
 	}
 }
 
+// verifyStatusCondition checks whether the status condition is set with the expected value
+func verifyStatusCondition(testFAR *v1alpha1.FenceAgentsRemediation, conditionType, expectedResult string, conditionStatus metav1.ConditionStatus) {
```
slintes, line 368: I won't block on this, but using this "expectedResult" string is strange at least. Type and status params should be enough for the tests, not? Either they match or not.

razo7: Yes, but I added the `expectedResult` so the Eventually can "catch" an expected error, e.g., in [unit-test](https://github.com/medik8s/fence-agents-remediation/pull/69/files#diff-fac0cf56dda42683c1e24b46c87cfa4cca46c444145ee69de9e264db50f10289R201-R204) when the conditions haven't been set.
I have changed the name to `verifyExpectedStatusConditionError` as it better captures the essence of this function.

api/v1alpha1/fenceagentsremediation_types.go
razo7: I have moved them to `errors.go` file

controllers/fenceagentsremediation_controller_test.go
slintes: you could make `conditionStatus` a pointer and use `nil`, if you want to test that a condition isn't set 🙂

razo7: But then the test would fail, since eventually will expect `utils.ConditionSetAndMatchSuccess` while verifyStatusCondition/verifyExpectedStatusConditionError would return `utils.ConditionSetButNoMatchError`.

slintes (approved):

controllers/fenceagentsremediation_controller_test.go
slintes: this works: https://github.com/slintes/fence-agents-remediation/commit/55781eac1d0652cfa99f739aeafef21b62d32d18?diff=unified

slintes:
oh, CI missed to add the label

/lgtm

PR #70, opened 2023-07-30
Improve Verification of FA Success Response

razo7:
`checkFarLogs` was used to verify whether FAR pod had a success response from FA command.

- When we are running more than one remediation then `checkFarLogs` function could have find the success message of the first CR since we have searched for the response without any CR identification (see [old discussion](https://github.com/medik8s/fence-agents-remediation/pull/55#discussion_r1223232556)). Now we are using the node name to verify that (since we won't remediate a node more than once).

- When the FAR pod resides in the unhealthy node, then it will be evicted and one FA command is run only once (after #69), then when it will be run again on a new node it won't try to execute the FA command. Therefore, for this scenario we are skipping the check of checkFarLogs function.

### DIFF
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -353,15 +353,16 @@ func buildExpectedLogOutput(nodeName, successMessage string) string {
 	return expectedString
 }
 
-// checkFarLogs gets the FAR pod and checks whether it's logs have logString
-func checkFarLogs(farNodeName, logString string) {
+// checkFarLogs gets the FAR pod and checks whether it's logs have logString, and if the pod was in the unhealthyNode
+// then we don't look for the expected logString
+func checkFarLogs(unhealthyNodeName, logString string) {
 	EventuallyWithOffset(1, func() string {
 		pod, err := utils.GetFenceAgentsRemediationPod(k8sClient)
 		if err != nil {
 			log.Error(err, "failed to get FAR pod. Might try again")
 			return ""
 		}
-		if pod.Spec.NodeName == farNodeName {
+		if pod.Spec.NodeName == unhealthyNodeName {
 			// When reboot is running on FAR node, then FAR pod will be recreated on a new node
 			// and since the FA command won't be executed again, then the log won't include
 			// any success message, so we won't verfiy the FAR success message on this scenario
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -346,6 +346,13 @@ func makeNodeUnready(node *corev1.Node) {
 	log.Info("node is unready", "node name", node.GetName())
 }
 
+// buildExpectedLogOutput returns a string with a node identifier and a success message for the reboot action
+func buildExpectedLogOutput(nodeName, successMessage string) string {
+	expectedString := fmt.Sprintf("\"Node name\": \"%s\", \"Response\": \"%s", nodeName, successMessage)
+	log.Info("Substring to search in the logs", "expectedString", expectedString)
+	return expectedString
+}
+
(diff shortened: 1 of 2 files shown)


test/e2e/far_e2e_test.go
```diff
@@ -309,14 +309,28 @@ func wasFarTaintAdded(nodeName string) {
...
+func checkFarLogs(farNodeName, logString string) {
 	EventuallyWithOffset(1, func() string {
 		pod, err := utils.GetFenceAgentsRemediationPod(k8sClient)
 		if err != nil {
 			log.Error(err, "failed to get FAR pod. Might try again")
 			return ""
 		}
+		if pod.Spec.NodeName == farNodeName {
+			// When reboot is running on FAR node, then FAR pod will be recreated on a new node
+			// and since the FA command won't be exuected again, then the log won't include
```
clobrano, line 329: typo: s/exuected/executed

test/e2e/far_e2e_test.go
```diff
@@ -309,14 +309,28 @@ func wasFarTaintAdded(nodeName string) {
...
+			// any success message, so we won't verfiy the FAR success messsage on this scenario
```
clobrano, line 330: typos:
- verfiy
- messsage

test/e2e/far_e2e_test.go
```diff
@@ -309,14 +309,28 @@ func wasFarTaintAdded(nodeName string) {
 	log.Info("FAR taint was added", "node name", node.Name, "taint key", farTaint.Key, "taint effect", farTaint.Effect)
 }
 
-// checkFarLogs gets the FAR pod and checks whether its logs have logString
-func checkFarLogs(logString string) {
+// buildExpectedLogOutput returns a string with a node identifier and a success message for the reboot action
+func buildExpectedLogOutput(nodeName, sucesssMessage string) string {
```
clobrano, line 313: typo `sucesssMessage`

test/e2e/far_e2e_test.go
```diff
@@ -309,14 +309,28 @@ func wasFarTaintAdded(nodeName string) {
...
+	expectedString := "\"Node name\": \"" + nodeName + "\", \"Response\": \"" + sucesssMessage
```
clobrano, line 314: is there a reason not to use `fmt.Sprintf`? I should be easier to read

clobrano (changes requested):
Some typos and an observation

test/e2e/far_e2e_test.go
razo7: Good point

test/e2e/far_e2e_test.go
```diff
@@ -346,14 +346,28 @@ func makeNodeUnready(node *corev1.Node) {
...
+
+// checkFarLogs gets the FAR pod and checks whether it's logs have logString
+func checkFarLogs(farNodeName, logString string) {
 	EventuallyWithOffset(1, func() string {
 		pod, err := utils.GetFenceAgentsRemediationPod(k8sClient)
 		if err != nil {
 			log.Error(err, "failed to get FAR pod. Might try again")
 			return ""
 		}
+		if pod.Spec.NodeName == farNodeName {
```
mshitrit, line 364: Nit: IIUC `farNodeName` is the node that being remidiated.
I think that the name is a bit confusing in that context (initially I thought `farNodeName` is the node that Far Pod is running on).

razo7: Thanks, it was indeed confusing.

clobrano (approved):
/lgtm 
Giving others a chance to review old comments, feel free to unhold
/hold

razo7:
I think it is safe to merge
/unhold

PR #72, opened 2023-07-31
Stop Remediation When NHC Timed Out Annotation Exists

razo7:
Stop remediation when NHC timed out annotation exists

[ECOPROJECT-1485](https://issues.redhat.com//browse/ECOPROJECT-1485)

### DIFF
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	commonAnnotations "github.com/medik8s/common/pkg/annotations"
 
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


controllers/fenceagentsremediation_controller.go
```diff
@@ -129,6 +132,13 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		return emptyResult, nil
 	}
 
+	// Check NHC timeout annotation
+	if isTimedOutByNHC(far) {
```
slintes, line 136: I think this can go before the if block above?

controllers/fenceagentsremediation_controller.go
```diff
@@ -129,6 +132,13 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
+		r.Log.Info("FAR remediation was stopped by Node Healthcheck Operator")
+		// TODO: update status and return its error
+		return emptyResult, errors.New(errorNhcTimedOut)
```
slintes, line 139: why returning an error?

controllers/fenceagentsremediation_controller.go
razo7: Yes, it can

controllers/fenceagentsremediation_controller.go
slintes: > the CR will be stopped here until the NHC annotation would be removed from the CR

exactly, and removing the annotation will trigger a reconcile automatically, no need to requeue ourself 🙂

slintes (approved):
lgtm

putting a hold on it to let you decide if you want this one to be merged first, or the condition PR 🙂 

/hold

razo7:
I prefer that whatever can be merged will be merged. There is no precedence between them.

/unhold

PR #73, opened 2023-08-02
Improve verifyStatusCondition Function Readability & Gomega Expect Update

razo7:
- Discard an error.go file and improve the usage of verifyStatusCondition for non existing status condition or expected value is different than actual value
- Pass an explicit instance of Gomega Expect to the Eventually/Consistently functions- for more see in [Gomega docs](https://onsi.github.io/gomega/#consistently).


slintes (approved):

razo7:
Due to channel change from the last merged PR I have rebased the PR

PR #74, opened 2023-08-14
Stale Cache on Fast Reconcile

razo7:
Sometime status is not updated on a fast reconcile, after a reconcile ends a new reconcile occur immediately (`ctrl.Result{Requeue: true}`), due to stale cache. To overcome it we add the `LastUpdateTime` field to status so we can use  it for comparing far object CR from the beginning of the reconcile and the latest (which is supposed to have the updated status).

Similar PR from NHC - https://github.com/medik8s/node-healthcheck-operator/pull/245
[ECOPROJECT-1523](https://issues.redhat.com//browse/ECOPROJECT-1523) [ECOPROJECT-1411](https://issues.redhat.com//browse/ECOPROJECT-1411)


controllers/fenceagentsremediation_controller.go
```diff
@@ -246,7 +247,7 @@ func isTimedOutByNHC(far *v1alpha1.FenceAgentsRemediation) bool {
 func (r *FenceAgentsRemediationReconciler) updateStatus(ctx context.Context, far *v1alpha1.FenceAgentsRemediation) error {
 	if err := r.Client.Status().Update(ctx, far); err != nil {
 		if !apiErrors.IsConflict(err) {
```
slintes, line 249: Please also log the conflict, it should happen very seldom only when NHC updated the CR for setting the timeout annotation. We are hiding the issue you are resolving with this PR without logging.

slintes (changes requested):

slintes:
/hold

let's consider a better way for dealing with stale cache on quick requeue

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -184,7 +185,7 @@ var _ = Describe("FAR Controller", func() {
 
 				By("Not having finalizer")
 				farNamespacedName.Name = underTestFAR.Name
-				Eventually(func(g Gomega) bool {
+				Consistently(func(g Gomega) bool {
```
slintes, line 188: I'm wondering how useful Consitently is with a 100ms duration only 🤔

razo7: Before the change eventually ran only once, 10 ms, and now it waits longer which helps underTestFAR in [line 204](https://github.com/medik8s/fence-agents-remediation/pull/74/files#diff-fac0cf56dda42683c1e24b46c87cfa4cca46c444145ee69de9e264db50f10289R204) to be updated with the last status update, so we can check that `LastUpdateTime` is not nil

controllers/fenceagentsremediation_controller.go
```diff
@@ -247,27 +250,45 @@ func isTimedOutByNHC(far *v1alpha1.FenceAgentsRemediation) bool {
...
+	// Wait until the cache is updated in order to prevent reading a stale status in the next reconcile
+	// and making wrong decisions based on it.
+	pollErr := wait.PollUntilContextTimeout(ctx, 200*time.Millisecond, 5*time.Second, true, func(ctx context.Context) (bool, error) {
+		tmpFar := &v1alpha1.FenceAgentsRemediation{}
+		if err := r.Client.Get(ctx, client.ObjectKeyFromObject(far), tmpFar); err != nil {
+			if apiErrors.IsNotFound(err) {
+				// nothing we can do anymore
+				return true, nil
+			}
+			return false, err
```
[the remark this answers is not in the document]
razo7: Correct, a leftover of my misunderstanding of the `PollUntilContextTimeout` usage

controllers/fenceagentsremediation_controller_test.go
slintes: sorry, but that's the wrong approach:

- with eventually you wait until something happens
- with consistently you ensure that something does NOT change

So here you want to ensure that no finalizer is added. For that using Consistently is correct, but you did it for the wrong reason 😉 That low duration though does add that much value to the test IMHO. But that's a nit.

When you have a issue in line 204, fix it there, not here. Use Eventually over there if you need to wait.

slintes (approved):

PR #75, opened 2023-08-14
Fix VA Deletion Check in E2E Testing

razo7:
Check with Consistently instead of Eventually and use offset

### DIFF
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -124,7 +124,7 @@ var _ = Describe("FAR E2e", func() {
 			pod = e2eUtils.GetPod(nodeName, testContainerName)
 			pod.Name = testPodName
 			pod.Namespace = testNsName
-			Expect(k8sClient.Create(context.Background(), pod)).To(Succeed())
+			ExpectWithOffset(1, k8sClient.Create(context.Background(), pod)).To(Succeed())
 			log.Info("Tested pod has been created", "pod", testPodName)
 			creationTimePod = metav1.Now().Time
 			va = createVA(nodeName)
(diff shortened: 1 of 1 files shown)


test/e2e/far_e2e_test.go
```diff
@@ -405,7 +405,7 @@ func wasNodeRebooted(nodeName string, nodeBootTimeBefore time.Time) {
 
 // checkVaDeleted verifies if the va has already been deleted due to resource deletion
 func checkVaDeleted(va *storagev1.VolumeAttachment) {
-	EventuallyWithOffset(1, func() bool {
+	ConsistentlyWithOffset(1, func() bool {
```
slintes, line 408: Why this change? For waiting until something happens, Eventually is the right tool 🤔

test/e2e/far_e2e_test.go
```diff
@@ -455,12 +455,12 @@ func checkRemediation(nodeName string, nodeBootTimeBefore time.Time, oldPodCreat
 	By("Getting new node's boot time")
 	wasNodeRebooted(nodeName, nodeBootTimeBefore)
 
-	By("checking if old VA has been deleted")
-	checkVaDeleted(va)
-
 	By("checking if old pod has been deleted")
 	checkPodDeleted(pod)
 
+	By("checking if old volume attachment has been deleted")
```
slintes, line 461: Why the change in order? AFAIK in a real world the pod won't be deleted before its related VA is deleted (that's why we delete the VA actually). So first checking the VA makes more sense to me.

razo7:
Closing the PR as it is not needed anymore:
1. Marc's remarks are correct
2. VA deletion by FAR could interfere with the process of detaching the VAs by kube-control-manager.

PR #76, opened 2023-08-14
Add Two Reasons for Updating and Stopping Remediation

razo7:
- Adding two new reasons in order to update the CR three conditions to `false`:
		- `RemediationInterruptedByNHC` - FAR CR name doesn't match a node name
		- `RemediationFinishedNodeNotFound` -   Node Healthcheck timeout annotation has been set
- Don't stop an update of condition from false to true.

### DIFF
--- a/api/v1alpha1/fenceagentsremediation_types.go
+++ b/api/v1alpha1/fenceagentsremediation_types.go
@@ -43,8 +43,8 @@ type ConditionsChangeReason string
 const (
 	// RemediationFinishedNodeNotFound - CR was found but its name doesn't matche a node
 	RemediationFinishedNodeNotFound ConditionsChangeReason = "RemediationFinishedNodeNotFound"
-	// RemediationStoppedByNHC - Remediation was interrupted by NHC timeout annotation
-	RemediationStoppedByNHC ConditionsChangeReason = "RemediationStoppedByNHC"
+	// RemediationInterruptedByNHC - Remediation was interrupted by NHC timeout annotation
+	RemediationInterruptedByNHC ConditionsChangeReason = "RemediationInterruptedByNHC"
 	// RemediationStarted - CR was found, its name matches a node, and a finalizer was set
 	RemediationStarted ConditionsChangeReason = "RemediationStarted"
 	// FenceAgentSucceeded - FAR taint was added, fence agent command has been created and executed successfully
--- a/api/v1alpha1/fenceagentsremediation_types.go
+++ b/api/v1alpha1/fenceagentsremediation_types.go
@@ -31,7 +31,8 @@ const (
 	// FenceAgentActionSucceededType is the condition type used to signal whether the Fence Agent action was succeeded successfully or not
 	FenceAgentActionSucceededType = "FenceAgentActionSucceeded"
 	// condition messages
-	RemediationInterruptedConditionMessage          = "FAR CR name doesn't match a node name or Node Healthcheck timeout annotation has been set"
+	RemediationFinishedNodeNotFoundConditionMessage = "Node Healthcheck timeout annotation has been set"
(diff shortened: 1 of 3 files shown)


api/v1alpha1/fenceagentsremediation_types.go
```diff
@@ -31,21 +31,26 @@ const (
 	// FenceAgentActionSucceededType is the condition type used to signal whether the Fence Agent action was succeeded successfully or not
 	FenceAgentActionSucceededType = "FenceAgentActionSucceeded"
 	// condition messages
-	RemediationStartedConditionMessage  = "FAR CR was found, the CR name matches one of the cluster nodes, and a finalizer was set"
-	FenceAgentSucceededConditionMessage = "FAR taint was added, fence agent command has been created and executed successfully"
-	RemediationFinishedConditionMessage = "The unhealthy node was fully remediated (it was tainted, fenced using FA and all the node resources have been deleted)"
+	RemediationInterruptedConditionMessage          = "FAR CR name doesn't match a node name or Node Healthcheck timeout annotation has been set"
```
slintes, line 34: one reason for actual 2 reasons isn't that helpful for users, is it?

razo7: Correct, I neglected the value of separating the messages.

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -200,9 +200,9 @@ var _ = Describe("FAR Controller", func() {
 				testPodDeletion(testPodName, resourceDeletionWasTriggered)
 
 				By("Not having any condition set")
```
slintes, line 202: message does not fit anymore

controllers/fenceagentsremediation_controller.go
```diff
@@ -253,20 +254,31 @@ func (r *FenceAgentsRemediationReconciler) updateStatus(ctx context.Context, far
...
 	// FenceAgentSucceeded can only happen after RemediationStarted happened
-	// RemediationFinished can only happen after FenceAgentSucceeded happened
+	// RemediationFinishedSuccessfully can only happen after FenceAgentSucceeded happened
 
 	switch reason {
+	case v1alpha1.RemediationFinishedNodeNotFound, v1alpha1.RemediationInterruptedByNHC:
+		processingConditionStatus = metav1.ConditionFalse
+		fenceAgentActionSucceededConditionStatus = metav1.ConditionFalse
+		succeededConditionStatus = metav1.ConditionFalse
+		if reason == v1alpha1.RemediationFinishedNodeNotFound {
```
slintes, line 276: IMHO a switch makes more sense here, which errors when a reason isn't handled. Prevents forgetting this part when adding another reason to the case statement above.

slintes (commented):
2 nits...

controllers/fenceagentsremediation_controller.go
razo7: Good point

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -166,7 +166,7 @@ var _ = Describe("FAR Controller", func() {
 				testVADeletion(vaName2, resourceDeletionWasTriggered)
 				testPodDeletion(testPodName, resourceDeletionWasTriggered)
 
-				By("Having Succeed condition set to true")
+				By("Having Succeeded, FenceAgentActionSucceeded conditions set to true, and Processing set to false")
```
slintes, line 169: just a general note: I would avoid being too detailed in the these comments, and use something more generic like "Verifying correct conditions". The details are very readable in the actual test code, and in the test logs in case a test fails. That way you don't need to update the comment ever time the test changes 🙂

slintes (approved):

PR #78, opened 2023-08-23
Match Condition Message to Variable Name

razo7:
Condition message didn't match the variable name.
Correct condition message between Node not found and NHC time out scenarios


slintes (approved):
and this is why I try to avoid using production consts in tests 😉

PR #84, opened 2023-09-28
Stop Update Status When a Finalizer is Missing

razo7:
When FAR doesn't include a finalizer then there is no need to update the status, since it will be removed soon.

Should fix ECOPROJECT-1590


controllers/fenceagentsremediation_controller.go
```diff
@@ -247,6 +247,11 @@ func isTimedOutByNHC(far *v1alpha1.FenceAgentsRemediation) bool {
 
 // updateStatus updates the CR status, and returns an error if it fails
 func (r *FenceAgentsRemediationReconciler) updateStatus(ctx context.Context, far *v1alpha1.FenceAgentsRemediation) error {
+	// When FAR doesn't include a finalizer after successful remediation then there is no need to update the status, since it will be removed soon.
```
clobrano, line 250: What if there is no finalizer, but the conditions are not "True"? Could we still get this error?

razo7: The bug that we faced was a race condition between two operators - when NHC removed far CR, and in the meantime, the reconcile loop was finished and tried to update the CR status. The suggested fix will prevent _updateStatus_ from updating the status of CR at the end of the reconcile loop when some conditions values have been met (remediation is completed) and a finalizer was removed (it has to be created beforehand based on the values of the conditions).

>What if there is no finalizer, 

There is no finalizer in two scenarios:

1. Finalizer couldn't be created once we have reached this function - due to NHC timeout or far CR name being invalid.
2. Or it was created and deleted already - therefore, the node was fully remediated.

>Could we still get this error?

I don't think so, but can you see a way/chance when we should still see this kind of error?

clobrano: ok then

/lgtm 
Giving others a chance to review as well, feel free to unhold
/hold

slintes (changes requested):

controllers/fenceagentsremediation_controller.go
razo7: Sounds even better using `deletionTimestamp`.

PR #85, opened 2023-09-28
Fix FenceAgentSucceededConditionMessage Condition Message

razo7:
Missing `and the` in the condition message.


api/v1alpha1/fenceagentsremediation_types.go
```diff
@@ -34,7 +34,7 @@ const (
 	RemediationFinishedNodeNotFoundConditionMessage = "FAR CR name doesn't match a node name"
 	RemediationInterruptedByNHCConditionMessage     = "Node Healthcheck timeout annotation has been set"
 	RemediationStartedConditionMessage              = "FAR CR was found, its name matches one of the cluster nodes, and a finalizer was set to the CR"
-	FenceAgentSucceededConditionMessage             = "FAR taint was added, fence agent command has been created and executed successfully"
+	FenceAgentSucceededConditionMessage             = "FAR taint was added, and fence agent command has been created and executed successfully"
```
clobrano, line 37: Nit: I think you don't need the comma here, it is not a series of things, where usually the Oxford comma is used.

razo7: looks better?

PR #87, opened 2023-10-02
[WIP] Use Jobs to execute fencing agents

clobrano:
Use Jobs to execute fencing agents

- [x] split current image to have one for the manager and one for the fencing-agents
- [ ] create a Job to execute fencing-agents on a node
- [ ] handle Job's life cycle
- [ ] stop Job in case of time out

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,6 +86,9 @@ CATALOG_IMG ?= $(IMAGE_TAG_BASE)-operator-catalog:$(IMAGE_TAG)
 # Image URL to use all building/pushing image targets
 IMG ?= $(IMAGE_TAG_BASE)-operator:$(IMAGE_TAG)
 
+# Image URL to use for building/pushing fencing agents image
+JOB_IMG ?= $(IMAGE_TAG_BASE)-fencing-agents:$(IMAGE_TAG)
+
 # BUNDLE_GEN_FLAGS are the flags passed to the operator-sdk generate bundle command
 BUNDLE_GEN_FLAGS ?= -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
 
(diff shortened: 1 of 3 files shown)


clobrano:
Really a WIP :smiley:

Makefile
```diff
@@ -86,6 +86,9 @@ CATALOG_IMG ?= $(IMAGE_TAG_BASE)-operator-catalog:$(IMAGE_TAG)
 # Image URL to use all building/pushing image targets
 IMG ?= $(IMAGE_TAG_BASE)-operator:$(IMAGE_TAG)
 
+# Image URL to use for building/pushing fencing agents image
+JOB_IMG ?= $(IMAGE_TAG_BASE)-fencing-agents:$(IMAGE_TAG)
```
razo7, line 90: NIT: I would prefer to use `fence-agent` or `fencing-agent` instead of `fencing-agents` since this image will be used for running one fence agent by a Job.
This comment is related to all the other appearances of `fencing-agents`

clobrano: I understand, but this image actually contains all the fencing agents, so I think calling it `fance-agent` might be misleading

clobrano:
> Would it make more sense to merge this PR to a feature branch, job-custom-fa, rather than the main branch?

yes, it makes totally sense. Not sure I can change this PR, or I need to make a new one

razo7:
> > Would it make more sense to merge this PR to a feature branch, job-custom-fa, rather than the main branch?
> 
> yes, it makes totally sense. Not sure I can change this PR, or I need to make a new one

I think you can change it, click on `edit` at the beginning of the page..

clobrano:
> I think you can change it, click on `edit` at the beginning of the page..

correct, and sorry for the noise of the other PRs :sweat:

Issue #90, opened 2023-10-10
Unable to fence nodes with `fence_azure_arm` agent


jcanocan:
Hi! 

I'm currently playing around with FAR with Azure VMs.  I've been able to install NHC, FAR in an OCP 4.13 cluster, to create the FAR Template and start the remediation process. This is the FAR Template I'm currently using:
```yaml
apiVersion: fence-agents-remediation.medik8s.io/v1alpha1
kind: FenceAgentsRemediationTemplate
metadata:
  name: fenceagentsremediationtemplate-default
  namespace: openshift-operators
spec:
  template:
    spec:
      sharedparameters:
        '--action': reboot
        '-l': ea6bxxx
        '-p': y~xxx
        '--resourceGroup': jcano-cluster-mfxww-rg
        '--tenantId': 60xxx
        '--subscriptionId': 89xxx
      nodeparameters:
        '--plug=':
          jcano-cluster-mfxww-master-0: jcano-cluster-mfxww-master-0
          jcano-cluster-mfxww-master-1: jcano-cluster-mfxww-master-1
          jcano-cluster-mfxww-master-2: jcano-cluster-mfxww-master-2
          jcano-cluster-mfxww-worker-germanywestcentral1-b58kw: jcano-cluster-mfxww-worker-germanywestcentral1-b58kw
          jcano-cluster-mfxww-worker-germanywestcentral2-h6zwd: jcano-cluster-mfxww-worker-germanywestcentral2-h6zwd
          jcano-cluster-mfxww-worker-germanywestcentral3-xd7h5: jcano-cluster-mfxww-worker-germanywestcentral3-xd7h5
      agent: fence_azure_arm
```
I've tried with fence_azure_arm tool standalone locally to restart a faulty VM where an OCP node is running. For that purpose, I stopped the kubelet process to bring a node to an unhealthy state, and it worked but requires a tiny modification, see: https://github.com/Azure/azure-sdk-for-python/issues/30983#issuecomment-1647081509

Nevertheless, it is not working along with FAR operator. It throws the following errors:

```sh
2023-10-10T15:08:07.128294848Z	INFO	controllers.FenceAgentsRemediation	Begin FenceAgentsRemediation Reconcile
2023-10-10T15:08:07.128341449Z	INFO	controllers.FenceAgentsRemediation	Check FAR CR's name
[...]

PR #91, opened 2023-10-11
Add fence-azure-arm agent

jcanocan:
Currently, if the agent fence_azure_arm is used, FAR throws an error showing that the `fence_azure_arm` is not found. It adds the Azure fence agent.

Fix: https://github.com/medik8s/fence-agents-remediation/issues/90


razo7:
/retest
/approve
Even though using the `fence-azure-arm agent` won't work due to a known bug, with the https://github.com/ClusterLabs/fence-agents/pull/562 fix it should work.
/lgtm

PR #92, opened 2023-10-12
Enable out-of-service taint in FAR

k-keiichi-rh:
This PR is adding a new remediation strategy based on https://github.com/kubernetes/enhancements/pull/1116

The following is the new remediation strategy for the out-of-service taint:
1. One of the nodes failed
2. FAR adds NoExecute taint to the failed node
=> Ensure that any workloads are not executed after rebooting the failed node
3. FAR reboots the failed node via the Fence Agent
=>  After rebooting, there are no stateless workloads which were not evicted by the taint in the failed node
4. FAR sets the out-of-service taint
=> This taint expects that the node is in shutdown or power off state (not in the middle of restarting).
5. After the failed node becomes healthy, the NoExecute taint in Step 2 and the out-of-service taint in Step 5 are removed and the node becomes schedulable again.

[ToDo]
- [x] Add unit tests
- [x] Add comments to clarify the changes of this patch
- [x] Add e2e test cases
- [x] Add a validation test of the Kubernetes version (https://github.com/medik8s/self-node-remediation/pull/104)

[ECOPROJECT-1326](https://issues.redhat.com/browse/ECOPROJECT-1326)

### DIFF
--- a/api/v1alpha1/fenceagentsremediation_types.go
+++ b/api/v1alpha1/fenceagentsremediation_types.go
@@ -44,10 +44,14 @@ const (
 	FenceAgentSucceeded ConditionsChangeReason = "FenceAgentSucceeded"
 	// RemediationFinishedSuccessfully - The unhealthy node was fully remediated/fenced (it was tainted, fenced by FA and all of its resources have been deleted)
 	RemediationFinishedSuccessfully ConditionsChangeReason = "RemediationFinishedSuccessfully"
+
+	ResourceDeletionRemediationStrategy  = RemediationStrategyType("ResourceDeletion")
+	OutOfServiceTaintRemediationStrategy = RemediationStrategyType("OutOfServiceTaint")
 )
 
 type ParameterName string
 type NodeName string
+type RemediationStrategyType string
 
 // FenceAgentsRemediationSpec defines the desired state of FenceAgentsRemediation
 type FenceAgentsRemediationSpec struct {
@@ -84,6 +88,15 @@ type FenceAgentsRemediationSpec struct {
 	// NodeParameters are passed to the fencing agent according to the node that is fenced, since they are node specific
 	//+operator-sdk:csv:customresourcedefinitions:type=spec
 	NodeParameters map[ParameterName]map[NodeName]string `json:"nodeparameters,omitempty"`
+
+	// RemediationStrategy is the remediation method for unhealthy nodes.
+	// Currently, it could be either "OutOfServiceTaint" or "ResourceDeletion".
+	// ResourceDeletion will iterate over all pods related to the unhealthy node and delete them.
+	// OutOfServiceTaint will add the out-of-service taint which is a new well-known taint "node.kubernetes.io/out-of-service"
+	// that enables automatic deletion of pv-attached pods on failed nodes, "out-of-service" taint is only supported on clusters with k8s version 1.26+ or OCP/OKD version 4.13+.
+	// +kubebuilder:default:="ResourceDeletion"
+	// +kubebuilder:validation:Enum=ResourceDeletion;OutOfServiceTaint
+	RemediationStrategy RemediationStrategyType `json:"remediationStrategy,omitempty"`
 }
 
 // FenceAgentsRemediationStatus defines the observed state of FenceAgentsRemediation
--- a/api/v1alpha1/fenceagentsremediation_webhook.go
+++ b/api/v1alpha1/fenceagentsremediation_webhook.go
@@ -53,19 +53,26 @@ var _ webhook.Validator = &FenceAgentsRemediation{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (far *FenceAgentsRemediation) ValidateCreate() (admission.Warnings, error) {
 	webhookFARLog.Info("validate create", "name", far.Name)
-	return validateAgentName(far.Spec.Agent)
+	return validateFAR(&far.Spec)
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (far *FenceAgentsRemediation) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
 	webhookFARLog.Info("validate update", "name", far.Name)
-	return validateAgentName(far.Spec.Agent)
+	return validateFAR(&far.Spec)
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type
 func (far *FenceAgentsRemediation) ValidateDelete() (admission.Warnings, error) {
 	webhookFARLog.Info("validate delete", "name", far.Name)
 	return nil, nil
+}
+
+func validateFAR(farSpec *FenceAgentsRemediationSpec) (admission.Warnings, error) {
+	if _, err := validateAgentName(farSpec.Agent); err != nil {
+		return nil, err
+	}
+	return validateStrategy(farSpec.RemediationStrategy)
 }
 
 func validateAgentName(agent string) (admission.Warnings, error) {
@@ -75,6 +82,13 @@ func validateAgentName(agent string) (admission.Warnings, error) {
 	}
 	if !exists {
 		return nil, fmt.Errorf("unsupported fence agent: %s", agent)
+	}
+	return nil, nil
+}
+
+func validateStrategy(farRemStrategy RemediationStrategyType) (admission.Warnings, error) {
+	if farRemStrategy == OutOfServiceTaintRemediationStrategy && !validation.IsOutOfServiceTaintSupported {
+		return nil, fmt.Errorf("%s remediation strategy is not supported at kubernetes version lower than 1.26, please use a different remediation strategy", OutOfServiceTaintRemediationStrategy)
 	}
 	return nil, nil
 }
--- a/api/v1alpha1/fenceagentsremediation_webhook.go
+++ b/api/v1alpha1/fenceagentsremediation_webhook.go
@@ -35,6 +35,8 @@ var (
 	webhookFARLog = logf.Log.WithName("fenceagentsremediation-resource")
 	// verify agent existence with os.Stat function
 	agentValidator = validation.NewAgentValidator()
+	// isOutOfServiceTaintSupported will be set to true in case out-of-service taint is supported (k8s 1.26 or higher)
+	isOutOfServiceTaintSupported bool
 )
 
 func (r *FenceAgentsRemediation) SetupWebhookWithManager(mgr ctrl.Manager) error {
@@ -75,6 +77,10 @@ func validateFAR(farSpec *FenceAgentsRemediationSpec) (admission.Warnings, error
 	return validateStrategy(farSpec.RemediationStrategy)
 }
 
+func InitOutOfServiceTaintSupportedFlag(outOfServiceTaintSupported bool) {
+	isOutOfServiceTaintSupported = outOfServiceTaintSupported
+}
+
 func validateAgentName(agent string) (admission.Warnings, error) {
 	exists, err := agentValidator.ValidateAgentName(agent)
 	if err != nil {
@@ -87,7 +93,7 @@ func validateAgentName(agent string) (admission.Warnings, error) {
 }
 
 func validateStrategy(farRemStrategy RemediationStrategyType) (admission.Warnings, error) {
-	if farRemStrategy == OutOfServiceTaintRemediationStrategy && !validation.IsOutOfServiceTaintSupported {
+	if farRemStrategy == OutOfServiceTaintRemediationStrategy && !isOutOfServiceTaintSupported {
 		return nil, fmt.Errorf("%s remediation strategy is not supported at kubernetes version lower than 1.26, please use a different remediation strategy", OutOfServiceTaintRemediationStrategy)
 	}
 	return nil, nil
--- a/api/v1alpha1/fenceagentsremediation_webhook_test.go
+++ b/api/v1alpha1/fenceagentsremediation_webhook_test.go
@@ -5,8 +5,6 @@ import (
 	. "github.com/onsi/gomega"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	"github.com/medik8s/fence-agents-remediation/pkg/validation"
 )
 
 var _ = Describe("FenceAgentsRemediation Validation", func() {
@@ -31,14 +29,14 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
 			var outOfServiceStrategy *FenceAgentsRemediation
 
 			BeforeEach(func() {
-				orgValue := validation.IsOutOfServiceTaintSupported
-				DeferCleanup(func() { validation.IsOutOfServiceTaintSupported = orgValue })
+				orgValue := isOutOfServiceTaintSupported
+				DeferCleanup(func() { isOutOfServiceTaintSupported = orgValue })
 
 				outOfServiceStrategy = getFAR(validAgentName, OutOfServiceTaintRemediationStrategy)
 			})
 			When("out of service taint is supported", func() {
 				BeforeEach(func() {
-					validation.IsOutOfServiceTaintSupported = true
+					isOutOfServiceTaintSupported = true
 				})
 				It("should be allowed", func() {
 					Expect(outOfServiceStrategy.ValidateCreate()).Error().NotTo(HaveOccurred())
@@ -46,7 +44,7 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
 			})
 			When("out of service taint is not supported", func() {
 				BeforeEach(func() {
-					validation.IsOutOfServiceTaintSupported = false
+					isOutOfServiceTaintSupported = false
 				})
 				It("should be denied", func() {
 					Expect(outOfServiceStrategy.ValidateCreate()).Error().To(MatchError(ContainSubstring(outOfServiceTaintUnsupportedMsg)))
@@ -82,15 +80,15 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
 			var resourceDeletionStrategy *FenceAgentsRemediation
 
 			BeforeEach(func() {
-				orgValue := validation.IsOutOfServiceTaintSupported
-				DeferCleanup(func() { validation.IsOutOfServiceTaintSupported = orgValue })
+				orgValue := isOutOfServiceTaintSupported
+				DeferCleanup(func() { isOutOfServiceTaintSupported = orgValue })
 
 				outOfServiceStrategy = getFAR(validAgentName, OutOfServiceTaintRemediationStrategy)
 				resourceDeletionStrategy = getFAR(validAgentName, ResourceDeletionRemediationStrategy)
 			})
 			When("out of service taint is supported", func() {
 				BeforeEach(func() {
-					validation.IsOutOfServiceTaintSupported = true
+					isOutOfServiceTaintSupported = true
 				})
 				It("should be allowed", func() {
 					Expect(outOfServiceStrategy.ValidateUpdate(resourceDeletionStrategy)).Error().NotTo(HaveOccurred())
@@ -98,7 +96,7 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
 			})
 			When("out of service taint is not supported", func() {
 				BeforeEach(func() {
-					validation.IsOutOfServiceTaintSupported = false
+					isOutOfServiceTaintSupported = false
 				})
 				It("should be denied", func() {
 					Expect(outOfServiceStrategy.ValidateUpdate(resourceDeletionStrategy)).Error().To(MatchError(ContainSubstring(outOfServiceTaintUnsupportedMsg)))
--- a/api/v1alpha1/fenceagentsremediation_webhook_test.go
+++ b/api/v1alpha1/fenceagentsremediation_webhook_test.go
@@ -5,6 +5,8 @@ import (
 	. "github.com/onsi/gomega"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/medik8s/fence-agents-remediation/pkg/validation"
 )
 
 var _ = Describe("FenceAgentsRemediation Validation", func() {
@@ -14,17 +16,41 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
 		When("agent name match format and binary", func() {
 			It("should be accepted", func() {
 				far := getTestFAR(validAgentName)
-				_, err := far.ValidateCreate()
-				Expect(err).ToNot(HaveOccurred())
+				Expect(far.ValidateCreate()).Error().NotTo(HaveOccurred())
 			})
 		})
 
 		When("agent name was not found ", func() {
 			It("should be rejected", func() {
 				far := getTestFAR(invalidAgentName)
-				_, err := far.ValidateCreate()
-				Expect(err).To(HaveOccurred())
-				Expect(err.Error()).To(ContainSubstring("unsupported fence agent: %s", invalidAgentName))
+				Expect(far.ValidateCreate()).Error().To(MatchError(ContainSubstring("unsupported fence agent: %s", invalidAgentName)))
+			})
+		})
+
+		Context("with OutOfServiceTaint strategy", func() {
+			var outOfServiceStrategy *FenceAgentsRemediation
+
+			BeforeEach(func() {
+				orgValue := validation.IsOutOfServiceTaintSupported
+				DeferCleanup(func() { validation.IsOutOfServiceTaintSupported = orgValue })
+
+				outOfServiceStrategy = getFAR(validAgentName, OutOfServiceTaintRemediationStrategy)
+			})
+			When("out of service taint is supported", func() {
+				BeforeEach(func() {
+					validation.IsOutOfServiceTaintSupported = true
+				})
+				It("should be allowed", func() {
+					Expect(outOfServiceStrategy.ValidateCreate()).Error().NotTo(HaveOccurred())
+				})
+			})
+			When("out of service taint is not supported", func() {
+				BeforeEach(func() {
+					validation.IsOutOfServiceTaintSupported = false
+				})
+				It("should be denied", func() {
+					Expect(outOfServiceStrategy.ValidateCreate()).Error().To(MatchError(ContainSubstring(outOfServiceTaintUnsupportedMsg)))
+				})
 			})
 		})
 	})
@@ -37,8 +63,7 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
 			})
 			It("should be accepted", func() {
 				far := getTestFAR(validAgentName)
-				_, err := far.ValidateUpdate(oldFAR)
-				Expect(err).ToNot(HaveOccurred())
+				Expect(far.ValidateUpdate(oldFAR)).Error().NotTo(HaveOccurred())
 			})
 		})
 
@@ -48,21 +73,53 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
 			})
 			It("should be rejected", func() {
 				far := getTestFAR(invalidAgentName)
-				_, err := far.ValidateUpdate(oldFAR)
-				Expect(err).To(HaveOccurred())
-				Expect(err.Error()).To(ContainSubstring("unsupported fence agent: %s", invalidAgentName))
+				Expect(far.ValidateUpdate(oldFAR)).Error().To(MatchError(ContainSubstring("unsupported fence agent: %s", invalidAgentName)))
+			})
+		})
+
+		Context("with OutOfServiceTaint strategy", func() {
+			var outOfServiceStrategy *FenceAgentsRemediation
+			var resourceDeletionStrategy *FenceAgentsRemediation
+
+			BeforeEach(func() {
+				orgValue := validation.IsOutOfServiceTaintSupported
+				DeferCleanup(func() { validation.IsOutOfServiceTaintSupported = orgValue })
+
+				outOfServiceStrategy = getFAR(validAgentName, OutOfServiceTaintRemediationStrategy)
+				resourceDeletionStrategy = getFAR(validAgentName, ResourceDeletionRemediationStrategy)
+			})
+			When("out of service taint is supported", func() {
+				BeforeEach(func() {
+					validation.IsOutOfServiceTaintSupported = true
+				})
+				It("should be allowed", func() {
+					Expect(outOfServiceStrategy.ValidateUpdate(resourceDeletionStrategy)).Error().NotTo(HaveOccurred())
+				})
+			})
+			When("out of service taint is not supported", func() {
+				BeforeEach(func() {
+					validation.IsOutOfServiceTaintSupported = false
+				})
+				It("should be denied", func() {
+					Expect(outOfServiceStrategy.ValidateUpdate(resourceDeletionStrategy)).Error().To(MatchError(ContainSubstring(outOfServiceTaintUnsupportedMsg)))
+				})
 			})
 		})
 	})
 })
 
 func getTestFAR(agentName string) *FenceAgentsRemediation {
+	return getFAR(agentName, ResourceDeletionRemediationStrategy)
+}
+
+func getFAR(agentName string, strategy RemediationStrategyType) *FenceAgentsRemediation {
 	return &FenceAgentsRemediation{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: "test-" + agentName,
 		},
 		Spec: FenceAgentsRemediationSpec{
-			Agent: agentName,
+			Agent:               agentName,
+			RemediationStrategy: strategy,
 		},
 	}
 }
--- a/api/v1alpha1/fenceagentsremediationtemplate_webhook.go
+++ b/api/v1alpha1/fenceagentsremediationtemplate_webhook.go
@@ -62,13 +62,13 @@ var _ webhook.Validator = &FenceAgentsRemediationTemplate{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (farTemplate *FenceAgentsRemediationTemplate) ValidateCreate() (admission.Warnings, error) {
 	webhookFARTemplateLog.Info("validate create", "name", farTemplate.Name)
-	return validateAgentName(farTemplate.Spec.Template.Spec.Agent)
+	return validateFAR(&farTemplate.Spec.Template.Spec)
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (farTemplate *FenceAgentsRemediationTemplate) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
 	webhookFARTemplateLog.Info("validate update", "name", farTemplate.Name)
-	return validateAgentName(farTemplate.Spec.Template.Spec.Agent)
+	return validateFAR(&farTemplate.Spec.Template.Spec)
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type
--- a/api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
+++ b/api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
@@ -5,8 +5,6 @@ import (
 	. "github.com/onsi/gomega"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	"github.com/medik8s/fence-agents-remediation/pkg/validation"
 )
 
 var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
@@ -31,15 +29,15 @@ var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
 			var outOfServiceStrategy *FenceAgentsRemediationTemplate
 
 			BeforeEach(func() {
-				orgValue := validation.IsOutOfServiceTaintSupported
-				DeferCleanup(func() { validation.IsOutOfServiceTaintSupported = orgValue })
+				orgValue := isOutOfServiceTaintSupported
+				DeferCleanup(func() { isOutOfServiceTaintSupported = orgValue })
 
 				outOfServiceStrategy = getFARTemplate(validAgentName, OutOfServiceTaintRemediationStrategy)
 			})
 
 			When("out of service taint is supported", func() {
 				BeforeEach(func() {
-					validation.IsOutOfServiceTaintSupported = true
+					isOutOfServiceTaintSupported = true
 				})
 				It("should be allowed", func() {
 					Expect(outOfServiceStrategy.ValidateCreate()).Error().NotTo(HaveOccurred())
@@ -48,7 +46,7 @@ var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
 
 			When("out of service taint is not supported", func() {
 				BeforeEach(func() {
-					validation.IsOutOfServiceTaintSupported = false
+					isOutOfServiceTaintSupported = false
 				})
 				It("should be denied", func() {
 					Expect(outOfServiceStrategy.ValidateCreate()).Error().To(MatchError(ContainSubstring(outOfServiceTaintUnsupportedMsg)))
@@ -84,8 +82,8 @@ var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
 			var resourceDeletionStrategy *FenceAgentsRemediationTemplate
 
 			BeforeEach(func() {
-				orgValue := validation.IsOutOfServiceTaintSupported
-				DeferCleanup(func() { validation.IsOutOfServiceTaintSupported = orgValue })
+				orgValue := isOutOfServiceTaintSupported
+				DeferCleanup(func() { isOutOfServiceTaintSupported = orgValue })
 
 				outOfServiceStrategy = getFARTemplate(validAgentName, OutOfServiceTaintRemediationStrategy)
 				resourceDeletionStrategy = getFARTemplate(validAgentName, ResourceDeletionRemediationStrategy)
@@ -93,7 +91,7 @@ var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
 
 			When("out of service taint is supported", func() {
 				BeforeEach(func() {
-					validation.IsOutOfServiceTaintSupported = true
+					isOutOfServiceTaintSupported = true
 				})
 				It("should be allowed", func() {
 					Expect(outOfServiceStrategy.ValidateUpdate(resourceDeletionStrategy)).Error().NotTo(HaveOccurred())
@@ -102,7 +100,7 @@ var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
 
 			When("out of service taint is not supported", func() {
 				BeforeEach(func() {
-					validation.IsOutOfServiceTaintSupported = false
+					isOutOfServiceTaintSupported = false
 				})
 				It("should be denied", func() {
 					Expect(outOfServiceStrategy.ValidateUpdate(resourceDeletionStrategy)).Error().To(MatchError(ContainSubstring(outOfServiceTaintUnsupportedMsg)))
--- a/api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
+++ b/api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
@@ -5,6 +5,8 @@ import (
 	. "github.com/onsi/gomega"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/medik8s/fence-agents-remediation/pkg/validation"
 )
 
 var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
@@ -14,17 +16,43 @@ var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
 		When("agent name match format and binary", func() {
 			It("should be accepted", func() {
 				farTemplate := getTestFARTemplate(validAgentName)
-				_, err := farTemplate.ValidateCreate()
-				Expect(err).ToNot(HaveOccurred())
+				Expect(farTemplate.ValidateCreate()).Error().NotTo(HaveOccurred())
 			})
 		})
 
 		When("agent name was not found ", func() {
 			It("should be rejected", func() {
 				farTemplate := getTestFARTemplate(invalidAgentName)
-				_, err := farTemplate.ValidateCreate()
-				Expect(err).To(HaveOccurred())
-				Expect(err.Error()).To(ContainSubstring("unsupported fence agent: %s", invalidAgentName))
+				Expect(farTemplate.ValidateCreate()).Error().To(MatchError(ContainSubstring("unsupported fence agent: %s", invalidAgentName)))
+			})
+		})
+
+		Context("with OutOfServiceTaint strategy", func() {
+			var outOfServiceStrategy *FenceAgentsRemediationTemplate
+
+			BeforeEach(func() {
+				orgValue := validation.IsOutOfServiceTaintSupported
+				DeferCleanup(func() { validation.IsOutOfServiceTaintSupported = orgValue })
+
+				outOfServiceStrategy = getFARTemplate(validAgentName, OutOfServiceTaintRemediationStrategy)
+			})
+
+			When("out of service taint is supported", func() {
+				BeforeEach(func() {
+					validation.IsOutOfServiceTaintSupported = true
+				})
+				It("should be allowed", func() {
+					Expect(outOfServiceStrategy.ValidateCreate()).Error().NotTo(HaveOccurred())
+				})
+			})
+
+			When("out of service taint is not supported", func() {
+				BeforeEach(func() {
+					validation.IsOutOfServiceTaintSupported = false
+				})
+				It("should be denied", func() {
+					Expect(outOfServiceStrategy.ValidateCreate()).Error().To(MatchError(ContainSubstring(outOfServiceTaintUnsupportedMsg)))
+				})
 			})
 		})
 	})
@@ -37,8 +65,7 @@ var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
 			})
 			It("should be accepted", func() {
 				farTemplate := getTestFARTemplate(validAgentName)
-				_, err := farTemplate.ValidateUpdate(oldFARTemplate)
-				Expect(err).ToNot(HaveOccurred())
+				Expect(farTemplate.ValidateUpdate(oldFARTemplate)).Error().NotTo(HaveOccurred())
 			})
 		})
 
@@ -48,15 +75,48 @@ var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
 			})
 			It("should be rejected", func() {
 				farTemplate := getTestFARTemplate(invalidAgentName)
-				_, err := farTemplate.ValidateUpdate(oldFARTemplate)
-				Expect(err).To(HaveOccurred())
-				Expect(err.Error()).To(ContainSubstring("unsupported fence agent: %s", invalidAgentName))
+				Expect(farTemplate.ValidateUpdate(oldFARTemplate)).Error().To(MatchError(ContainSubstring("unsupported fence agent: %s", invalidAgentName)))
+			})
+		})
+
+		Context("with OutOfServiceTaint strategy", func() {
+			var outOfServiceStrategy *FenceAgentsRemediationTemplate
+			var resourceDeletionStrategy *FenceAgentsRemediationTemplate
+
+			BeforeEach(func() {
+				orgValue := validation.IsOutOfServiceTaintSupported
+				DeferCleanup(func() { validation.IsOutOfServiceTaintSupported = orgValue })
+
+				outOfServiceStrategy = getFARTemplate(validAgentName, OutOfServiceTaintRemediationStrategy)
+				resourceDeletionStrategy = getFARTemplate(validAgentName, ResourceDeletionRemediationStrategy)
+			})
+
+			When("out of service taint is supported", func() {
+				BeforeEach(func() {
+					validation.IsOutOfServiceTaintSupported = true
+				})
+				It("should be allowed", func() {
+					Expect(outOfServiceStrategy.ValidateUpdate(resourceDeletionStrategy)).Error().NotTo(HaveOccurred())
+				})
+			})
+
+			When("out of service taint is not supported", func() {
+				BeforeEach(func() {
+					validation.IsOutOfServiceTaintSupported = false
+				})
+				It("should be denied", func() {
+					Expect(outOfServiceStrategy.ValidateUpdate(resourceDeletionStrategy)).Error().To(MatchError(ContainSubstring(outOfServiceTaintUnsupportedMsg)))
+				})
 			})
 		})
 	})
 })
 
 func getTestFARTemplate(agentName string) *FenceAgentsRemediationTemplate {
+	return getFARTemplate(agentName, ResourceDeletionRemediationStrategy)
+}
+
+func getFARTemplate(agentName string, strategy RemediationStrategyType) *FenceAgentsRemediationTemplate {
 	return &FenceAgentsRemediationTemplate{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: "test-" + agentName + "-template",
@@ -64,7 +124,8 @@ func getTestFARTemplate(agentName string) *FenceAgentsRemediationTemplate {
 		Spec: FenceAgentsRemediationTemplateSpec{
 			Template: FenceAgentsRemediationTemplateResource{
 				Spec: FenceAgentsRemediationSpec{
-					Agent: agentName,
+					Agent:               agentName,
+					RemediationStrategy: strategy,
 				},
 			},
 		},
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -166,6 +166,23 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 			r.Executor.Remove(far.GetUID())
 		}
 
+		// remove out-of-service taint when using OutOfServiceTaint remediation
+		if far.Spec.RemediationStrategy == v1alpha1.OutOfServiceTaintRemediationStrategy {
+			r.Log.Info("Removing out-of-service taint", "Fence Agent", far.Spec.Agent, "Node Name", node.Name)
+			taint := utils.CreateOutOfServiceTaint()
+			if err := utils.RemoveTaint(r.Client, node.Name, taint); err != nil {
+				if apiErrors.IsConflict(err) {
+					r.Log.Error(err, "Failed to remove taint from node due to node update, retrying... ,", "node name", node.Name, "taint key", taint.Key, "taint effect", taint.Effect)
+					return ctrl.Result{RequeueAfter: time.Second}, nil
+				} else if !apiErrors.IsNotFound(err) {
+					r.Log.Error(err, "Failed to remove taint from node,", "node name", node.Name, "taint key", taint.Key, "taint effect", taint.Effect)
+					return emptyResult, err
+				}
+			}
+			r.Log.Info("out-of-service taint was removed", "Node Name", req.Name)
+			commonEvents.NormalEvent(r.Recorder, node, utils.EventReasonRemoveOutOfServiceTaint, utils.EventMessageRemoveOutOfServiceTaint)
+		}
+
 		// remove node's taints
 		taint := utils.CreateRemediationTaint()
 		if err := utils.RemoveTaint(r.Client, node.Name, taint); err != nil {
@@ -191,7 +208,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		return emptyResult, nil
 	}
 	// Add FAR (medik8s) remediation taint
-	taintAdded, err := utils.AppendTaint(r.Client, node.Name)
+	taintAdded, err := utils.AppendTaint(r.Client, node.Name, utils.CreateRemediationTaint())
 	if err != nil {
 		return emptyResult, err
 	} else if taintAdded {
@@ -228,11 +245,32 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		// - try to remove workloads
 		// - clean up Executor routine
 
-		r.Log.Info("Manual workload deletion", "Node Name", node.Name)
-		commonEvents.NormalEvent(r.Recorder, node, utils.EventReasonDeleteResources, utils.EventMessageDeleteResources)
-		if err := commonResources.DeletePods(ctx, r.Client, node.Name); err != nil {
-			r.Log.Error(err, "Manual workload deletion has failed", "CR's Name", req.Name)
-			return emptyResult, err
+		switch far.Spec.RemediationStrategy {
+		case v1alpha1.ResourceDeletionRemediationStrategy, "":
+			// Basically RemediationStrategy should be set to ResourceDeletion strategy as the default strategy.
+			// However it will be empty when the CS was created when ResourceDeletion strategy was the only strategy.
+			// In this case, the empty strategy should be treated as if ResourceDeletion strategy selected.
+			r.Log.Info("Remediation strategy is ResourceDeletion which explicitly deletes resources - manually deleting workload", "Node Name", req.Name)
+			commonEvents.NormalEvent(r.Recorder, node, utils.EventReasonDeleteResources, utils.EventMessageDeleteResources)
+			if err := commonResources.DeletePods(ctx, r.Client, node.Name); err != nil {
+				r.Log.Error(err, "Resource deletion has failed", "CR's Name", node.Name)
+				return emptyResult, err
+			}
+		case v1alpha1.OutOfServiceTaintRemediationStrategy:
+			r.Log.Info("Remediation strategy is OutOfServiceTaint which implicitly deletes resources - adding out-of-service taint", "Node Name", req.Name)
+			taintAdded, err := utils.AppendTaint(r.Client, node.Name, utils.CreateOutOfServiceTaint())
+			if err != nil {
+				r.Log.Error(err, "Failed to add out-of-service taint", "CR's Name", node.Name)
+				return emptyResult, err
+			} else if taintAdded {
+				r.Log.Info("out-of-service taint was added", "Node Name", node.Name)
+				commonEvents.NormalEvent(r.Recorder, node, utils.EventReasonAddOutOfServiceTaint, utils.EventMessageAddOutOfServiceTaint)
+			}
+		default:
+			//this should never happen since we enforce valid values with kubebuilder
+			err := errors.New("unsupported remediation strategy")
+			r.Log.Error(err, "Encountered unsupported remediation strategy. Please check template spec", "strategy", far.Spec.RemediationStrategy)
+			return emptyResult, nil
 		}
 		utils.UpdateConditions(utils.RemediationFinishedSuccessfully, far, r.Log)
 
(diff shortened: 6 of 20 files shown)


controllers/fenceagentsremediation_controller.go
```diff
@@ -157,8 +157,17 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 				"fenceAgentActionSucceeded condition", fenceAgentActionSucceededCondition, "succeeded condition", succeededCondition)
 		}
 
+		// remove out-of-service taint when using OutOfServiceTaint remediation
+		if far.Spec.RemediationStrategy == v1alpha1.OutOfServiceTaintRemediationStrategy {
+			r.Log.Info("Removing OutOfService taint", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
+			if err := utils.RemoveTaint(r.Client, far.Name, utils.CreateOutOfServiceTaint()); err != nil && !apiErrors.IsNotFound(err) {
+				r.Log.Error(err, "Failed to add out-of-service taint", "CR's Name", req.Name)
```
razo7, line 164: `Failed to add` -> `Failed to remove`

controllers/fenceagentsremediation_controller.go
```diff
@@ -221,10 +230,22 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 	}
 	if meta.IsStatusConditionTrue(far.Status.Conditions, v1alpha1.FenceAgentActionSucceededType) &&
 		!meta.IsStatusConditionTrue(far.Status.Conditions, commonConditions.SucceededType) {
-		// Fence agent action succeeded, and now we try to remove workloads (pods and their volume attachments)
-		r.Log.Info("Manual workload deletion", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
-		if err := utils.DeleteResources(ctx, r.Client, req.Name); err != nil {
-			r.Log.Error(err, "Manual workload deletion has failed", "CR's Name", req.Name)
+		if far.Spec.RemediationStrategy == v1alpha1.ResourceDeletionRemediationStrategy {
+			// Fence agent action succeeded, and now we try to remove workloads (pods and their volume attachments)
```
razo7, line 234: NIT: This comment can and should be between lines 232 and 233 as it is not specific to _ResourceDeletionRemediationStrategy_.

controllers/fenceagentsremediation_controller.go
```diff
@@ -221,10 +230,22 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
+			r.Log.Info("Manual workload deletion", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
```
razo7, line 235: It might be best to change the term `Manual workload deletion` to `Resource deletion - manually deleting workload`, since there are `OutOfServiceTaint` and `ResourceDeletion` remediation types

controllers/fenceagentsremediation_controller.go
```diff
@@ -221,10 +230,22 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
+			if err := utils.DeleteResources(ctx, r.Client, req.Name); err != nil {
+				r.Log.Error(err, "Manual workload deletion has failed", "CR's Name", req.Name)
```
razo7, line 237: Same here.
`Manual workload deletion has failed` -> `Resource deletion has failed`

pkg/utils/taints.go
```diff
@@ -51,15 +51,22 @@ func CreateFARNoExecuteTaint() corev1.Taint {
 	}
 }
 
+// CreateOutOfServiceTaint returns an OutOfService taint
+func CreateOutOfServiceTaint() corev1.Taint {
+	return corev1.Taint{
+		Key:    v1alpha1.OutOfServiceTaintKey,
+		Effect: corev1.TaintEffectNoExecute,
```
razo7, line 58: IIUC the taint also has a value (`nodeshutdown`) that we should include in the taint. This addition won't affect the taint effect, since the taint is unique by Key and Effect, but I believe it makes sense to follow standards as using this value https://kubernetes.io/blog/2023/08/16/kubernetes-1-28-non-graceful-node-shutdown-ga/#what-is-a-non-graceful-node-shutdown

>kubectl taint nodes <node-name> node.kubernetes.io/out-of-service=nodeshutdown:NoExecute

controllers/fenceagentsremediation_controller.go
```diff
@@ -221,10 +230,22 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 	}
 	if meta.IsStatusConditionTrue(far.Status.Conditions, v1alpha1.FenceAgentActionSucceededType) &&
 		!meta.IsStatusConditionTrue(far.Status.Conditions, commonConditions.SucceededType) {
-		// Fence agent action succeeded, and now we try to remove workloads (pods and their volume attachments)
-		r.Log.Info("Manual workload deletion", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
-		if err := utils.DeleteResources(ctx, r.Client, req.Name); err != nil {
-			r.Log.Error(err, "Manual workload deletion has failed", "CR's Name", req.Name)
+		if far.Spec.RemediationStrategy == v1alpha1.ResourceDeletionRemediationStrategy {
```
razo7, line 233: NIT: Switch-case might be better readable inside an if instead of an if-else clause.

k-keiichi-rh:
> > => This taint expects that the node is in shutdown or power off state (not in the middle of restarting).
> 
> Moreover, regarding your comment, I am not sure whether the node will be in a shutdown or power-off state when FAR adds the out-of-service taint. The only supported fencing action that FAR has is `reboot` which would do power off and then power on. Therefore, the node won't be in your desired state after the fence agent succeeds.

We can use the same approach of SNR we discussed in [here](https://github.com/medik8s/self-node-remediation/issues/17#issuecomment-1155947689). 

There are the following cases after the `reboot` action in FAR:

1. The failed node is rebooted and it becomes healthy again
   => The out-of-service taint doesn't take effect. The taint is ignored.  The node can report its status to the control plane and the control plane can delete the stateful workload instead of the out-of-service taint.
2. The failed node is rebooted, but it keeps unhealthy(the node can not report its status to the control plane)
   => **The out-of-service taint takes effect.**
   => The taint will trigger deleting the stateful workloads.
3. The failed node is not rebooted due to either failing to do power off or failing to do then power on
   => The out-of-service taint doesn't take effect because it's not added to the failed node. 
   => FAR is checking the result of executing the fence agent. If it's failed, the exponential backoff is triggered

> One more thing to raise/add is whether we want to do a validation test of the Kubernetes version ([similar to what SNR does](https://github.com/medik8s/self-node-remediation/pull/104)) since the out-of-service taint is fairly new in the community and not supported in old versions.

I will add this topic to my todo list. Thank you for sharing it.

razo7:
> We can use the same approach of SNR we discussed in https://github.com/medik8s/self-node-remediation/issues/17#issuecomment-1155947689.

SGTM

pkg/utils/taints.go
razo7: It was addressed above :)

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -271,6 +357,22 @@ func createVA(vaName, unhealthyNodeName string) *storagev1.VolumeAttachment {
 	return va
 }
 
+// deleteVA delete the specified volume attachment
```
razo7, line 360: NIT : `delete` -> `deletes`

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -271,6 +357,22 @@ func createVA(vaName, unhealthyNodeName string) *storagev1.VolumeAttachment {
...
+func deleteVA(vaName, nodeName string) {
```
razo7, line 361: I think function `deleteVA` could be minimized, since there is no need to create the _ObjectMeta_ and _Spec_ of an already created VA in order to delete it.
In line 234 we can assign the VA to a variable (`va1 := createVA(vaName1, workerNode)`) and then in line 252 simply call `deleteVA(va1)` to run the _ExpectWithOffset_ command.

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -252,6 +312,32 @@ func createRunningPod(containerName, podName, nodeName string) *corev1.Pod {
...
+	}
+
+	var grace client.GracePeriodSeconds = 0
+	ExpectWithOffset(1, k8sClient.Delete(context.Background(), pod, grace)).To(Succeed())
+
+	EventuallyWithOffset(1, func() bool {
+		err := k8sClient.Get(context.Background(), podKey, pod)
+		return apierrors.IsNotFound(err)
+	}, 10*time.Second, 100*time.Millisecond).Should(BeTrue())
+}
```
razo7, line 339: Unless I am missing something but the timeout could be shorten and poll interval could be higher for deleting objects. Since it is used in _Eventually_, then it might not make any change for running the unit tests in most cases.

Can you use new consts for these variables or use `timeoutDeletion` and `pollInterval` consts?
(this comment is true for the other two apperances of `10*time.Second`, and `100*time.Millisecond`)

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -289,10 +391,28 @@ func cleanupTestedResources(va1, va2 *storagev1.VolumeAttachment, pod *corev1.Po
 	podTest := &corev1.Pod{}
 	if err := k8sClient.Get(context.Background(), client.ObjectKeyFromObject(pod), podTest); err == nil {
 		log.Info("Cleanup: clean pod", "pod name", podTest.Name)
-		Expect(k8sClient.Delete(context.Background(), podTest)).To(Succeed())
+		var grace client.GracePeriodSeconds = 0
```
razo7, line 394: Can it be a const then it will be used in many functions and not per function whenever we delete a pod?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -252,6 +312,32 @@ func createRunningPod(containerName, podName, nodeName string) *corev1.Pod {
...
 
+func createTerminatingPod(containerName, podName, nodeName string) *corev1.Pod {
+	pod := buildPod(containerName, podName, nodeName)
+	now := metav1.Now()
+	pod.ObjectMeta = metav1.ObjectMeta{Name: pod.Name, Namespace: pod.Namespace, DeletionTimestamp: &now}
+	ExpectWithOffset(1, k8sClient.Create(context.Background(), pod)).To(Succeed())
+	return pod
+}
+
+func deleteTerminatingPod(podName string) {
```
razo7, line 323: I think function `deleteTerminatingPod`` could be minimized, since there is no need to create a new pod based on its name in order to delete it.

In line 235 we can assign the pod to a variable (`podTerminate1 := createTerminatingPod("far-terminate-1", testPodName, workerNode)`) and then in line 253 simply call `deleteTerminatingPod(podTerminate1)` to crate nodeKey, delete the inputed pod and verify that it is gone.

controllers/fenceagentsremediation_controller.go
```diff
@@ -222,9 +234,22 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
-			r.Log.Error(err, "Manual workload deletion has failed", "CR's Name", req.Name)
+		switch far.Spec.RemediationStrategy {
+		case v1alpha1.ResourceDeletionRemediationStrategy:
+			r.Log.Info("Resource deletion - manually deleting workload", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
+			if err := utils.DeleteResources(ctx, r.Client, req.Name); err != nil {
+				r.Log.Error(err, "Resource deletion has failed", "CR's Name", req.Name)
+				return emptyResult, err
+			}
+		case v1alpha1.OutOfServiceTaintRemediationStrategy:
+			r.Log.Info("Resource deletion - adding OutOfService taint", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
```
razo7, line 245: I think it is a bit confusing to state resource deletion for OutOfService taint since the other remediation strategy is called `Resource deletion` (even though the taint causes a resource deletion).

Therefore I suggest `Resource deletion - adding OutOfService taint` -> `Remediation strategy is OutOfService taint which implicitly deletes resources - adding OutOfService taint`.

controllers/fenceagentsremediation_controller.go
```diff
@@ -222,9 +234,22 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 	if meta.IsStatusConditionTrue(far.Status.Conditions, v1alpha1.FenceAgentActionSucceededType) &&
 		!meta.IsStatusConditionTrue(far.Status.Conditions, commonConditions.SucceededType) {
 		// Fence agent action succeeded, and now we try to remove workloads (pods and their volume attachments)
-		r.Log.Info("Manual workload deletion", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
-		if err := utils.DeleteResources(ctx, r.Client, req.Name); err != nil {
-			r.Log.Error(err, "Manual workload deletion has failed", "CR's Name", req.Name)
+		switch far.Spec.RemediationStrategy {
+		case v1alpha1.ResourceDeletionRemediationStrategy:
+			r.Log.Info("Resource deletion - manually deleting workload", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
```
razo7, line 239: See my other comment on the OutOfService log message below.

`Resource deletion - manually deleting workload` -> `Remediation strategy is resource deletion which explicitly deletes resources - manually deleting workload`.

razo7 (commented):
Thank you for addressing my last comments!

I have added some more comments :) Mostly minor nits on phrasing, missing log, consts, and simulating the deletion of pod and VA.

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -252,6 +312,32 @@ func createRunningPod(containerName, podName, nodeName string) *corev1.Pod {
 	return pod
 }
 
+func createTerminatingPod(containerName, podName, nodeName string) *corev1.Pod {
+	pod := buildPod(containerName, podName, nodeName)
+	now := metav1.Now()
+	pod.ObjectMeta = metav1.ObjectMeta{Name: pod.Name, Namespace: pod.Namespace, DeletionTimestamp: &now}
+	ExpectWithOffset(1, k8sClient.Create(context.Background(), pod)).To(Succeed())
```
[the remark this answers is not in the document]
slintes: @razo7 the offset improves Gomega logs in case an `Expect` in helper functions like this fails. 
See https://onsi.github.io/gomega/#making-assertions-in-helper-functions
(Maybe we can better use `GinkgoHelper()` instead though, which I only just learned about)

slintes:
> > > => This taint expects that the node is in shutdown or power off state (not in the middle of restarting).
> > 
> > 
> > Moreover, regarding your comment, I am not sure whether the node will be in a shutdown or power-off state when FAR adds the out-of-service taint. The only supported fencing action that FAR has is `reboot` which would do power off and then power on. Therefore, the node won't be in your desired state after the fence agent succeeds.
> 
> We can use the same approach of SNR we discussed in [here](https://github.com/medik8s/self-node-remediation/issues/17#issuecomment-1155947689).

I'm not sure if the same arguments as stated on SNR apply for FAR. The timing is different:

SNR:
- remediation starts, node reboot is triggered
- some time expires
- node reboot completed
- _**some more time expires until safeTimeToAssumeNodeRebooted is reached**_ 
- taint is added only when node is still unhealthy

FAR:
- remediation starts, node reboot is triggered
- _**the taint is added immediately**_
- some time expires
- node reboot completed
- node might be healthy now but has the taint already. I understand this should be avoided.

controllers/fenceagentsremediation_controller_test.go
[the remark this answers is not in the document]
razo7: IIUC `GinkgoHelper()` is helpful for better distinguishing between multiple assertions in one function, and here we have only one.

k-keiichi-rh:
> I'm not sure if the same arguments as stated on SNR apply for FAR. The timing is different:

I may not understand your point correctly. So please let me confirm it just in case.
 
> SNR:
> 
> * remediation starts, node reboot is triggered
> * some time expires
> * node reboot completed
> * _**some more time expires until safeTimeToAssumeNodeRebooted is reached**_
> * taint is added only when node is still unhealthy

In the current OutOfService remediation in SNR, the out-of-service taint is added to the node who becomes healthy after the node reboot. However the out-of-service taint is deleted right after checking if there is no stateful workload on the node.

So should we avoid adding the out-of-service taint to the healthy node by checking if the SNR CR is being deleted by NHC/MHC?

> FAR:
> 
> * remediation starts, node reboot is triggered
> * _**the taint is added immediately**_
> * some time expires
> * node reboot completed
> * node might be healthy now but has the taint already. I understand this should be avoided.

If the node becomes healthy again, the FAR CR is deleted by NHC/MHC and the recovery action(deleting the out-of-service taint) is also executed. In this case, the healthy node won't have the out-of-service taint. So the node will come back to the cluster again.
So we can avoid the situation that the healthy node will keep having the out-of-service taint after rebooting.

slintes:
> > I'm not sure if the same arguments as stated on SNR apply for FAR. The timing is different:
> 
> I may not understand your point correctly. So please let me confirm it just in case.
> 
> > SNR:
> > 
> > * remediation starts, node reboot is triggered
> > * some time expires
> > * node reboot completed
> > * _**some more time expires until safeTimeToAssumeNodeRebooted is reached**_
> > * taint is added only when node is still unhealthy
> 
> In the current OutOfService remediation in SNR, the out-of-service taint is added to the node who becomes healthy after the node reboot. However the out-of-service taint is deleted right after checking if there is no stateful workload on the node.
> 
> So should we avoid adding the out-of-service taint to the healthy node by checking if the SNR CR is being deleted by NHC/MHC?

I thought we already do this, but just double checked the code, and we don't.
Yes, I think we should stop any further fencing action when the CR has the deletion timestamp set. We should just do cleanup.
I understood that rebooting already is a grey area for using the taint, because of "This taint expects that the node is in shutdown or power off state (not in the middle of restarting)".
Isn't putting the taint on a node which finished rebooting, and is healthy now, an issue? Even when we remove it afterwards?

/cc @mshitrit fyi

> 
> > FAR:
> > 
> > * remediation starts, node reboot is triggered
> > * _**the taint is added immediately**_
> > * some time expires
> > * node reboot completed
> > * node might be healthy now but has the taint already. I understand this should be avoided.
> 
> If the node becomes healthy again, the FAR CR is deleted by NHC/MHC and the recovery action(deleting the out-of-service taint) is also executed. In this case, the healthy node won't have the out-of-service taint. So the node will come back to the cluster again. So we can avoid the situation that the healthy node will keep having the out-of-service taint after rebooting.

k-keiichi-rh:
> Yes, I think we should stop any further fencing action when the CR has the deletion timestamp set. We should just do cleanup.
I understood that rebooting already is a grey area for using the taint, because of "This taint expects that the node is in shutdown or power off state (not in the middle of restarting)".

I agree with you. I will fix it to the out-of-service taint remediation and check if there is no side effect by the change.
As for stopping any further fencing action on the healthy node, I think the basic idea here is that the control-plane should handle  the fencing action if the failed node can communicate with the control-plane. So we don't need to do nothing in SNR.
If yes, does the same apply to the Resource Deletion Remediation as well as the OutOfServiceTaint Remediation?

> Isn't putting the taint on a node which finished rebooting, and is healthy now, an issue? Even when we remove it afterwards?

As far as I checked [the effect of the out-of-service taint](https://github.com/medik8s/self-node-remediation/issues/16#issuecomment-1158188745), putting the taint is not an issue and has no side effect.

In the "After rebooting" phase of SNR, the failed node has both the normal NoExecute taint and the NoSchedule taint and we expect that there are no stateful workloads on the node. So the out-of-service taint won't do anything.

slintes:
> > Yes, I think we should stop any further fencing action when the CR has the deletion timestamp set. We should just do cleanup.
> > I understood that rebooting already is a grey area for using the taint, because of "This taint expects that the node is in shutdown or power off state (not in the middle of restarting)".
> 
> I agree with you. I will fix it to the out-of-service taint remediation and check if there is no side effect by the change. As for stopping any further fencing action on the healthy node, I think the basic idea here is that the control-plane should handle the fencing action if the failed node can communicate with the control-plane. So we don't need to do nothing in SNR. If yes, does the same apply to the Resource Deletion Remediation as well as the OutOfServiceTaint Remediation?

- not sure if I understand, what do you mean with "the control-plane should handle the fencing action"?
- I think we need to do some "cleanup" in SNR, e.g removing taints which were already set in the pre-reboot phase. Maybe we can just switch to the fencing completed phase directly, it should do everything we need for cleanup?
- yes, I think the the same applies for the ResourceDeletion strategy
- before changing anything, we should wait for comment of @mshitrit 
- I will create an issue for SNR to have the discussion at the right place 🙂 

> 
> > Isn't putting the taint on a node which finished rebooting, and is healthy now, an issue? Even when we remove it afterwards?
> 
> As far as I checked [the effect of the out-of-service taint](https://github.com/medik8s/self-node-remediation/issues/16#issuecomment-1158188745), putting the taint is not an issue and has no side effect.
> 
> In the "After rebooting" phase of SNR, the failed node has both the normal NoExecute taint and the NoSchedule taint and we expect that there are no stateful workloads on the node. So the out-of-service taint won't do anything.

Ok, then my concerns were wrong, and it that makes the SNR topic much less urgent. Sorry for the noise and thanks for the discussion!

slintes:
for the SNR related discussion let's continue here: https://github.com/medik8s/self-node-remediation/issues/159

k-keiichi-rh:
> I have added some more comments :) Mostly minor nits on phrasing, missing log, consts, and simulating the deletion of pod and VA. Please add a new commit after each review so it will be easier to review the changes between the commits and the last review.

@razo7 Thank you for taking your time and your review.
I reflected your comments. Please check it.

By the way, are my replies to your comments visible?
My replies are noted with the "Pending" tag.

razo7:
>By the way, are my replies to your comments visible?
My replies are noted with the "Pending" tag.

No, I can't see your replies since you haven't submitted your review. Please see [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request#submitting-your-review) on how to submit them.

controllers/fenceagentsremediation_controller_test.go
razo7: Thanks :)

k-keiichi-rh (commented):
Sorry that I have not submitted my review.
I reflected all your comments and there is no questions for the comments.

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -209,16 +211,77 @@ var _ = Describe("FAR Controller", func() {
...
+				Expect(underTestFAR.Status.LastUpdateTime).ToNot(BeNil())
+				verifyStatusCondition(workerNode, commonConditions.ProcessingType, conditionStatusPointer(metav1.ConditionFalse))
+				verifyStatusCondition(workerNode, v1alpha1.FenceAgentActionSucceededType, conditionStatusPointer(metav1.ConditionTrue))
+				verifyStatusCondition(workerNode, commonConditions.SucceededType, conditionStatusPointer(metav1.ConditionTrue))
+
+				By("Having no remediation taint and no out-of-service taint")
+				deleteFAR(underTestFAR)
+				Eventually(func(g Gomega) bool {
+					g.Expect(k8sClient.Get(context.Background(), nodeKey, node)).To(Succeed())
+					return utils.TaintExists(node.Spec.Taints, &farNoExecuteTaint) && utils.TaintExists(node.Spec.Taints, &outOfServiceTaint)
```
slintes, line 268: ` &&`  should be ` ||`  I guess , otherwise the test succeeds if one taint was removed only

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -209,16 +211,77 @@ var _ = Describe("FAR Controller", func() {
...
+			DeferCleanup(deletePod, farPod)
+		})
+		JustBeforeEach(func() {
+			// Create node, and at the end clean it up with DeferCleanup
+			Expect(k8sClient.Create(context.Background(), node)).To(Succeed())
+			DeferCleanup(k8sClient.Delete, context.Background(), node)
+		})
+
+		When("creating valid FAR CR", func() {
+			var va *storagev1.VolumeAttachment
```
slintes, line 231: I think everything about VAs can be removed, see https://github.com/medik8s/fence-agents-remediation/pull/102

slintes (changes requested):
Mostly straightforward PR :) Needs rebase though.
Some comments inline.

pkg/utils/taints.go
```diff
@@ -47,19 +47,28 @@ func deleteTaint(taints []corev1.Taint, taintToDelete *corev1.Taint) ([]corev1.T
 func CreateFARNoExecuteTaint() corev1.Taint {
 	return corev1.Taint{
 		Key:    v1alpha1.FARNoExecuteTaintKey,
+		Value:  "fence-agents-remediation",
```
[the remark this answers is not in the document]
k-keiichi-rh: I tried to change the same thing as [SNR](https://github.com/medik8s/self-node-remediation/blob/b5894be3fe86c4a8c66c40ef5691ed2d599b5ab8/controllers/selfnoderemediation_controller.go#L75):
```
	NodeNoExecuteTaint = &v1.Taint{
		Key:    "medik8s.io/remediation",
		Value:  "self-node-remediation",
		Effect: v1.TaintEffectNoExecute,
	}
```
But I failed to do that. Anyway I will drop this change in the next patch.

pkg/utils/resources.go
```diff
@@ -72,3 +72,31 @@ func DeleteResources(ctx context.Context, r client.Client, nodeName string) erro
 
 	return nil
 }
+
+func IsResourceDeletionCompleted(r client.Client, nodeName string) bool {
```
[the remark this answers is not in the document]
k-keiichi-rh: I will do this.

pkg/utils/resources.go
```diff
@@ -72,3 +72,31 @@ func DeleteResources(ctx context.Context, r client.Client, nodeName string) erro
...
+	pods := &corev1.PodList{}
+	if err := r.List(context.Background(), pods); err != nil {
+		log.Error(err, "failed to get pod list")
+		return false
```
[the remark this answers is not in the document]
k-keiichi-rh: I agree. I will change it.

pkg/utils/resources.go
```diff
@@ -72,3 +72,31 @@ func DeleteResources(ctx context.Context, r client.Client, nodeName string) erro
...
+	}
+	for _, pod := range pods.Items {
+		if pod.Spec.NodeName == nodeName && pod.ObjectMeta.DeletionTimestamp != nil {
+			log.Info("waiting for terminating pod ", "pod name", pod.Name, "phase", pod.Status.Phase)
+			return false
+		}
+	}
+
+	volumeAttachments := &storagev1.VolumeAttachmentList{}
```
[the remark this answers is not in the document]
k-keiichi-rh: Thanks for pointing. I will drop checking VAs.

controllers/fenceagentsremediation_controller.go
```diff
@@ -158,8 +158,20 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 				"fenceAgentActionSucceeded condition", fenceAgentActionSucceededCondition, "succeeded condition", succeededCondition)
 		}
 
+		// remove out-of-service taint when using OutOfServiceTaint remediation
+		if far.Spec.RemediationStrategy == v1alpha1.OutOfServiceTaintRemediationStrategy {
+			r.Log.Info("Removing OutOfService taint", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
+			if !utils.IsResourceDeletionCompleted(r.Client, req.Name) {
```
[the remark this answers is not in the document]
k-keiichi-rh: > Do we really need this check?

I think the ResourceDeletionRemediationStrategy tries to forcefully delete all of pods explicitly. We have the way to check the terminating pods are deleted by checking the result of the deletion. So I am 100% sure we don't need this check.

However, in OutOfServiceTaintRemediationStrategy, I have 1% doubt if the terminating pods are deleted.
If NHC identifies the node becomes healthy, the control-plane or kubelet deletes the terminating pods. So we can expect there is no terminating pod in this stage and may not need this check.
However we can't control the behavior of control-plane or kubelet. And we can just expect the terminating pods are deleted indirectly by them compared to the ResourceDeletionRemediationStrategy. 

This remaining 1% was the reason why I thought we need this check. 
But I may be thinking about it too much. So I will drop this change.

> Are we sure that always all Pods get the DeletionTimestamp?
> What about Pods which tolerate the taint?

The current out-of-service taint focuses on only the terminating pods which has the DeletionTimestamp to enable workloads to failover to another node. If we can confirm if there is no terminating pod, it means we can move all workloads on the failed node to another node. If we can not, we need to improve the out-of-service taint code in k8s.

controllers/fenceagentsremediation_controller_test.go
k-keiichi-rh: Thanks. I will fix this.

controllers/fenceagentsremediation_controller_test.go
k-keiichi-rh: Same here. I will remove this change.

pkg/utils/taints.go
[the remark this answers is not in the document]
slintes: aligning with SNR is ok, I just was curious 🙂

controllers/fenceagentsremediation_controller.go
```diff
@@ -159,8 +159,17 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 				"fenceAgentActionSucceeded condition", fenceAgentActionSucceededCondition, "succeeded condition", succeededCondition)
 		}
 
+		// remove out-of-service taint when using OutOfServiceTaint remediation
+		if far.Spec.RemediationStrategy == v1alpha1.OutOfServiceTaintRemediationStrategy {
+			r.Log.Info("Removing OutOfService taint", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
```
razo7, line 164: NIT: can we use `out-of-service` instead of `OutOfService`? I saw that is referred to as `out-of-service`  in the [official Kubernetes documentation](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-kubernetes-io-out-of-service).

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
```diff
@@ -0,0 +1,72 @@
...
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *FenceAgentsRemediationTemplate) ValidateDelete() (admission.Warnings, error) {
+	// unused for now, add "delete" when needed to verbs in the kubebuilder annotation above
+	webhookTemplateLog.Info("validate delete", "name", r.Name)
+	return nil, nil
+}
+
+func validateStrategy(farSpec FenceAgentsRemediationSpec) (admission.Warnings, error) {
+	if farSpec.RemediationStrategy == OutOfServiceTaintRemediationStrategy && !validation.IsOutOfServiceTaintSupported {
+		//if farSpec.RemediationStrategy == OutOfServiceTaintRemediationStrategy {
```
razo7, line 68: NIT: redundant line

controllers/fenceagentsremediation_controller.go
k-keiichi-rh: Replaced all of them with out-of-service.

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
k-keiichi-rh: Sorry, this was a debug and test line. Removed.

test/e2e/far_e2e_test.go
```diff
@@ -130,16 +131,32 @@ var _ = Describe("FAR E2e", func() {
...
 			It("should successfully remediate the first node", func() {
-				checkRemediation(nodeName, nodeBootTimeBefore, pod)
+				checkRemediation(nodeName, nodeBootTimeBefore, pod, remediationStrategy)
 				remediationTimes = append(remediationTimes, time.Since(startTime))
 			})
 			It("should successfully remediate the second node", func() {
-				checkRemediation(nodeName, nodeBootTimeBefore, pod)
+				checkRemediation(nodeName, nodeBootTimeBefore, pod, remediationStrategy)
 				remediationTimes = append(remediationTimes, time.Since(startTime))
 			})
```
razo7, line 161: While creating the e2e tests we had in mind of three worker nodes in the cluster where we [randomly pick one](https://github.com/medik8s/fence-agents-remediation/pull/92/files#diff-b9729ac4458c7637438e3bf558999b35adedcaf4d13af1e3eb752db501fd4997R104) and filter it from the list of nodes for the next e2e test. When there are 2 e2e tests, it is okay, but with 4 there are not enough worker nodes. Therefore the fourth e2e test won't have a node to test in the current code.

Either discard the filtering of nodes or reset the list of nodes that can be remediated after the first two e2e tests or just discard the last e2e test.

k-keiichi-rh: Thank you for the clear clarification. I understand the current situation.
I will change the e2e tests to discard the filtering of nodes before running the next e2e block.

k-keiichi-rh: It became complicated when trying to skip the e2e test for OutOfServiceTaint remediation on OCP4.12.
So I chose to discard last e2e test and changed the following patches from the last commit:
https://github.com/medik8s/fence-agents-remediation/pull/92/commits/c3c496d758bbe075a16d0f0b56dfbab594e82fa6 This patch is  a new implementation of the e2e test.
https://github.com/medik8s/fence-agents-remediation/pull/92/commits/ed764b60d660e75d00cb2a258ff975d1d399d0b2 This patch implements a env variable way to skip the e2e test for OutOfServiceTaint remediation.

There are nothing changed for the rest of the patches.

slintes (changes requested):
needs rebase and 2 comments inline

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -208,16 +208,68 @@ var _ = Describe("FAR Controller", func() {
...
+			})
+			It("should have both far taint and out-of-service taint, and at the end they will be deleted", func() {
+				By("Searching for remediation taint")
+				Eventually(func(g Gomega) bool {
+					g.Expect(k8sClient.Get(context.Background(), nodeKey, node)).To(Succeed())
+					g.Expect(k8sClient.Get(context.Background(), farNamespacedName, underTestFAR)).To(Succeed())
+					res, _ := cliCommandsEquality(underTestFAR)
+					return utils.TaintExists(node.Spec.Taints, &farNoExecuteTaint) && utils.TaintExists(node.Spec.Taints, &outOfServiceTaint) && res
+				}, timeoutFinalizer, pollInterval).Should(BeTrue(), "both far taint and out-of-service taint should be added, and command format is correct")
+
```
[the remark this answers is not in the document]
k-keiichi-rh: I think it doesn't make sense for the OutOfServiceTaint remediation.
To get the effect of out-of-service taint, the Pod GC controller, part of kube-controller-manager, should be executed.
But the envtest framework doesn't run them and the test pod won't be deleted by the out-of-service taint. 
So, we can't check if the taint works well.

As for the ResourceDeletion remediation, it tries to delete the test pod manually as part of the logic of FAR. So it will make sense to test if pods were deleted with verifyPodDeleted().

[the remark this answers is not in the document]
slintes: > But the envtest framework doesn't run them

ah yes, of course, I probably mixed unit and e2e test, sorry 👍🏼

razo7:
@k-keiichi-rh can you rebase again? #119 just added a webhook, so no need to recreate it again :)

main.go
```diff
@@ -105,6 +106,10 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := validation.InitOutOfServiceTaintSupportedFlag(mgr.GetConfig()); err != nil {
+		setupLog.Error(err, "unable to verify out-of-service taint support. out-of-service taint isn't supported")
```
razo7, line 110: I think this message is a little bit misleading.
We receive an error when we are unable to verify the K8s version which can indicate whether the out-of-service taint is supported.

```suggestion
		setupLog.Error(err, "Unable to verify Kubernetes version for indicating the out-of-service taint support. out-of-service taint isn't supported")
```

test/e2e/far_e2e_test.go
```diff
@@ -78,71 +80,95 @@ var _ = Describe("FAR E2e", func() {
...
 	})
+	Context("stress cluster with ResourceDeletion remediation strategy", func() {
+		var (
+			filteredNodes                 *corev1.NodeList
+			selectedNode                  *corev1.Node
+			nodeName                      string
+			pod                           *corev1.Pod
+			startTime, nodeBootTimeBefore time.Time
+			err                           error
+			remediationStrategy           v1alpha1.RemediationStrategyType
```
razo7, line 91: I believe the following vars can be globally shared between the ResourceDeletion Context and OutOfServiceTaint Context.
Can you move them from here to https://github.com/medik8s/fence-agents-remediation/pull/92/files#diff-b9729ac4458c7637438e3bf558999b35adedcaf4d13af1e3eb752db501fd4997R54-R58 and remove them from OutOfServiceTaint Context.
Only `filteredNodes` is important to be set as nil per context.

razo7 (commented):
Looks good that all the e2e tests have passed!
/approve
Left some nits and some nice/minor improvements.
The only thing which is missing is a file like https://github.com/medik8s/self-node-remediation/blob/main/pkg/utils/taints_test.go for testing the out-of-service taint support in `setOutOfTaintFlags` function

api/v1alpha1/fenceagentsremediation_webhook.go
```diff
@@ -52,13 +52,25 @@ var _ webhook.Validator = &FenceAgentsRemediation{}
...
 func (far *FenceAgentsRemediation) ValidateCreate() (admission.Warnings, error) {
 	webhookFARLog.Info("validate create", "name", far.Name)
-	return validateAgentName(far.Spec.Agent)
+	if _, err := validateAgentName(far.Spec.Agent); err != nil {
+		return nil, err
+	}
+	if _, err := validateStrategy(far.Spec); err != nil {
+		return nil, err
+	}
+	return nil, nil
```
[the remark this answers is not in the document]
k-keiichi-rh: Fixed all of them.

api/v1alpha1/fenceagentsremediation_webhook.go
```diff
@@ -77,3 +89,10 @@ func validateAgentName(agent string) (admission.Warnings, error) {
 	}
 	return nil, nil
 }
+
+func validateStrategy(farSpec FenceAgentsRemediationSpec) (admission.Warnings, error) {
+	if farSpec.RemediationStrategy == OutOfServiceTaintRemediationStrategy && !validation.IsOutOfServiceTaintSupported {
```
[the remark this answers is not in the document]
k-keiichi-rh: Fixed it.

api/v1alpha1/fenceagentsremediation_webhook_test.go
```diff
@@ -27,6 +29,35 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
...
+
+				outOfServiceStrategy = getFAR(validAgentName, OutOfServiceTaintRemediationStrategy)
+			})
+			When("out of service taint is supported", func() {
+				BeforeEach(func() {
+					validation.IsOutOfServiceTaintSupported = true
+				})
+				It("should be allowed", func() {
+					_, err := outOfServiceStrategy.ValidateCreate()
+					Expect(err).To(Succeed())
```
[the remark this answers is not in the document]
k-keiichi-rh: Now all of check are changed to the one liner like the following:
```
-               When("agent name was not found ", func() {
-                       It("should be rejected", func() {
-                               farTemplate := getTestFARTemplate(invalidAgentName)
-                               _, err := farTemplate.ValidateCreate()
-                               Expect(err).To(HaveOccurred())
-                               Expect(err.Error()).To(ContainSubstring("unsupported fence agent: %s", invalidAgentName))
+                       When("agent name was not found ", func() {
+                               It("should be rejected", func() {
+                                       farTemplate := getTestFARTemplate(invalidAgentName)
+                                       Expect(farTemplate.ValidateCreate()).Error().To(MatchError(ContainSubstring("unsupported fence agent: %s", invalidAgentName)))
+                               })
```
The "Expect(err).To(HaveOccurred())" is removed because we can check both of checks with the one liner.

api/v1alpha1/fenceagentsremediation_webhook_test.go
```diff
@@ -27,6 +29,35 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
...
+				})
+			})
+			When("out of service taint is not supported", func() {
+				BeforeEach(func() {
+					validation.IsOutOfServiceTaintSupported = false
+				})
+				It("should be denied", func() {
+					_, err := outOfServiceStrategy.ValidateCreate()
+					Expect(err).To(MatchError(ContainSubstring(outOfServiceTaintUnsupportedMsg)))
```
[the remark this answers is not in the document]
k-keiichi-rh: Now all of checks are changed to the one liner like the following:
```
-               When("agent name was not found ", func() {
-                       It("should be rejected", func() {
-                               farTemplate := getTestFARTemplate(invalidAgentName)
-                               _, err := farTemplate.ValidateCreate()
-                               Expect(err).To(HaveOccurred())
-                               Expect(err.Error()).To(ContainSubstring("unsupported fence agent: %s", invalidAgentName))
+                       When("agent name was not found ", func() {
+                               It("should be rejected", func() {
+                                       farTemplate := getTestFARTemplate(invalidAgentName)
+                                       Expect(farTemplate.ValidateCreate()).Error().To(MatchError(ContainSubstring("unsupported fence agent: %s", invalidAgentName)))
+                               })
```
The "Expect(err).To(HaveOccurred())" is removed because we can check both of checks with the one liner.

api/v1alpha1/fenceagentsremediation_webhook_test.go
```diff
@@ -27,6 +29,35 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
 				Expect(err.Error()).To(ContainSubstring("unsupported fence agent: %s", invalidAgentName))
 			})
 		})
+
+		Context("with OutOfServiceTaint strategy", func() {
```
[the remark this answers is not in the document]
k-keiichi-rh: Added the Context to the resourceDeletion.

pkg/validation/validation.go
```diff
@@ -43,3 +63,39 @@ func NewCustomAgentValidator(agentExists AgentExists) AgentValidator {
 func (vfe *validateAgentExistence) ValidateAgentName(agent string) (bool, error) {
 	return vfe.agentExists(agent)
 }
+
+// RemoveTaint removes taint from the taint list when it is existed, and returns error if it fails in the process
+func InitOutOfServiceTaintSupportedFlag(config *rest.Config) error {
```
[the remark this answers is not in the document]
k-keiichi-rh: Sorry. Fixed it with the following:
```
+// InitOutOfServiceTaintSupportedFlag checks a cluster's k8s version and
+// set the IsOutOfServiceTaintSupported flag to true if out-of-service is supported on the cluster
```

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -47,6 +47,7 @@ const (
 	vaName1        = "va-test-1"
 	vaName2        = "va-test-2"
 
+	forced client.GracePeriodSeconds = 0
```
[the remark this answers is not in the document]
k-keiichi-rh: It's not needed anymore. Removed it.

main.go
k-keiichi-rh: Fixed it.

test/e2e/far_e2e_test.go
k-keiichi-rh: Fixed it as you suggested. Now setting filteredNodes as nil per context.

k-keiichi-rh:
> Left some nits and some nice/minor improvements. The only thing which is missing is a file like https://github.com/medik8s/self-node-remediation/blob/main/pkg/utils/taints_test.go for testing the out-of-service taint support in `setOutOfTaintFlags` function

Thank you for your comments and suggestions. I reflected all of your comments and updated my commits.
The following is all of changes from my last commit:
https://github.com/medik8s/fence-agents-remediation/commit/d7b7d32f09f1ad8fc47a9d85a4ca0861b0edbf29

test/e2e/far_e2e_test.go
razo7: I think you got me wrong here, as I wanted to move all the above 7 variables from ResourceDeletion Context and OutOfServiceTaint Context to their "parent Context", _Describe_ `FAR E2e`. These variables are first set (e.g., `pod = createTestedPod(nodeName, testContainerName)`) at the BeforeEach per strategy Context, and only later their values are used (e.g., `checkRemediation(nodeName, nodeBootTimeBefore, pod, remediationStrategy)`). Therefore they can be shared between the strategy Contexts, under Describe `FAR E2e`.

> Now setting filteredNodes as nil per context.

I didn't suggest moving `filteredNodes` variable to be var of Describe, but now the fourth e2e test is failing (there is no fourth worker). I would just revert and keep this variable under each strategy Context.

k-keiichi-rh: Thank you for correcting me. I just kept the filteredNodes variable under each strategy Context.
The changes from the last commit is https://github.com/k-keiichi-rh/fence-agents-remediation/commit/b0c26dcc28ca18a57aa725ba052468ab91069f48.

controllers/fenceagentsremediation_controller.go
```diff
@@ -219,10 +230,28 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
+				return emptyResult, err
+			} else if taintAdded {
+				r.Log.Info("out-of-service taint was added", "Node Name", req.Name)
+				commonEvents.NormalEvent(r.Recorder, node, utils.EventReasonAddOutOfServiceTaint, utils.EventMessageAddOutOfServiceTaint)
+			}
+		default:
+			//this should never happen since we enforce valid values with kubebuilder
+			err := errors.New("unsupported remediation strategy")
+			r.Log.Error(err, "Encountered unsupported remediation strategy. Please check template spec", "strategy", far.Spec.RemediationStrategy)
 			return emptyResult, err
```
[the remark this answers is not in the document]
k-keiichi-rh: Fixed. Now it's changed to "return emptyResult, nil".

pkg/validation/validation.go
```diff
@@ -4,6 +4,26 @@ import (
...
+	"k8s.io/apimachinery/pkg/version"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+var (
+	loggerValidation = ctrl.Log.WithName("validation")
+	//IsOutOfServiceTaintSupported will be set to true in case OutOfServiceTaint is supported (k8s 1.26 or higher)
+	IsOutOfServiceTaintSupported bool
```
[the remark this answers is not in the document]
k-keiichi-rh: Changed to pass OutOfServiceTaintSupported info instead of global var based on https://github.com/medik8s/node-maintenance-operator/pull/110/commits/472dae39e5f054ece3448b7c4f1989e6d95bb86e.
[This](https://github.com/medik8s/fence-agents-remediation/pull/92/commits/f54978f9be62979cc0e40bbdf34a482208ccb831) is the changes to do that.

api/v1alpha1/fenceagentsremediation_webhook.go
```diff
@@ -52,13 +54,19 @@ var _ webhook.Validator = &FenceAgentsRemediation{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (far *FenceAgentsRemediation) ValidateCreate() (admission.Warnings, error) {
 	webhookFARLog.Info("validate create", "name", far.Name)
-	return validateAgentName(far.Spec.Agent)
+	if _, err := validateAgentName(far.Spec.Agent); err != nil {
```
slintes, line 57: would be nice to run all validations and return an aggregated error, see e.g. https://github.com/medik8s/node-healthcheck-operator/blob/a295befac6e76e4dce4d92fdd2093b6ebf39239a/api/v1alpha1/nodehealthcheck_webhook.go#L93

api/v1alpha1/fenceagentsremediation_webhook.go
```diff
@@ -52,13 +54,19 @@ var _ webhook.Validator = &FenceAgentsRemediation{}
...
+		return nil, err
+	}
+	return validateStrategy(far.Spec.RemediationStrategy)
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (far *FenceAgentsRemediation) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
 	webhookFARLog.Info("validate update", "name", far.Name)
-	return validateAgentName(far.Spec.Agent)
+	if _, err := validateAgentName(far.Spec.Agent); err != nil {
```
slintes, line 66: since both Create and Update run the same validations, what about extracting both checks into a generic validation function?

api/v1alpha1/fenceagentsremediation_webhook.go
```diff
@@ -67,6 +75,10 @@ func (far *FenceAgentsRemediation) ValidateDelete() (admission.Warnings, error)
 	return nil, nil
 }
 
+func InitOutOfServiceTaintSupportedFlag(outOfServiceTaintSupported bool) {
```
slintes, line 78: is there a reason to inject the flag, and not get it from the utils package like in SNR (https://github.com/medik8s/self-node-remediation/blob/5df6e10931eabaa2e4870f2bcd540fac6bbc0336/api/v1alpha1/selfnoderemediationtemplate_webhook.go#L65)? I think SNR's way is cleaner, and being consistent between both operators make maintenance easier :)

api/v1alpha1/fenceagentsremediation_webhook_test.go
```diff
@@ -11,58 +11,118 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
...
-			It("should be accepted", func() {
-				far := getTestFAR(validAgentName)
-				_, err := far.ValidateCreate()
-				Expect(err).ToNot(HaveOccurred())
+		Context("with ResourceDeletion strategy", func() {
+
+			When("agent name match format and binary", func() {
+				It("should be accepted", func() {
+					far := getTestFAR(validAgentName)
+					Expect(far.ValidateCreate()).Error().NotTo(HaveOccurred())
```
slintes, line 19: nit: Expect(far.ValidateCreate()).To(Succeeed()) is easier

api/v1alpha1/fenceagentsremediation_webhook_test.go
```diff
@@ -11,58 +11,118 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
 
 	Context("creating FenceAgentsRemediation", func() {
 
-		When("agent name match format and binary", func() {
-			It("should be accepted", func() {
-				far := getTestFAR(validAgentName)
-				_, err := far.ValidateCreate()
-				Expect(err).ToNot(HaveOccurred())
+		Context("with ResourceDeletion strategy", func() {
```
slintes, line 14: misleading description, agent name is always validated. I think this context can be deleted, and its content just moves a level up

api/v1alpha1/fenceagentsremediation_webhook_test.go
```diff
@@ -11,58 +11,118 @@ var _ = Describe("FenceAgentsRemediation Validation", func() {
...
 			})
 		})
 	})
 
 	Context("updating FenceAgentsRemediation", func() {
-		var oldFAR *FenceAgentsRemediation
-		When("agent name match format and binary", func() {
-			BeforeEach(func() {
-				oldFAR = getTestFAR(invalidAgentName)
+		Context("with ResourceDeletion strategy", func() {
```
slintes, line 60: same as above, the tests inside do not depend on strategy

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
```diff
@@ -44,13 +44,19 @@ var _ webhook.Validator = &FenceAgentsRemediationTemplate{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (farTemplate *FenceAgentsRemediationTemplate) ValidateCreate() (admission.Warnings, error) {
 	webhookFARTemplateLog.Info("validate create", "name", farTemplate.Name)
-	return validateAgentName(farTemplate.Spec.Template.Spec.Agent)
+	if _, err := validateAgentName(farTemplate.Spec.Template.Spec.Agent); err != nil {
```
slintes, line 47: when we extract both validations into a single validate function, use it here as well

api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
```diff
@@ -11,60 +11,125 @@ var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
 
 	Context("creating FenceAgentsRemediationTemplate", func() {
 
-		When("agent name match format and binary", func() {
-			It("should be accepted", func() {
-				farTemplate := getTestFARTemplate(validAgentName)
-				_, err := farTemplate.ValidateCreate()
-				Expect(err).ToNot(HaveOccurred())
+		Context("with ResourceDeletion strategy", func() {
```
slintes, line 14: same as in FenceAgentsRemediation tests

test/e2e/far_e2e_test.go
```diff
@@ -253,17 +273,70 @@ func randomizeWorkerNode(nodes *corev1.NodeList) *corev1.Node {
 	return &nodes.Items[r.Intn(len(nodes.Items))]
 }
 
+// pickRemediatedNode returns a next remediated node from the current available nodes,
+// and then the node is removed from the list of available nodes
+func pickRemediatedNode(availableNodes *corev1.NodeList) (*corev1.Node, *corev1.NodeList) {
```
slintes, line 278: wow, that's a lot of code for just selecting a worker node... but ok. One nit though: why returning the same node list which you get in the parameter? It's a pointer, you can modify the list as needed inside the function...

slintes (changes requested):
Sorry for the late review!

Some nits, comments and questions inside...

api/v1alpha1/fenceagentsremediation_webhook.go
k-keiichi-rh: I will change both validations to a generic one.

api/v1alpha1/fenceagentsremediation_webhook.go
k-keiichi-rh: I may misunderstand your point in https://github.com/medik8s/fence-agents-remediation/pull/92#discussion_r1462871861. So please correct me if my understanding is wrong.

Now there are two ways to enable the validation of k8s.
One is the utils package way in SNR based on a global var you mentioned.
Another is the Node Maintenance Operator way mentioned in https://github.com/medik8s/node-maintenance-operator/pull/110#discussion_r1458680419.

The following patch implemented a kind of the utils package way in SNR.
https://github.com/medik8s/fence-agents-remediation/pull/92/commits/f0f74b21f67aa49ff1b1eecc5e9e63241780e371

And the following patch implemented the Node Maintenance Operator way.
https://github.com/medik8s/fence-agents-remediation/pull/92/commits/fbc4a9714d568b42c5d49b2ea535dc05d43d36fd

How do you think which way is a better way to implement the validation?
I prefer the SNR's way because it's more simple than the Node Maintenance Operator way.
But I could not avoid using a globael variable.

api/v1alpha1/fenceagentsremediation_types.go
```diff
@@ -84,6 +88,11 @@ type FenceAgentsRemediationSpec struct {
 	// NodeParameters are passed to the fencing agent according to the node that is fenced, since they are node specific
 	//+operator-sdk:csv:customresourcedefinitions:type=spec
 	NodeParameters map[ParameterName]map[NodeName]string `json:"nodeparameters,omitempty"`
+
+	//RemediationStrategy is the remediation method for unhealthy nodes.
```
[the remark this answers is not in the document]
k-keiichi-rh: Fixed it.

api/v1alpha1/fenceagentsremediation_webhook.go
```diff
@@ -35,6 +35,8 @@ var (
 	webhookFARLog = logf.Log.WithName("fenceagentsremediation-resource")
 	// verify agent existence with os.Stat function
 	agentValidator = validation.NewAgentValidator()
+	//isOutOfServiceTaintSupported will be set to true in case out-of-service taint is supported (k8s 1.26 or higher)
```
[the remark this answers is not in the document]
k-keiichi-rh: Fixed it.

api/v1alpha1/fenceagentsremediation_webhook.go
k-keiichi-rh: Not yet fixed it, but i will change it to the aggregated error in NHC.

api/v1alpha1/fenceagentsremediation_webhook.go
k-keiichi-rh: Changed to the generic one.

api/v1alpha1/fenceagentsremediation_webhook.go
k-keiichi-rh: Thank you for the clarification.
I will keep the current way to avoid using a global var. But let me know if you have any comment.

api/v1alpha1/fenceagentsremediation_webhook_test.go
k-keiichi-rh: ValidateCreate() returns "(admission.Warnings, error)". So this way doesn't work well.

api/v1alpha1/fenceagentsremediation_webhook_test.go
k-keiichi-rh: Fixed it.

api/v1alpha1/fenceagentsremediation_webhook_test.go
k-keiichi-rh: Fixed it.

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
k-keiichi-rh: Fixed it. Changed it to the generic one.

api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
k-keiichi-rh: Fixed it.

test/e2e/far_e2e_test.go
k-keiichi-rh: Simplified the current logic and change it to modify the list inside the function.

k-keiichi-rh:
@slintes Thank you for the comments. I totally agree with you and fixed the code with you comments.
The following is all of changes from my last commit: https://github.com/k-keiichi-rh/fence-agents-remediation/commit/f1aa62e497ba817c46b5797a1aecb2435ec303ea

api/v1alpha1/fenceagentsremediation_types.go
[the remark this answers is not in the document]
slintes: you need to run `make bundle` in order to get the up to date description into the manifests in `/bundle`. I'm fixing the test to catch this in this PR: https://github.com/medik8s/fence-agents-remediation/pull/140

api/v1alpha1/fenceagentsremediation_webhook_test.go
slintes: you're right, sorry, I forgot the new signature of the validation methods

controllers/fenceagentsremediation_controller.go
```diff
@@ -166,6 +166,23 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 			r.Executor.Remove(far.GetUID())
 		}
 
+		// remove out-of-service taint when using OutOfServiceTaint remediation
+		if far.Spec.RemediationStrategy == v1alpha1.OutOfServiceTaintRemediationStrategy {
+			r.Log.Info("Removing out-of-service taint", "Fence Agent", far.Spec.Agent, "Node Name", node.Name)
+			taint := utils.CreateOutOfServiceTaint()
+			if err := utils.RemoveTaint(r.Client, node.Name, taint); err != nil {
+				if apiErrors.IsNotFound(err) {
```
slintes, line 174: this should be IsConflict

pkg/utils/events.go
```diff
@@ -27,5 +29,7 @@ const (
 	EventMessageFenceAgentExecuted       = "Fence agent was executed"
 	EventMessageFenceAgentSucceeded      = "Fence agent was succeeded"
 	EventMessageDeleteResources          = "Manually delete pods from the unhealthy node"
+	EventMessageAddOutOfServiceTaint     = "out-of-service taint was added"
```
slintes, line 32: nit: please align with other messages and start with an uppercase letter (same below)

slintes (changes requested):
will review the e2e test tomorrow, 2 comments inline

controllers/fenceagentsremediation_controller.go
k-keiichi-rh: Fixed it. Thank you for pointing out.

pkg/utils/events.go
k-keiichi-rh: Fixed it.

k-keiichi-rh:
@slintes Thank you for the comments again. I reflected your comments.
The following is all of changes from my last commit: https://github.com/k-keiichi-rh/fence-agents-remediation/commits/ecoproject-1326-with-review/

slintes:
@k-keiichi-rh fyi, we have a CI outage at the moment, e2e tests are expected to fail until further notice 🙁

test/e2e/far_e2e_test.go
```diff
@@ -244,26 +266,71 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
...
+	if len(availableNodes.Items) < 1 {
+		Fail("No available node found for remediation")
+	}
 	// Generate a random seed based on the current time
 	r := rand.New(rand.NewSource(time.Now().UnixNano()))
 	// Randomly select a worker node
-	return &nodes.Items[r.Intn(len(nodes.Items))]
+	selectedNodeIndex := r.Intn(len(availableNodes.Items))
+	// Delete the selected node from the list of available nodes
+	availableNodes.Items = append(availableNodes.Items[:selectedNodeIndex], availableNodes.Items[selectedNodeIndex+1:]...)
```
slintes, line 293: you can't delete the item from the slice and then try to return in the line below

controllers/fenceagentsremediation_controller.go
```diff
@@ -228,11 +245,29 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
+			taintAdded, err := utils.AppendTaint(r.Client, node.Name, utils.CreateOutOfServiceTaint())
+			if err != nil {
+				r.Log.Error(err, "Failed to add out-of-service taint", "CR's Name", node.Name)
+				return emptyResult, err
+			} else if taintAdded {
+				r.Log.Info("out-of-service taint was added", "Node Name", node.Name)
+				commonEvents.NormalEvent(r.Recorder, node, utils.EventReasonAddOutOfServiceTaint, utils.EventMessageAddOutOfServiceTaint)
+			}
+		default:
+			//this should never happen since we enforce valid values with kubebuilder
```
slintes, line 267: what about an empty strategy when the CR already exists when the operator is updated? Would it make sense to add the empty strategy case to the ResourceDeletion case?

slintes:
there is duplicated code in the e2e test, but we clean it up in a follow up in order to get this in for the next release...

test/e2e/far_e2e_test.go
k-keiichi-rh: Is this a comment to the wrong comment("// Delete the selected node from the list of available nodes")?
Because the above codes just try to recreate a slice without the element of selectedNodeIndex.

controllers/fenceagentsremediation_controller.go
k-keiichi-rh: I misunderstand your point, but please let me confirm it.
Is there any case that  far.Spec.RemediationStrategy becomes empty?
I thought it's set to "ResourceDeletion" when RemediationStrategy is empty.
'''
        // RemediationStrategy is the remediation method for unhealthy nodes.
        // Currently, it could be either "OutOfServiceTaint" or "ResourceDeletion".
        // ResourceDeletion will iterate over all pods related to the unhealthy node and delete them.
        // OutOfServiceTaint will add the out-of-service taint which is a new well-known taint "node.kubernetes.io/out-of-service"
        // that enables automatic deletion of pv-attached pods on failed nodes, "out-of-service" taint is only supported on clusters with k8s version 1.26+ \
or OCP/OKD version 4.13+.
        // +kubebuilder:default:="ResourceDeletion"
        // +kubebuilder:validation:Enum=ResourceDeletion;OutOfServiceTaint
        RemediationStrategy RemediationStrategyType `json:"remediationStrategy,omitempty"`
'''

test/e2e/far_e2e_test.go
mshitrit: Hmm I'll try to explain how I understand Marc's comment with an example.

- Let's assume that the `availableNodes` contains only one node.
- selectedNodeIndex would be 0.
- after execution of 
```
availableNodes.Items = append(availableNodes.Items[:selectedNodeIndex], availableNodes.Items[selectedNodeIndex+1:]...)
```
availableNodes.Items would be empty
- `return &availableNodes.Items[selectedNodeIndex]` would cause a nil panic because it tries to reference a node that was removed from the slice

controllers/fenceagentsremediation_controller.go
mshitrit: I'll try to clarify with an example:

- let's say that a user has a previous version of FAR installed.
- In that version a CR is already exist, but the strategy is empty because it didn't exit yet (since  ``// +kubebuilder:validation:Enum=ResourceDeletion;OutOfServiceTaint`` will only be added in this PR)
- The user updates to the new version (but the old CR that was created in the previous version remains)
- For that particular CR the strategy would be empty even though now the operators is upgraded and uses the latest code including this PR.

IIUC Marc suggested that we'll treat such use case as if `ResourceDeletion` strategy selected - rational is that since we know this is an old CR when it was created  `ResourceDeletion`  was the only strategy.

test/e2e/far_e2e_test.go
k-keiichi-rh: @mshitrit @slintes Thank you for the explanation and correcting me.
I got the point  and fixed it with the following:
```
       // Randomly select a worker node
        selectedNodeIndex := r.Intn(len(availableNodes.Items))
+       selectedNode := availableNodes.Items[selectedNodeIndex]
        // Delete the selected node from the list of available nodes
        availableNodes.Items = append(availableNodes.Items[:selectedNodeIndex], availableNodes.Items[selectedNodeIndex+1:]...)
-       return &availableNodes.Items[selectedNodeIndex]
+       return &selectedNode
```

controllers/fenceagentsremediation_controller.go
k-keiichi-rh: @mshitrit @slintes Thank you for the explanation and correcting me again.
I fixed it with the following:
```
-               case v1alpha1.ResourceDeletionRemediationStrategy:
+               case v1alpha1.ResourceDeletionRemediationStrategy, "":
+                       // Basically RemediationStrategy should be set to ResourceDeletion strategy as the default strategy.
+                       // However it will be empty when the CS was created when ResourceDeletion strategy was the only strategy.
+                       // In this case, the empty strategy should be treated as if ResourceDeletion strategy selected.
```

mshitrit:
/lgtm
/hold
Since we are after Code Freeze, waiting for QE green light before merging

frajamomo:
/lgtm

image_360

image_720

PR #93, opened 2023-10-17
[WIP] Create far or far template CR Examples on OCP and Cloud Agents

razo7:
Introduce two bash scripts that automatically create a CR example for `fence_aws`, `fence_ipmilan`, and `fence_azure_arm` based on the platform (`AWS`, `BareMetal`, or `Azure`).
The main script, _create_cr_example_, uses the CR kind (_FenceAgentsRemediation_/_FenceAgentsRemediationTemplate_), as the first parameter and operator installation namespace as the second parameter for creating an example CR.

In case of creating an example on the **AWS** platform, then please run `make ocp-aws-credentials` to add the CredentialsRequest for OCP on AWS.

**The scripts are tested and work on OCP**

[ECOPROJECT-1665](https://issues.redhat.com//browse/ECOPROJECT-1665)

### DIFF
--- a/hack/example_creation/create_cr_example.sh
+++ b/hack/example_creation/create_cr_example.sh
@@ -1,40 +1,60 @@
 #!/bin/bash
-# The below script is meant for creating FenceAgentsRemediation/FenceAgentsRemediationTemplate CR for
-#  the following platforms: AWS, BareMetal, Azure, and GCP
-# Install fence-agents-azure-arm, and fence-agents-gce packages for Azure and GCP fence agents respectively
 # set -ex
 
-# Source the functions from get_parameters.sh
-source ./hack/example_creation/get_parameters.sh
+# The below script is meant for creating FenceAgentsRemediation/FenceAgentsRemediationTemplate CR for
+# the following platforms: AWS, Azure, BareMetal, and GCP
+# 1. AWS - https://www.mankier.com/8/fence_aws
+# 2. Azure - https://www.mankier.com/8/fence_azure_arm
+# 3. BareMetal - https://www.mankier.com/8/fence_ipmilan
+# 4. GCP - https://www.mankier.com/8/fence_gce
+# Install fence-agents-azure-arm, and fence-agents-gce packages for Azure and GCP fence agents respectively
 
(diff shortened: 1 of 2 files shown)


clobrano (commented):
This is quite complex to review, because this script does two things at once
1. build the whole CR structure
2. evaluate several CR fields according to platform and other dynamic conditions

so before going through the review, I'd like to understand if we could instead create a list of complete CR templates files (one for each platform) with bash variables as placeholders.

The script then will only need to evaluate these variables and replace them into the CRs. We might also think to have a script for each platform.

I expect this to be easier to review and maintain

razo7:
> this script does two things at once

Correct.

> if we could instead create a list of complete CR templates files (one for each platform) with bash variables as placeholders.

Doesn't seem difficult to accomplish, but it would require creating the far CRs in addition to the far templates as you mentioned. So assuming we support or show examples of 4 fence agents then we would end up with 8 yamls.

> We might also think to have a script for each platform.

I thought of using fewer scripts since there is a lot of similarity between them. But if you see a lot of value, then we can go in this direction. I can imagine having one short general script to identify the platform and receiving the CR kind for calling one out of four scripts, each per platform, for modifying a yaml for FenceAgentsRemediation/FenceAgentsRemediationTemplate.

> I expect this to be easier to review and maintain

I agree with that. Less dependency.

clobrano:
> Doesn't seem difficult to accomplish, but it would require creating the far CRs in addition to the far templates as you mentioned.

sorry, bad wording on my side. I meant "complete FAR templates" only, so only 4 yamls

Regarding the rest of your reply, basically yes, I think it would also be easier to check if the resulting template is structurally right just by having a look at the yamls

clobrano:
I only noticed now that the pr is for "creating FAR **OR** FAR template" :facepalm: 

So, my idea is
- create the structure of each yaml we want to autogenerate (so 1 FAR and 1 FAR Template for each platform, then 8 in total yes)
- create some intelligence to fill the yamls fields dinamically (one script for all, or one script for each only depends on the resulting complexity, I think)

slintes:
will close to save test resources, feel free to reopen when going on with this

/close

PR #94, opened 2023-10-17
Fix http/2 CVE

razo7:
Use golang.org/x/net v0.17.0 and updated k8s.io/apimachinery to fix http/2 CVE


slintes:
flaky test

/override ci/prow/4.12-openshift-e2e

PR #95, opened 2023-10-17
Add fence_gce Agent for GCP

razo7:
Adding a fence agent for Google Cloud Platform (GCP) platform - [fence_gce](https://www.mankier.com/8/fence_gce)


clobrano:
/lgtm
out of curiosity, how big the image would result after this addition?
/hold
feel free to unhold

razo7:
>out of curiosity, how big the image would result after this addition?

Before it was 159.7 MiB, and with the new agent it is 160.9 MiB, so the new image is only adding 1.2 MiB.

/unhold

PR #97, opened 2023-10-20
Disable HTTP/2 in metrics server

clobrano:
Disable HTTP/2 in Metric server to avoid CVE HTTP/2 Rapid Reset.

Currently, the metrics server enables/disables HTTP/2 support only if
SecureServing is enabled, which is not. However this change adds the
disabling logic anyway to avoid future issues.

see https://issues.redhat.com/browse/ECOPROJECT-1738


razo7:
flaky test
/override ci/prow/4.13-openshift-e2e

/cherry-pick release-0.2
/lgtm

PR #102, opened 2023-11-05
Remove volumeAttachment deletion

razo7:
The [kube-control-manager (KCM)](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/#:~:text=The%20Kubernetes%20controller%20manager%20is,the%20state%20of%20the%20system.) detects the pod deletion and automatically detaches the Volume Attachments (VAs) that are not connected to any pod, forcefully deleting the VA has the risk of interfering with the process of detaching the VAs.

Therefore, FAR should not delete any volume attachments as part of deleting the resources, except pods deletion. Similar to [SNR's PR](https://github.com/medik8s/self-node-remediation/pull/158).

[ECOPROJECT-1754](https://issues.redhat.com//browse/ECOPROJECT-1754)


controllers/fenceagentsremediation_controller.go
```diff
@@ -224,7 +225,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		!meta.IsStatusConditionTrue(far.Status.Conditions, commonConditions.SucceededType) {
 		// Fence agent action succeeded, and now we try to remove workloads (pods and their volume attachments)
 		r.Log.Info("Manual workload deletion", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
-		if err := utils.DeleteResources(ctx, r.Client, req.Name); err != nil {
```
clobrano, line 227: do we need to remove also the function implementation? It doesn't seem used elsewhere anymore

razo7: >do we need to remove also the function implementation?

Are you referring to `utils.DeleteResources`? I have deleted this logic, minimized it, and then moved to [Common](https://github.com/medik8s/common/pull/12) as `DeletePods`.

clobrano: Sorry, I don't know why, but at the beginning I couldn't see the changes in `pkg/utils` :+1:

PR #103, opened 2023-11-09
Add bundle-cleanup target

clobrano:
Add bundle-cleanup target to remove the bundle installed via bundle-run

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,6 +191,10 @@ export BUNDLE_RUN_NAMESPACE ?= openshift-operators
 bundle-run: operator-sdk ## Run bundle image. Default NS is "openshift-operators", redefine BUNDLE_RUN_NAMESPACE to override it.
 	$(OPERATOR_SDK) -n $(BUNDLE_RUN_NAMESPACE) run bundle $(BUNDLE_IMG)
 
+.PHONY: bundle-cleanup
+bundle-cleanup: operator-sdk ## Remove bundle installed via bundle-run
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -191,6 +191,10 @@ export BUNDLE_RUN_NAMESPACE ?= openshift-operators
 bundle-run: operator-sdk ## Run bundle image. Default NS is "openshift-operators", redefine BUNDLE_RUN_NAMESPACE to override it.
 	$(OPERATOR_SDK) -n $(BUNDLE_RUN_NAMESPACE) run bundle $(BUNDLE_IMG)
 
+.PHONY: bundle-cleanup
+bundle-cleanup: operator-sdk ## Remove bundle installed via bundle-run
+	$(OPERATOR_SDK) -n openshift-operators cleanup $(OPERATOR_NAME)
```
razo7, line 196: NIT:  would have been nicer if the namespace would be easily adjustable for cleaning up installation on any namespace.

razo7 (commented):
Thanks for adding :) left small nit that we can follow up later

Makefile
razo7: >discussion about the new installation namespace

I think this discussion is about the installation namespace for downstream FAR.

clobrano: > I think this discussion is about the installation namespace for downstream FAR.

I see, I didn't get this at first. Also, just noticed that `bundle-run` already is adjustable so I'll change `bundle-cleanup` accordingly

PR #105, opened 2023-11-13
Bundle-cleanup to use BUNDLE_RUN_NAMESPACE for namespace

clobrano:
`bundle-run` allows for custom defined Namespace via `BUNDLE_RUN_NAMESPACE`, so also `bundle-cleanup` target should use
the same variable when removing the bundle.


razo7:
Follow up to #103 
/lgtm

PR #106, opened 2023-11-21
Rework running Fence Agent command 

clobrano:
Run the fence agent command asynchronously in a dedicated goroutine on the same controller's container.
The goroutine is also responsible to update FAR status with the command outcome. For this reason two new Status Conditions have been added to take into account fence agents failures or timeouts.

The fence agent command has three new, optional, Spec values:

- **RetryCount** is the number of times the fencing agent will be executed in case of failures (default: 5)
- **RetryInterval** is the interval between each fencing agent retries (default 5s)
- **Timeout** is the timeout for each fencing agent execution (default 60s)

---

TODO
- [x] replace fence_ipmilan mock with software mock of `exec.Command` only
- [x] add fence agent failure case to `updateConditions`
- [x] use `updateConditions` in executor
- [x] split change in smaller chunks
- [x] clean up Executor runners Map when goroutine ends
- [x] investigate why e2e tests with AWS cluster time out: flaky tests :shrug:


pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,160 @@
...
-	}
-
 	return &executer{
-		log:       logger,
-		config:    config,
-		clientSet: clientSet,
+		Client:   client,
+		log:      logger,
+		ctx:      ctx,
+		routines: make(map[types.UID]bool),
```
razo7, line 46: Is `mapLock` should not be initialized here as well?

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,160 @@
...
+	e.log.Info("Updating Status Condition", "fenceAgentActionSucceededConditionStatus", status, "reason", reason, "LastUpdateTime", far.Status.LastUpdateTime)
+
+	meta.SetStatusCondition(&far.Status.Conditions, metav1.Condition{
+		Type:    v1alpha1.FenceAgentActionSucceededType,
+		Status:  status,
+		Reason:  reason,
+		Message: message,
+	})
+
+	return e.Status().Update(e.ctx, far)
```
razo7, line 159: Maybe we can update the status with [`updateConditions`](https://github.com/medik8s/fence-agents-remediation/blob/main/controllers/fenceagentsremediation_controller.go#L285-L373)  when there is no error?

clobrano: at first, I didn't want to export the function, but it would be a good plus having all the update centralized, so I'd say either use it for both outcomes (success and error), or none

pkg/cli/cliexecuter.go
clobrano: IIUC it is automatically initialized

clobrano:
Working on a different approach for testing, so you might want to wait before reviewing this PR

mshitrit:
> Working on a different approach for testing, so you might want to wait before reviewing this PR

Might want to add a [WIP] suffix on the PR until ready for review

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,168 @@
...
+	runner   runnerFunc
 }
 
-type executer struct {
-	log       logr.Logger
-	config    *restclient.Config
-	clientSet *kubernetes.Clientset
+type ShellCommand interface {
+	Run() error
 }
```
razo7, line 33: Is it used somewhere or is it redundant?

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,168 @@
...
+
+// run runs the command in the container and updates the status of the FAR instance maching the UID
+func run(ctx context.Context, logger logr.Logger, uid types.UID, command []string) (stdout, stderr string, err error) {
+	cmd := exec.CommandContext(ctx, command[0], command[1:]...)
+
+	var outBuilder, errBuilder strings.Builder
+	cmd.Stdout = &outBuilder
+	cmd.Stderr = &errBuilder
+
+	for i := 0; i < 5; i++ {
```
razo7, line 122: Why do we try to run the fence agent only 5 times?
Before this change, we didn't loop on the fence agent output. We used the reconcile back-off to retrigger reconciliation and execute the command. it might be better to continue using that and returning an error in https://github.com/medik8s/fence-agents-remediation/pull/106/files#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R198

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,168 @@
...
+}
+
+func (e *Executer) updateStatus(far *v1alpha1.FenceAgentsRemediation, err error) error {
+	var reason utils.ConditionsChangeReason
+
+	if err == nil {
+		reason = utils.FenceAgentSucceeded
 	} else {
-		e.log.Info("Command has been executed successfully", "stdout", stdoutBuf.String())
+		reason = utils.FenceAgentFailed
```
razo7, line 159: IIUC then after 5 times running the fence agent command, we considered that the cause of failures was not flakiness and we stopped running the fence agent. Right?

pkg/cli/cliexecuter.go
clobrano: Not needed indeed, thank you for catching it

pkg/cli/cliexecuter.go
clobrano: My idea was to consider the failure of the fence agent possible. I agree that the back-off retry is better than a constant delay, but with the previous implementation either the FA works or the reconcile never ends.

We cannot return error in [#106 (files)](https://github.com/medik8s/fence-agents-remediation/pull/106/files#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R198). That function only spawn a goroutine and returns immediately.
If we want to treat FA errors the way it was previously, we might let Reconcile loop handle Status Condition "FA failure" to call the fence agent again.

pkg/cli/cliexecuter.go
clobrano: Correct

pkg/cli/cliexecuter.go
razo7: >either the FA works or the reconcile never ends.

Correct. If the agent fails then IMHO the reconciliation should not proceed. At least as long it does not block the reconciliation of another far CR.

>We cannot return error in [#106 (files)](https://github.com/medik8s/fence-agents-remediation/pull/106/files#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R198). That function only spawn a goroutine and returns immediately.

Yes, I meant the line afterward where ATM we return no error and no requeue.

> If we want to treat FA errors the way it was previously, we might let Reconcile loop handle Status Condition "FA failure" to call the fence agent again.

SGTM but I prefer to limit the dependency on CR status and the reconciliation decisions as much as possible if we can understand whether the agent was succeeded or not.

pkg/cli/cliexecuter.go
razo7: I assume 5 is just a magic number that makes sense to you. Nothing more than that.
I think there is no harm in executing the command again and again until success. In the meantime, we can log/print the number of failure times or just keep presenting "processing...".

But I would just let the CLI executer run once (and not x times), if it fails return an error in reconciliation and simply try again until success. In the meantime log or present an event for running again the fence agent.
Unless we gain something from running the CLI executer in a for loop except the creation time of the command which should be in favor of looping with for (for x times) instead of looping the reconcile.

pkg/cli/cliexecuter.go
clobrano: > Yes, I meant the line afterward where ATM we return no error and no requeue.

Why do you want to return an error there? We don't need the Reconcile loop to be re-queued until we know that the FA goroutine has done something (either succeeded or failed).

> SGTM but I prefer to limit the dependency on CR status and the reconciliation decisions as much as possible if we can understand whether the agent was succeeded or not.

I have two in-progress branches, one of those uses `event.GenericEvent` to wake up the controller and check the result, but I preferred this one where the synchronization between main loop and goroutine is triggered by the status update. We could discuss further in the chat what the pros and cons are.

pkg/cli/cliexecuter.go
clobrano: > I assume 5 is just a magic number that makes sense to you. Nothing more than that.

Of course, since we cannot know what the error means, unless we have a full map telling us which error is temporary and which not, I see the retry just for be completely sure the error is of the first kind.

api/v1alpha1/fenceagentsremediation_types.go
```diff
@@ -63,6 +55,25 @@ type FenceAgentsRemediationSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec
 	Agent string `json:"agent"`
 
+	// RetryCount is the number of times the fencing agent will be executed
+	//+kubebuilder:default:=100
```
slintes, line 59: huh, that's a lot IMHO. We are talking about issues here when executing the fence agent command. I think there is low chance that it succeeds after some time once it failed, because probably there is either misconfiguration, or an issue with IPMI etc.. I would go with 5, which results in > 5 minutes already, in combination with the default 1m timeout

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,236 @@
...
+			// ExponentialBackoff to handle the fence agent command execution where:
+			// - the command fails: the command is retried until the retryCount is reached
+			// - the command times out: the command is retried until the retryCount is reached
+			// - the FA context times out: the command is cancelled and the status is updated
+			// - the FA context is cancelled: the command is cancelled and the status is not updated
+			// - the command succeeds: the command is not retried and the status is updated
+			err := wait.ExponentialBackoffWithContext(faCtx,
+				backoff,
+				func(ctx context.Context) (bool, error) {
+					ctxWithTimeout, cancel := context.WithTimeout(faCtx, timeout)
```
slintes, line 98: what about extracting this into its own func? Same for the second ExponentialBackoffWithContext call

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -56,8 +52,9 @@ const (
 )
 
 var (
-	faPodLabels = map[string]string{"app.kubernetes.io/name": "fence-agents-remediation-operator"}
-	log         = ctrl.Log.WithName("controllers-unit-test")
+	faPodLabels  = map[string]string{"app.kubernetes.io/name": "fence-agents-remediation-operator"}
+	log          = ctrl.Log.WithName("controllers-unit-test")
+	underTestFAR = &v1alpha1.FenceAgentsRemediation{}
```
slintes, line 57: I would prefer to avoid making this a package wide var...

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -102,7 +103,7 @@ var _ = Describe("FAR Controller", func() {
 					validShareString, err := buildFenceAgentParams(underTestFAR)
 					Expect(err).NotTo(HaveOccurred())
 					// Eventually buildFenceAgentParams would return the same shareParam
-					Expect(isEqualStringLists(invalidShareString, validShareString)).To(BeTrue())
+					Expect(invalidShareString).To(ConsistOf(validShareString))
```
slintes, line 106: nice :)

slintes (changes requested):
first round of review done, some nits and remarks inside though

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,236 @@
...
+	log      logr.Logger
+	routines map[types.UID]*routine
+	mapLock  sync.Mutex
+	runner   runnerFunc
 }
 
-var _ Executer = executer{}
+// runnerFunc is a function that runs the command and returns the stdout, stderr and error
+// it is configurable in Executer for testing purposes
+type runnerFunc func(ctx context.Context, uid types.UID, command []string, logger logr.Logger) (string, string, error)
```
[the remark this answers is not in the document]
clobrano: correct, I'll remove them

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,236 @@
...
 
-type executer struct {
-	log       logr.Logger
-	config    *restclient.Config
-	clientSet *kubernetes.Clientset
+type Executer struct {
+	client.Client
+	log      logr.Logger
+	routines map[types.UID]*routine
+	mapLock  sync.Mutex
```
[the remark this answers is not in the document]
clobrano: yes, I agree

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,236 @@
...
-func (e executer) Execute(pod *corev1.Pod, command []string) (stdout string, stderr string, err error) {
-	if len(pod.Spec.Containers) == 0 {
-		err := errors.New("create cli executer failed")
-		e.log.Error(err, "No container found in Pod", "Pod Name", pod.Name)
-		return "", "", err
+// AsyncExecute runs the command in a goroutine mapped to the UID
+func (e *Executer) AsyncExecute(ctx context.Context, uid types.UID, command []string, retryCount int, retryDuration, timeout time.Duration) {
+	e.mapLock.Lock()
+	defer e.mapLock.Unlock()
+	if _, exist := e.routines[uid]; !exist {
```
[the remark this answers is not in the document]
clobrano: I agree, I'll revert the check

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,236 @@
...
+		// create a context for the fence agent command that the controller can cancel
+		cancellableCtx, cancel := context.WithCancel(ctx)
+		routine := routine{
+			cancel: cancel,
+		}
+		e.routines[uid] = &routine
+
+		go func(faCtx, crCtx context.Context, uid types.UID, command []string) {
```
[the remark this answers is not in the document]
clobrano: > also: why do we need both context here?

Initially I wanted to be able to cancel the fence agent runner context (first ExponentialBackoff) without cancelling the context of the status update (second ExponentialBackoff).

However, I didn't take into account that in this final version if the `faCtx` is cancelled, the goroutine does not update the status anyway, so the double context should be unnecessary.

api/v1alpha1/fenceagentsremediation_types.go
clobrano: I need to test this one better. Initially my understanding was that aws agent fails a lot (like a lot), but at the end it was the cluster misconfigured and forgot to review the number of retries.

I agree that 5 is a good number

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,236 @@
...
+							e.log.Info("could not update status", "FAR uid", uid, "reason", err)
+							return false, err
+						}
+
+						e.log.Error(err, "could not update status", "FAR uid", uid)
+						return false, err
+					}
+
+					if err := e.updateStatus(crCtx, far, faErr); err != nil {
+						if apiErrors.IsConflict(err) &&
```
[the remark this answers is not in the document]
clobrano: yes, this way doesn't make sense :sweat_smile:

controllers/fenceagentsremediation_controller_test.go
clobrano: fixed [here](https://github.com/medik8s/fence-agents-remediation/compare/fb7e571984f5c43af2d91c20602fbf20a1ecb934..b63a143313dbfa4e3b6b3c826e54bb09910f4f6f#diff-fac0cf56dda42683c1e24b46c87cfa4cca46c444145ee69de9e264db50f10289R52)

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -218,6 +357,10 @@ func getFenceAgentsRemediation(nodeName, agent string, sharedparameters map[v1al
 			Agent:            agent,
 			SharedParameters: sharedparameters,
 			NodeParameters:   nodeparameters,
+			// Set the retry count to the minimum for the majority of the tests
```
razo7, line 360: Similar to e2e tests- please set these vars as Const in the beginning

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -273,47 +416,28 @@ func createVA(vaName, unhealthyNodeName string) *storagev1.VolumeAttachment {
...
-	sort.Strings(s1)
-	sort.Strings(s2)
-	return reflect.DeepEqual(s1, s2)
-}
 
-// cliCommandsEquality creates the command for CLI and compares it with the production command
-func cliCommandsEquality(far *v1alpha1.FenceAgentsRemediation) (bool, error) {
-	if mocksExecuter.command == nil {
-		return false, errors.New("The command from mocksExecuter is null")
+		// Delete the resource immediately
```
razo7, line 433: I don't think the next lines are needed.
`Expect(k8sClient.Delete(context.Background(), podTest, forced)).To(Succeed())` is good enough IMHO.

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -324,10 +448,10 @@ func verifyPodDeleted(podName string) {
 		err := k8sClient.Get(context.Background(), podKey, pod)
 		return apierrors.IsNotFound(err)
 	}, timeoutDeletion, pollInterval).Should(BeTrue())
-	log.Info("Pod is no longer exist", "pod", podName)
+	log.Info("Pod does not longer exist", "pod", podName)
```
razo7, line 451: Or maybe?
```suggestion
	log.Info("Pod no longer exists", "pod", podName)
```

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -123,44 +123,45 @@ var _ = Describe("FAR Controller", func() {
...
 		})
+
 		JustBeforeEach(func() {
 			// Create node, and FAR CR, and at the end clean them up with DeferCleanup
 			Expect(k8sClient.Create(context.Background(), node)).To(Succeed())
 			DeferCleanup(k8sClient.Delete, context.Background(), node)
+
 			Expect(k8sClient.Create(context.Background(), underTestFAR)).To(Succeed())
-			DeferCleanup(k8sClient.Delete, context.Background(), underTestFAR)
+			DeferCleanup(cleanupFar(), context.Background(), underTestFAR)
```
razo7, line 148: Is this function really needed?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -200,11 +201,149 @@ var _ = Describe("FAR Controller", func() {
...
+			BeforeEach(func() {
+				plogs.Clear()
+				node = utils.GetNode("", workerNode)
+
+				underTestFAR = getFenceAgentsRemediation(workerNode, fenceAgentIPMI, testShareParam, testNodeParam)
+			})
+
+			When("CR is deleted in between fence agent retries", func() {
+				BeforeEach(func() {
+					// Set the controlledRun to fail to simulate a case where CR deletion occurs in between fence agent
```
razo7, line 222:
```suggestion
					// Fail controlledRun to simulate a case where CR deletion occurs in between fence agent
```

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -200,11 +201,149 @@ var _ = Describe("FAR Controller", func() {
...
-				verifyStatusCondition(dummyNode, commonConditions.ProcessingType, conditionStatusPointer(metav1.ConditionFalse))
-				verifyStatusCondition(dummyNode, v1alpha1.FenceAgentActionSucceededType, conditionStatusPointer(metav1.ConditionFalse))
-				verifyStatusCondition(dummyNode, commonConditions.SucceededType, conditionStatusPointer(metav1.ConditionFalse))
+				verifyStatusCondition(underTestFAR.Status.Conditions, dummyNode, commonConditions.ProcessingType, conditionStatusPointer(metav1.ConditionFalse))
+				verifyStatusCondition(underTestFAR.Status.Conditions, dummyNode, utils.FenceAgentActionSucceededType, conditionStatusPointer(metav1.ConditionFalse))
+				verifyStatusCondition(underTestFAR.Status.Conditions, dummyNode, commonConditions.SucceededType, conditionStatusPointer(metav1.ConditionFalse))
+			})
+		})
+
+		Context("Fence Agent Failures", func() {
```
razo7, line 212: Nit: no need for capital letters.
```suggestion
		Context("Fence agent failures", func() {
```

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -200,11 +201,149 @@ var _ = Describe("FAR Controller", func() {
...
+					Expect(k8sClient.Delete(context.Background(), underTestFAR)).To(Succeed())
+
+					By("Verifying goroutine stopped without trying to update the conditions")
+					Eventually(func() bool {
+						return plogs.Contains("fence agent context canceled. Nothing to do")
+					}).Should(BeTrue())
+				})
+			})
+
+			When("CR is deleted in during fence agent execution", func() {
```
razo7, line 253:
```suggestion
			When("CR is deleted during fence agent execution", func() {
```

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -200,11 +201,149 @@ var _ = Describe("FAR Controller", func() {
...
+
+					By("Deleting the CR")
+					Expect(k8sClient.Get(context.Background(), client.ObjectKey{Name: workerNode, Namespace: defaultNamespace}, underTestFAR)).To(Succeed())
+					Expect(k8sClient.Delete(context.Background(), underTestFAR)).To(Succeed())
+
+					By("Verifying goroutine stopped without trying to update the conditions")
+					Eventually(func() bool {
+						return plogs.Contains("fence agent context canceled. Nothing to do")
+					}).Should(BeTrue())
+				})
```
razo7, line 250: This unit test and the next one are quite similar. Can they be under one function which will be called twice for each `when` and `BeforeEach`?

razo7 (commented):
left some comments, but overall looks very nice. The reconcile loop got simplified :)

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,263 @@
...
+	// - the command succeeds: the command is not retried and the status is updated
+
+	// Linear backoff
+	backoff := wait.Backoff{
+		Steps:    retryCount,
+		Duration: retryInterval,
+		Factor:   1.0,
+	}
+
+	e.log.Info("fence agent start", "uid", uid, "command", command, "retryCount", retryCount, "retryInterval", retryInterval, "timeout", timeout)
```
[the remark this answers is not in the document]
razo7: Don't expose the command in the log, it could include credentials for the API that the fence agent is using. It is not something we want to expose. We could print the fence agent name and the node/cr name for unique and easier-to-identify prints

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,263 @@
...
+			if wait.Interrupted(faErr) {
+				e.log.Error(faErr, "fence agent timeout", "uid", uid)
+				return false, faErr
+			}
+
+			e.log.Info("command failed", "uid", uid, "response", stdout, "errMessage", stderr, "err", faErr)
+			return false, nil
+		})
+
+	e.log.Info("fence agent done", "uid", uid, "command", command, "stdout", stdout, "stderr", stderr, "err", faErr)
```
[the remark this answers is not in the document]
razo7: See my above comment on printing the command.

test/e2e/far_e2e_test.go
```diff
@@ -283,7 +284,9 @@ func deleteFAR(far *v1alpha1.FenceAgentsRemediation) {
 func cleanupTestedResources(pod *corev1.Pod) {
 	newPod := &corev1.Pod{}
 	if err := k8sClient.Get(context.Background(), client.ObjectKeyFromObject(pod), newPod); err == nil {
-		Expect(k8sClient.Delete(context.Background(), newPod, forced)).To(Succeed())
+		// set GracePeriodSeconds to 0 to delete the pod immediately
+		var force client.GracePeriodSeconds = 0
```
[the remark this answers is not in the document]
clobrano: I also removed the globally defined `force` from #110. Since it is only used in this place I think it is better keep it locally.

test/e2e/far_e2e_test.go
```diff
@@ -262,6 +260,9 @@ func createFAR(nodeName string, agent string, sharedParameters map[v1alpha1.Para
 			Agent:            agent,
 			SharedParameters: sharedParameters,
 			NodeParameters:   nodeParameters,
+			RetryCount:       5,
+			RetryInterval:    metav1.Duration{Duration: 20 * time.Second},
+			Timeout:          metav1.Duration{Duration: 60 * time.Second},
```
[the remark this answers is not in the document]
clobrano: I am unsure of the benefits. This values are only used here and it seems to me that they're meaning is already clear without a specific name.

controllers/fenceagentsremediation_controller_test.go
clobrano: In my tests the `JustBeforeEach` step always finds the previously created CR still available in the cluster and fails. Like the following

```
[JustBeforeEach] /home/clobrano/workspace/medik8s/upstream/fence-agents-remediation/rework-fa-async1/controllers/fenceagentsremediation_controller_test.go:142                                          
    190   [It] /home/clobrano/workspace/medik8s/upstream/fence-agents-remediation/rework-fa-async1/controllers/fenceagentsremediation_controller_test.go:231                                                      
      1                                                                                                                                                                                                           
      2   [FAILED] Expected success, but got an error:                                                                                                                                                            
      3       <*errors.StatusError | 0xc0004c4280>:                                                                                                                                                               
      4       object is being deleted: fenceagentsremediations.fence-agents-remediation.medik8s.io "worker-0" already exists
```
This is why `cleanupFar()` waits for the CR deletion.
I noticed though that the comment is wrong, it doesn't remove FAR finalizer. I'll update that

controllers/fenceagentsremediation_controller_test.go
clobrano: The first one needs to _wait for some retries_, while the latter will stuck in its first one. Unfortunately this part is right in the middle two exactly equal pieces of code, but it's decisive.

I noticed a wrong comment in the latter though :smile:

controllers/fenceagentsremediation_controller_test.go
razo7: So how about just delete far with force (zero grace period)?

controllers/fenceagentsremediation_controller_test.go
razo7: >Unfortunately this part is right in the middle two exactly equal pieces of code

Yes, I was thinking of having one function with similar code and only the changing part (the one in the middle) will be decided during runtime based on an input (bool/int/string). It might ease the reading of the unit test logic and easier to detect errors that might impact both of them (since they share similar code).

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,263 @@
...
+			}
+
+			e.log.Info("status updated", "FAR uid", uid)
+			return true, nil
+		})
+	return err
+}
+
+// Exists checks if there is already a running Fence Agent command mapped to the UID
+func (e *Executer) Exists(uid types.UID) bool {
```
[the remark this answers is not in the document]
clobrano: > let Remove function use the Exists function

`Remove` also deletes the `UID` from the map, which needs the lock anyway

> Maybe use `IsExist` as function name

I'd rather keep `Exists`, the meaning is clear as well as the expected return value.

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,263 @@
...
+		delete(e.routines, uid)
+	}
+}
+
+// run runs the command in the container and updates the status of the FAR instance maching the UID
+func run(ctx context.Context, command []string) (stdout, stderr string, err error) {
+	cmd := exec.CommandContext(ctx, command[0], command[1:]...)
+
+	var outBuilder, errBuilder strings.Builder
+	cmd.Stdout = &outBuilder
```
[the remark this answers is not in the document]
clobrano: `cmd.Stdout` and `cmd.Stderr` are `nil` unless defined otherwise.
In this case I want to build a `string` out of them, so I used `string.Builder`.

It's similar to what [CombinedOutput](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/os/exec/exec.go;l=995) does, but with `bytes.Buffer`

controllers/fenceagentsremediation_controller.go
```diff
@@ -178,60 +180,43 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
 		if err := commonResources.DeletePods(ctx, r.Client, req.Name); err != nil {
 			r.Log.Error(err, "Manual workload deletion has failed", "CR's Name", req.Name)
 			return emptyResult, err
 		}
-		if err := updateConditions(v1alpha1.RemediationFinishedSuccessfully, far, r.Log); err != nil {
+		if err := utils.UpdateConditions(utils.RemediationFinishedSuccessfully, far, r.Log); err != nil {
 			return emptyResult, err
 		}
+
+		r.Executor.Remove(far.GetUID())
```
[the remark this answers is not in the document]
clobrano: A first version had the `Remove` called at the beginning of _this section_ (at least, I think you meant that by "the beginning of the function"), but it frequently created conflicts (see "Delay cleanup executor" commit).
Basically the goroutine was still _kind of alive_ when the `Remove` was called even if it had done everything, which resulted in the routine's context to be canceled and error messages to appear with no real reason.

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,263 @@
...
-		e.log.Error(err, "Failed to run exec command", "stdout", stdoutBuf.String(), "stderr", stderrBuf.String())
+	go e.fenceAgentRoutine(cancellableCtx, uid, command, retryCount, retryInterval, timeout)
+}
+
+func (e *Executer) fenceAgentRoutine(ctx context.Context, uid types.UID, command []string, retryCount int, retryInterval, timeout time.Duration) {
+	// run the command and update the status
+	retryErr, cmdErr := e.runWithRetry(ctx, uid, command, retryCount, retryInterval, timeout)
+	if retryErr != nil {
+		switch {
+		case errors.Is(retryErr, context.Canceled):
```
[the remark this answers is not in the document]
clobrano: I did the same in recent unit tests, as I agree that for the test developer is easier, but I recently changed my mind.
On one side reading the test is easier if the actual text is printed, and moreover it catches typos when writing/changing the message.

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,263 @@
...
+	// - the command times out: the command is retried until the retryCount is reached
+	// - the FA context times out: the command is cancelled and the status is updated
+	// - the FA context is cancelled: the command is cancelled and the status is not updated
+	// - the command succeeds: the command is not retried and the status is updated
+
+	// Linear backoff
+	backoff := wait.Backoff{
+		Steps:    retryCount,
+		Duration: retryInterval,
+		Factor:   1.0,
```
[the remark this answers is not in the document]
clobrano: yes, that's why I commented the section with "Linear backoff" (indeed, I could have just named the variable `linearBackoff` :thinking:).
My idea is that we don't know why the command failed, so give more time in between retries it is not necessarily good.

pkg/cli/cliexecuter.go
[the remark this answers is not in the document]
razo7: >it catches typos when writing/changing the message.

But now if you change something in the log output, then you need to remember to apply it in many places. Using a const used multiple times will mitigate that, and there will be only one place to catch this kind of typos.

controllers/fenceagentsremediation_controller_test.go
clobrano: > the changing part (the one in the middle) will be decided during runtime based on an input (bool/int/string)

IMHO this makes the test harder to read. The reader should also figure out the logic of the input and if the changing part has to be done or not

razo7: ok, I though it would mainly simlify stuff.

controllers/controllers_suite_test.go
```diff
@@ -35,31 +38,79 @@ import (
...
 
 var _ = BeforeSuite(func() {
+	plogs = &peekLogger{logs: make([]string, 0)}
 	opts := zap.Options{
+		DestWriter:  plogs,
 		Development: true,
 		TimeEncoder: zapcore.RFC3339NanoTimeEncoder,
 	}
-	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseFlagOptions(&opts)))
+	logf.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
```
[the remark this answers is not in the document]
clobrano: I needed to pass the logs to the `peekLogger`

controllers/fenceagentsremediation_controller_test.go
clobrano: I am unsure of the benefits. These values are only used here and it seems to me that their meaning is already clear without a specific name.

controllers/fenceagentsremediation_controller_test.go
clobrano: Correct, I rewrote this. However, I prefer to keep the `force` variable locally

controllers/fenceagentsremediation_controller_test.go
clobrano: I forgot that `cleanupFar` also ignores when FAR CR does not exist already, which happens in one of the new tests.

> So how about just delete far with force (zero grace period)?

I just tried removing the final wait for the CR to be deleted and the error above appens again.

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
razo7: >Basically the goroutine was still kind of alive when the Remove was called even if it had done everything, which resulted in the routine's context to be canceled and error messages to appear with no real reason.

Logically, it makes sense to clean up Executor routine before beginning to remove workloads. The remove just happened to fast if we do it in the beginning of thes section (line 204). Can we maybe "catch" this error and recognize it as expected behavior?

controllers/fenceagentsremediation_controller_test.go
razo7: I wonder why it was not needed before or we haven't seen it earlier.

pkg/cli/cliexecuter.go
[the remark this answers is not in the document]
razo7: >let Remove function use the Exists function.


I thouth of the following change
```
func (e *Executer) Remove(uid types.UID) {
	if Exists(uid) {
        e.routinesLock.Lock()
	    defer e.routinesLock.Unlock()
        e.log.Info("cancelling fence agent routine", "uid", uid)
		routine.cancel()
    }
}
```

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -200,11 +201,149 @@ var _ = Describe("FAR Controller", func() {
...
+			When("Fence Agent command fails", func() {
+				BeforeEach(func() {
+					mockError = errors.New("mock error")
+					DeferCleanup(func() { mockError = nil })
+
+					underTestFAR.Spec.RetryCount = 3
+					underTestFAR.Spec.RetryInterval = metav1.Duration{Duration: 1 * time.Millisecond}
+				})
+
+				It("should retry the fence agent command as configured and update the status accordingly", func() {
```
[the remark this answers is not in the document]
clobrano: I created two new helper functions to collect similar code `verifyRemediationTaintExists` and `verifyRemediationConditions`

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,263 @@
...
+			e.log.Info("fence agent context canceled. Nothing to do")
+			return
+		case wait.Interrupted(retryErr):
+			e.log.Info("fence agent context timed out")
+		default:
+			e.log.Error(retryErr, "fence agent retry error")
+		}
+	}
+
+	retryErr = e.updateStatusWithRetry(ctx, uid, cmdErr)
```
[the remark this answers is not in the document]
clobrano: It can also be the retry error, but since it's just an error and it is not used elsewhere, I can replace it with 

```
if err := e.updateStatusWithRetry(...); err != nil {}
```

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,263 @@
...
+		retry.DefaultBackoff,
+		func(ctx context.Context) (bool, error) {
+			far, err := e.getFenceAgentsRemediationByUID(ctx, uid)
+			if err != nil {
+				if apiErrors.IsNotFound(err) {
+					e.log.Info("could not find FAR by UID", "FAR uid", uid)
+					return false, err
+				}
+
+				if wait.Interrupted(err) {
```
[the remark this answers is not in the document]
clobrano: The next line is different. It is used to catch and unexpected error and in fact the log is of type `Error`.
The message in this if clause instead is not an error (type `Info`), it is something that occurs normally if the context is canceled.

pkg/cli/cliexecuter.go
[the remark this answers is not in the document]
clobrano: I'll remove the command from the log.

Regarding the suggested change, I prefer to keep it as is. It is a mark in the log to show when the command starts and also it didn't started _yet_, so "is started" is not correct.

pkg/cli/cliexecuter.go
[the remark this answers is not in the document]
clobrano: I understand, but this is not equal. In the current implementation we check for routine existence AND we cancel it under the same routine Lock, so nothing can happen in the middle.

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
clobrano: It is not possible. The error messages I was talking about come from the goroutine and it cannot differentiate a legit context cancelling from this one.

controllers/fenceagentsremediation_controller_test.go
razo7: ok

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,268 @@
...
+		runner:   run,
+	}, nil
+}
 
-	return &executer{
-		log:       logger,
-		config:    config,
-		clientSet: clientSet,
+// NewExecuter builds an Executer with configurable runnerFunc for testing
+func NewFakeExecuter(client client.Client, fn runnerFunc) (*Executer, error) {
```
mshitrit, line 60: I think it's a good rule of thumb to separate test code from production code,  based on that I'd expect this method to be  in a different file (and can probably made private as well).

clobrano: good point, I'll move it to the test code

pkg/cli/cliexecuter.go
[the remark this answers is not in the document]
razo7: Oho, now I get it. You check if the wait was interupted multiple times.

I still don't understand why the log prints here are `could not update status`.
Only after `if err := e.updateStatus(ctx, far, fenceAgentErr)` we can print whether the status update failed or succeeded. Here I would expect print/log similar to `could not find FAR by UID`.

>The next lines are similar to the lines when the code doesn't enter the if clause.

I was comparing lines 173-179 with 183-196

pkg/cli/cliexecuter.go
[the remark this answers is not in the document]
razo7: It was just close to be equal

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
razo7: ok, thanks for the info and the compromise.

pkg/cli/cliexecuter.go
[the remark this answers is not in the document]
razo7: So it is not `"Linear backoff"` :) SGTM

pkg/cli/cliexecuter.go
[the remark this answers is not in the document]
clobrano: > I still don't understand why the log prints here are `could not update status`.

Consider that this message is the last function's log before the exit. I wanted a message referred to the purpose of the function `updateStatusWithRetry` and why it stopped.

From my point of view, in this context reading:

    "could not find FAR by UID: context cancelled"

has less information than

    "could not update status: context cancelled"

even if technically (for anyone knowing the code and not just a user) the interrupted step was finding FAR.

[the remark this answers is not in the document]
clobrano: > I was comparing lines 173-179 with 183-196

probably the code has shifted after latest commits. Can you please point me to the right lines again?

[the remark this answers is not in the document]
razo7: >Consider that this message is the last function's log before the exit. I wanted a message referred to the purpose of the function updateStatusWithRetry and why it stopped.

Gotcha, but I think the prints can be a little more meaningful, and printing the same message (even though one was info and the other was error) is confusing. But we can do it as a follow-up, it is mainly my preference.

1. Same message 1 - https://github.com/medik8s/fence-agents-remediation/pull/106/files#diff-f2169f37331a57117047ad17f0b9bac545f2606f32eadce37a69fedb2b92cc61R162 and https://github.com/medik8s/fence-agents-remediation/pull/106/files#diff-f2169f37331a57117047ad17f0b9bac545f2606f32eadce37a69fedb2b92cc61R166
2. same message 2 - https://github.com/medik8s/fence-agents-remediation/pull/106/files#diff-f2169f37331a57117047ad17f0b9bac545f2606f32eadce37a69fedb2b92cc61R172 and https://github.com/medik8s/fence-agents-remediation/pull/106/files#diff-f2169f37331a57117047ad17f0b9bac545f2606f32eadce37a69fedb2b92cc61R176

pkg/cli/cliexecuter.go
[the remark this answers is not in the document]
clobrano: Oh right, it's constant indeed

pkg/cli/cliexecuter.go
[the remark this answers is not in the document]
clobrano: I rephrased the messages

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,263 @@
...
+				}
+				if apiErrors.IsConflict(err) {
+					e.log.Error(err, "conflict while updating the status", "FAR uid", uid)
+					return true, nil
+				}
+				e.log.Error(err, "failed to update status", "FAR uid", uid)
+				return false, err
+			}
+
+			e.log.Info("status updated", "FAR uid", uid)
```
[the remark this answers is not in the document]
clobrano: I don't understand, it is not an error

clobrano:
/hold
waiting for @mshitrit feedback on his change request

pkg/cli/fake.go
```diff
@@ -0,0 +1,19 @@
+package cli
```
slintes, line 1: can this be fake_test.go?

pkg/cli/cliexecuter.go
slintes: The updated version is different from what I meant, but it's better than the old version a well, so I'm fine with it 🙂 
( I meant to leave wait.ExponentialBackoffWithContext() here, and just extract the function argument into a separate function)

pkg/cli/cliexecuter.go
```diff
@@ -1,97 +1,255 @@
...
+		Factor:   1.0,
+	}
+
+	e.log.Info("fence agent start", "uid", uid, "fence_agent", command[0], "retryCount", retryCount, "retryInterval", retryInterval, "timeout", timeout)
+
+	var stdout, stderr string
+	retryErr = wait.ExponentialBackoffWithContext(ctx,
+		backoff,
+		func(ctx context.Context) (bool, error) {
+			ctxWithTimeout, cancel := context.WithTimeout(ctx, timeout)
```
slintes, line 122: do we need this? :thinking: IIUC, the context we get here is the same which we pass to ExponentialBackoffWithContext. Why would we need to cancel that one when we leave this function. Even more, isn't that  an issue when do that?

slintes: but maybe it's just too late for me to understand it completely

slintes (commented):
overall lgtm... few comments inline. Not sure about context handling at one place :thinking:

pkg/cli/cliexecuter.go
clobrano: > Why would we need to cancel that one when we leave this function

Not when we leave the function, but when we are in the middle of the function (either in the retry or during a command call), and we want to stop it (e.g. NHC time out)

pkg/cli/cliexecuter.go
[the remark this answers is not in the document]
razo7: It is a more grammar issue and personal preference than an error.

pkg/cli/fake.go
clobrano: I think the `_test.go` makes this file a test and not usable as source for another test.
At least, only changing the name breaks the unit test

clobrano: In case I'm missing something, I'll fix this in a following PR

mshitrit:
I'm unholding this PR since only contain Nits and it has an E2E fix which is relevant for other PRs. 
As @clobrano mentioned in case needed it can be addressed in followup PR.
/unhold

pkg/cli/cliexecuter.go
slintes: > Not when we leave the function

but that's we do with the defer one line below, not? 🤔

PR #107, opened 2023-11-22
Add Events for far Remediation Process

razo7:
To ease the tracking of the remediation process of far [custom resource](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) we add events to indicate major milestones (such as fencing, workload deletion, etc.)

1. Adding events (and unit testing)
2. Check whether NoExecute taint was added to signal an event. 
3. Only use `GetNodeWithName` instead of `IsNodeNameValid` which is no longer needed.
4. Small refactoring to unit tests

[ECOPROJECT-1663](https://issues.redhat.com//browse/ECOPROJECT-1663)

### DIFF
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -121,12 +121,12 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 
 	// Validate FAR CR name to match a nodeName from the cluster
 	r.Log.Info("Check FAR CR's name")
-	node, valid, err := utils.IsNodeNameValid(r.Client, req.Name)
+	node, err := utils.GetNodeWithName(r.Client, req.Name)
 	if err != nil {
 		r.Log.Error(err, "Unexpected error when validating CR's name with nodes' names", "CR's Name", req.Name)
 		return emptyResult, err
 	}
-	if !valid {
+	if node == nil {
 		r.Log.Error(err, "Didn't find a node matching the CR's name", "CR's Name", req.Name)
 		utils.UpdateConditions(utils.RemediationFinishedNodeNotFound, far, r.Log)
 		commonEvents.WarningEvent(r.Recorder, far, utils.EventReasonCrNodeNotFound, utils.EventMessageCrNodeNotFound)
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -25,6 +25,7 @@ import (
 	"github.com/go-logr/logr"
 	commonAnnotations "github.com/medik8s/common/pkg/annotations"
 	commonConditions "github.com/medik8s/common/pkg/conditions"
+	commonEvents "github.com/medik8s/common/pkg/events"
 	commonResources "github.com/medik8s/common/pkg/resources"
 
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
@@ -33,6 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	utilErrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -57,6 +59,7 @@ type FenceAgentsRemediationReconciler struct {
 	client.Client
 	Log      logr.Logger
 	Scheme   *runtime.Scheme
+	Recorder record.EventRecorder
 	Executor *cli.Executer
 }
 
@@ -104,6 +107,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "Failed to get FenceAgentsRemediation CR")
 		return emptyResult, err
 	}
+	commonEvents.RemediationStarted(r.Recorder, far)
 
 	// At the end of each Reconcile we try to update CR's status
 	defer func() {
@@ -125,14 +129,16 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 	if !valid {
 		r.Log.Error(err, "Didn't find a node matching the CR's name", "CR's Name", req.Name)
 		utils.UpdateConditions(utils.RemediationFinishedNodeNotFound, far, r.Log)
+		commonEvents.WarningEvent(r.Recorder, far, utils.EventReasonCrNodeNotFound, utils.EventMessageCrNodeNotFound)
 		return emptyResult, err
 	}
 
 	// Check NHC timeout annotation
 	if isTimedOutByNHC(far) {
-		r.Log.Info("FAR remediation was stopped by Node Healthcheck Operator")
+		r.Log.Info(utils.EventMessageRemediationStoppedByNHC)
 		r.Executor.Remove(far.GetUID())
 		utils.UpdateConditions(utils.RemediationInterruptedByNHC, far, r.Log)
+		commonEvents.RemediationStoppedByNHC(r.Recorder, far)
 		return emptyResult, err
 	}
 
@@ -145,6 +151,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Info("Finalizer was added", "CR Name", req.Name)
 
 		utils.UpdateConditions(utils.RemediationStarted, far, r.Log)
+		commonEvents.NormalEvent(r.Recorder, far, utils.EventReasonAddFinalizer, utils.EventMessageAddFinalizer)
 		return requeueImmediately, nil
 	} else if controllerutil.ContainsFinalizer(far, v1alpha1.FARFinalizer) && !far.ObjectMeta.DeletionTimestamp.IsZero() {
 		// Delete CR only when a finalizer and DeletionTimestamp are set
@@ -169,6 +176,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 			return emptyResult, fmt.Errorf("failed to remove finalizer from CR - %w", err)
 		}
 		r.Log.Info("Finalizer was removed", "CR Name", req.Name)
+		commonEvents.NormalEvent(r.Recorder, far, utils.EventReasonRemoveFinalizer, utils.EventMessageRemoveFinalizer)
 		return emptyResult, nil
 	}
 	// Add FAR (medik8s) remediation taint
@@ -196,6 +204,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		cmd := append([]string{far.Spec.Agent}, faParams...)
 		r.Log.Info("Execute the fence agent", "Fence Agent", far.Spec.Agent, "Node Name", req.Name, "FAR uid", far.GetUID())
 		r.Executor.AsyncExecute(ctx, far.GetUID(), cmd, far.Spec.RetryCount, far.Spec.RetryInterval.Duration, far.Spec.Timeout.Duration)
+		commonEvents.NormalEvent(r.Recorder, far, utils.EventReasonFenceAgentExecuted, utils.EventMessageFenceAgentExecuted)
 		return emptyResult, nil
 	}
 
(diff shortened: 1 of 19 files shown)


controllers/fenceagentsremediation_controller.go
```diff
@@ -202,6 +213,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		cmd := append([]string{far.Spec.Agent}, faParams...)
 		// The Fence Agent is excutable and the parameters structure are valid, but we don't check their values
 		r.Log.Info("Execute the fence agent", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
+		eventGenerator(r.Recorder, far, v1alpha1.EventReasonFenceAgentExecuted)
```
mshitrit, line 216: I think the whole eventGenerator -> eventCreator is an over complication.
IMO they can be removed since they can be easily replaced by a clear one liner:
For example in this case:
instead of `eventGenerator(r.Recorder, far, v1alpha1.EventReasonFenceAgentExecuted)`
Just use: `r.Recorder.Event(far, v1alpha1.EventTypeNormal, v1alpha1.EventReasonFenceAgentExecuted, v1alpha1.EventMessageFenceAgentExecuted)`

razo7: >IMO they can be removed since they can be easily replaced by a clear one liner:

I agree it is complicated on the one hand but on the other hand, the one-liner seems long and less readable. I might prefer the one-liner eventually.

controllers/fenceagentsremediation_controller.go
```diff
@@ -163,18 +170,23 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
 		controllerutil.RemoveFinalizer(far, v1alpha1.FARFinalizer)
 		if err := r.Client.Update(context.Background(), far); err != nil {
 			return emptyResult, fmt.Errorf("failed to remove finalizer from CR - %w", err)
 		}
 		r.Log.Info("Finalizer was removed", "CR Name", req.Name)
+		commonEvents.NormalEvent(r.Recorder, far, v1alpha1.EventReasonRemoveFinalizer, v1alpha1.EventMessageRemoveFinalizer)
 		return emptyResult, nil
 	}
 	// Add FAR (medik8s) remediation taint
 	r.Log.Info("Try adding FAR (Medik8s) remediation taint", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
```
clobrano, line 184: At this point, would you consider moving the also the log in the else if branch at line 188?
 Changing the message from "Try to", to "Added taing", of course.

controllers/fenceagentsremediation_controller.go
```diff
@@ -202,6 +214,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		cmd := append([]string{far.Spec.Agent}, faParams...)
 		// The Fence Agent is excutable and the parameters structure are valid, but we don't check their values
 		r.Log.Info("Execute the fence agent", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
+		commonEvents.NormalEvent(r.Recorder, node, v1alpha1.EventReasonFenceAgentExecuted, v1alpha1.EventMessageFenceAgentExecuted)
```
clobrano, line 217: the fence angent was not executed yet a this point.  a message like "ExecuteFenceAgent", like the log above

controllers/fenceagentsremediation_controller.go
```diff
@@ -216,6 +229,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		}
 
 		r.Log.Info("Fence Agent command was finished successfully", "Fence Agent", far.Spec.Agent, "Node name", req.Name, "Response", SuccessFAResponse)
+		commonEvents.NormalEvent(r.Recorder, node, v1alpha1.EventReasonNodeRebooted, v1alpha1.EventMessageNodeRebooted)
```
clobrano, line 232: If I recall correctly, not all fence agents return successfully only after rebooting the node. For example, I remember ipmilan does it. If this is right, the event message "Node rebooted" might be misleading.

controllers/fenceagentsremediation_controller.go
```diff
@@ -225,6 +239,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		!meta.IsStatusConditionTrue(far.Status.Conditions, commonConditions.SucceededType) {
 		// Fence agent action succeeded, and now we try to remove workloads (pods and their volume attachments)
 		r.Log.Info("Manual workload deletion", "Fence Agent", far.Spec.Agent, "Node Name", req.Name)
+		commonEvents.NormalEvent(r.Recorder, node, v1alpha1.EventReasonDeleteResources, v1alpha1.EventMessageDeleteResources)
```
clobrano, line 242: Similar to the note about the message "Fence agent executed" above. This event is emitted before trying to delete the resources, it should reflect that, or it can be moved after the deletion request succeeded.

pkg/utils/nodes.go
```diff
@@ -22,17 +22,17 @@ func GetNodeWithName(r client.Reader, nodeName string) (*corev1.Node, error) {
 }
 
 // IsNodeNameValid returns an error if nodeName doesn't match any node name int the cluster, otherwise a nil
-func IsNodeNameValid(r client.Reader, nodeName string) (bool, error) {
-	_, err := GetNodeWithName(r, nodeName)
+func IsNodeNameValid(r client.Reader, nodeName string) (*corev1.Node, bool, error) {
```
clobrano, line 25: I think also the name of the function should be changed, since it does not only return if the node is valid, but the node itself (like "GetNodeWithName")

controllers/fenceagentsremediation_controller.go
razo7: Yes. That would be nicer

pkg/utils/taints.go
```diff
@@ -52,17 +52,17 @@ func CreateFARNoExecuteTaint() corev1.Taint {
 }
 
 // AppendTaint appends new taint to the taint list when it is not present, and returns error if it fails in the process
-func AppendTaint(r client.Client, nodeName string) error {
+func AppendTaint(r client.Client, nodeName string) (bool, error) {
 	// find node by name
```
clobrano, line 56: Please update the function documentation regarding the new bool parameter

clobrano: sorry, wrong line. I was referring to line 54

controllers/fenceagentsremediation_controller.go
razo7: SGTM but I will change it to fence agent succeeded instead of stating something about the node status.

controllers/fenceagentsremediation_controller.go
razo7: The idea was to send an event for every time the fence agent was executed

controllers/fenceagentsremediation_controller.go
razo7: The idea was to send an event every time we tried to delete resources in case it didn't succeed.

razo7: What makes you think it is going to be a warning event? It uses `commonEvents.NormalEvent`.

clobrano: I meant, this message originally made the reader think the deletion was already successful, while it had to be done yet.
The new message reflects instead that we will attempt to delete the resources :+1:

clobrano:
May I ask you to wait for #106 before merging this? Many events come from the fence agent execution that is reworked in that PR.

controllers/fenceagentsremediation_controller.go
```diff
@@ -218,6 +233,8 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 
 		r.Executor.Remove(far.GetUID())
 		r.Log.Info("FenceAgentsRemediation CR has completed to remediate the node", "Node Name", req.Name)
+		commonEvents.NormalEvent(r.Recorder, node, utils.EventReasonNodeRemediationCompleted, utils.EventMessageNodeRemediationCompleted)
```
clobrano, line 236: Isn't this a duplicated event?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -147,6 +146,9 @@ var _ = Describe("FAR Controller", func() {
 
 			Expect(k8sClient.Create(context.Background(), underTestFAR)).To(Succeed())
 			DeferCleanup(cleanupFar(), context.Background(), underTestFAR)
+
+			// Sleep for a second to ensure dummy reconicliation has begun running before the unit tests
```
clobrano, line 150: Typo: s/reconicliation/reconciliation
```

File: controllers/fenceagentsremediation_controller_test.go [hunk 10]
```diff
@@ -471,16 +477,75 @@ func verifyStatusCondition(far *v1alpha1.FenceAgentsRemediation, nodeName, condi
...
-	}, timeoutDeletion, pollInterval).Should(Succeed())
+	}, timeoutPostRemediation, pollInterval).Should(Succeed())
 }
 
-func verifyRemediationTaintExists(nodeName string, taint *corev1.Taint) {
-	By("Searching for remediation taint")
+// verifyPreRemediationSucceed checks if the remediation CR already has a finazliaer and a remediation taint
+func verifyPreRemediationSucceed(underTestFAR *v1alpha1.FenceAgentsRemediation, nodeName string, taint *corev1.Taint) *v1alpha1.FenceAgentsRemediation {
+	By("Searching for finalizer ")
+	Expect(k8sClient.Get(context.Background(), client.ObjectKey{Name: workerNode, Namespace: defaultNamespace}, underTestFAR)).To(Succeed())
```
clobrano, line 486: I think that using global variables to fetch data (e.g. workerNode, underTestFar, etc.) makes it prone to test coupling. It would be better to pass the values as input

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -240,21 +244,21 @@ var _ = Describe("FAR Controller", func() {
...
 
 					By("Wait some retries")
 					Eventually(func() int {
-						return plogs.CountOccurences("command failed")
+						return plogs.CountOccurences(cli.FenceAgentFailedCommandMessage)
 					}, "10s", "1s").Should(BeNumerically(">", 3))
 
-					By("Deleting the CR")
-					Expect(k8sClient.Get(context.Background(), client.ObjectKey{Name: workerNode, Namespace: defaultNamespace}, underTestFAR)).To(Succeed())
+					By("Deleting far CR")
```
clobrano, line 254: Nit: better use FAR, than "far"

controllers/fenceagentsremediation_controller.go
razo7: Duplication of which event? Can you please elaborate?

clobrano (changes requested):
I left some comments

controllers/fenceagentsremediation_controller.go
clobrano: at lines 236 and 237 it both uses 
```
commonEvents.NormalEvent(r.Recorder, node, utils.EventReasonNodeRemediationCompleted, utils.EventMessageNodeRemediationCompleted)
```

and

```
commonEvents.RemediationFinished(r.Recorder, far)
```

which, I believe, should result in similar information

razo7: Oho you mean this duplication 😅 
Yes, it is indeed redundant. I would keep the second one which is common between Medik8s operators.

controllers/fenceagentsremediation_controller_test.go
razo7: I usually address the CR kind by its [short name](https://github.com/medik8s/fence-agents-remediation/blob/main/api/v1alpha1/fenceagentsremediation_types.go#L110) - far.
I might haven't been consistent in the past, but I think it makes sense to follow this logic. At least for new changes.

controllers/fenceagentsremediation_controller_test.go
razo7: Good point

controllers/fenceagentsremediation_controller.go
razo7: Looking again it makes sense to have double events. 
One is for the node resource and one is for the far CR.

controllers/fenceagentsremediation_controller_test.go
clobrano: I understand, but "far" is also an English word, using it capitalize should be better, but no strong opinion

controllers/fenceagentsremediation_controller_test.go
clobrano: `workerNode` is a const, but it doesn't change the problem. This function will always look for a node named `workerNode`, which might be correct in this case, but making the request explicit (e.g. `verifyPreRemediationSucceded(targetNode string, ...)` would be clarer IMHO

controllers/fenceagentsremediation_controller_test.go
razo7: Good point.  IDK which way is better but for now I will use **FAR**

clobrano (approved):

PR #108, opened 2023-11-26
Randomize Unit Tests

razo7:
1. Randomize unit tests
2. Ensure unit tests are independent for randomly running the specs

### DIFF
--- a/controllers/fenceagentsremediation_controller_test.go
+++ b/controllers/fenceagentsremediation_controller_test.go
@@ -167,7 +167,7 @@ var _ = Describe("FAR Controller", func() {
 				Expect(controllerutil.ContainsFinalizer(underTestFAR, v1alpha1.FARFinalizer)).To(BeTrue())
 
 				By("Not having any test pod")
-				testPodDeletion(testPodName, true)
+				verifyPodDeleted(testPodName)
 
 				By("Verifying correct conditions for successfull remediation")
 				Expect(underTestFAR.Status.LastUpdateTime).ToNot(BeNil())
@@ -198,7 +198,7 @@ var _ = Describe("FAR Controller", func() {
 				Expect(utils.TaintExists(node.Spec.Taints, &farNoExecuteTaint)).To(BeFalse())
 
 				By("Still having one test pod")
-				testPodDeletion(testPodName, false)
+				verifyPodExists(testPodName)
 
 				By("Verifying correct conditions for unsuccessfull remediation")
 				Expect(underTestFAR.Status.LastUpdateTime).ToNot(BeNil())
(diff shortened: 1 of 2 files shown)


controllers/fenceagentsremediation_controller_test.go
```diff
@@ -285,7 +286,7 @@ func cleanupTestedResources(va1, va2 *storagev1.VolumeAttachment, pod *corev1.Po
 	podTest := &corev1.Pod{}
 	if err := k8sClient.Get(context.Background(), client.ObjectKeyFromObject(pod), podTest); err == nil {
 		log.Info("Cleanup: clean pod", "pod name", podTest.Name)
-		Expect(k8sClient.Delete(context.Background(), podTest)).To(Succeed())
+		Expect(k8sClient.Delete(context.Background(), podTest, forced)).To(Succeed())
```
clobrano, line 291: Would you do the same for "test-pod" in e2e tests?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -312,22 +315,21 @@ func cliCommandsEquality(far *v1alpha1.FenceAgentsRemediation) (bool, error) {
 
 // testPodDeletion tests whether the pod no longer exist for successful FAR CR
 // and consistently check if the pod exist and was not deleted
-func testPodDeletion(podName string, resourceDeletionWasTriggered bool) {
+func testPodDeletion(podName string, wasDeleted bool) {
```
slintes, line 318: The func name with the wasDeleted arg is confusing. I'd prefer 2 functions for clarity. E.g. `verifyPodExists` and `verifyPodDeleted`

slintes (changes requested):

slintes:
lgtm

but did we actually agree that we want to have this? I remember some discussion in chat... 🤔

razo7:
> but did we actually agree that we want to have this? I remember some discussion in chat... 🤔

IIRC we agreed not to do a repetition, but the concept of randomization was accepted.

controllers/fenceagentsremediation_controller_test.go
clobrano: So, maybe doing both in the next PR? Using this one only for randomization?

razo7: SGTM :D https://github.com/medik8s/fence-agents-remediation/pull/110

controllers/fenceagentsremediation_controller_test.go
clobrano: eventually, the two functions can simply call this one, since the code is basically the same

```
func  verifyPodDeleted(podname string) {
    checkPodIsNotFound(podname, true)
}

func checkPodIsNotFound(podName string, expected bool) {
	podKey := client.ObjectKey{
		Namespace: defaultNamespace,
		Name:      podName,
	}

	ConsistentlyWithOffset(1, func() bool {
		pod := &corev1.Pod{}
		err := k8sClient.Get(context.Background(), podKey, pod)
	}, timeoutDeletion, pollInterval).Should(Equal(expected))
}
```

razo7: >the same

They are close to being the same. One function uses `ConsistentlyWithOffset` and the other uses `EventuallyWithOffset`. But they can both use `ConsistentlyWithOffset`. SGTM

clobrano: oh, I missed that. Maybe that's why I got that problem in my e2e :sweat_smile:

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -148,9 +148,12 @@ var _ = Describe("FAR Controller", func() {
 		When("creating valid FAR CR", func() {
 			BeforeEach(func() {
 				node = utils.GetNode("", workerNode)
+				underTestFAR = getFenceAgentsRemediation(workerNode, fenceAgentIPMI, testShareParam, testNodeParam)
 			})
 			It("should have finalizer, taint, while the two VAs and one pod will be deleted", func() {
 				By("Searching for remediation taint")
+				nodeKey.Name = workerNode
```
clobrano, line 155: Modifying variables on the fly is confusing when reading the test
Why not using directly the `client.ObjectKey`?

```
client.ObjectKey{Name: workerNode, Namespace: defaultNamespace}
```

clobrano (approved):

PR #111, opened 2023-12-12
Print Remediation Time in E2E Tests

razo7:
Checking remediation time in e2e could help us identify slow remediations and identify if the PR changes impact the remediation time. We also calculate their average time for better understanding.


test/e2e/far_e2e_test.go
```diff
@@ -116,30 +117,41 @@ var _ = Describe("FAR E2e", func() {
...
+				remediationTimes = append(remediationTimes, time.Since(startTime))
 			})
 			It("should successfully remediate the second node", func() {
-				checkRemediation(nodeName, nodeBootTimeBefore, creationTimePod, pod)
+				checkRemediation(nodeName, nodeBootTimeBefore, pod)
+				remediationTimes = append(remediationTimes, time.Since(startTime))
 			})
 		})
+		AfterEach(func() {
+			if len(remediationTimes) == 2 {
```
razo7, line 145: I didn't find a better way to check that both tests are over. `AfterAll` didn't work for me (or I didn't use it properly).

clobrano: If I read [1] correctly, `AfterAll` can only be used in `Ordered` containers, which cannot be randomized though, so considering the latest changes, it doesn't seem a viable option.

[1] https://onsi.github.io/ginkgo/#setup-in-ordered-containers-beforeall-and-afterall

test/e2e/far_e2e_test.go
```diff
@@ -116,30 +117,41 @@ var _ = Describe("FAR E2e", func() {
...
 			pod.Namespace = testNsName
 			Expect(k8sClient.Create(context.Background(), pod)).To(Succeed())
-			log.Info("Tested pod has been created", "pod", testPodName)
-			creationTimePod = metav1.Now().Time
 			DeferCleanup(cleanupTestedResources, pod)
 
 			// set the node as "unhealthy" by disabling kubelet
 			makeNodeUnready(selectedNode)
 
+			startTime = time.Now()
```
clobrano, line 130: I wonder if this is accurate enough. It is taking into account the time to create the object (which is variable) other than the actual initial instant of the remediation.

razo7: > It is taking into account the time to create the object (which is variable) other than the actual initial instant of the remediation.

Are you suggesting just "storing" the time after creating far CR (after line 131) instead of before the creation? I believe it is negligible, but I don't mind doing so

clobrano: I was thinking of getting the cr creation time, but you're right that the difference should be minimal

clobrano:
/lgtm 
Giving others a chance to get more reviews (feel free to un-hold)
/hold

test/e2e/far_e2e_test.go
slintes: have a look at `AfterSuite`

slintes (changes requested):

test/e2e/far_e2e_test.go
```diff
@@ -116,30 +117,41 @@ var _ = Describe("FAR E2e", func() {
...
+				checkRemediation(nodeName, nodeBootTimeBefore, pod)
+				remediationTimes = append(remediationTimes, time.Since(startTime))
 			})
 		})
+		AfterEach(func() {
+			if len(remediationTimes) == 2 {
+				averageTimeDuration := 0.0
+				for index, remTime := range remediationTimes {
+					averageTimeDuration += remTime.Seconds()
+					fmt.Printf("\nRemediation time #%d: %s\n", index+1, remTime)
```
slintes, line 149: please use the logger `log` instead of `fmt.Print`

razo7: I see that have used fmt.Print in more places in the e2e test file, so I will update them all.

test/e2e/far_e2e_test.go
```diff
@@ -134,27 +135,28 @@ var _ = Describe("FAR E2e", func() {
 		When("running FAR to reboot two nodes", func() {
 			It("should successfully remediate the first node", func() {
 				checkRemediation(nodeName, nodeBootTimeBefore, pod)
-				remediationTimes = append(remediationTimes, time.Since(startTime))
+				remediationsTime = append(remediationsTime, time.Since(startTime))
```
slintes, line 138: hm, I think the old name was correct... 🤔 
multiple time*s* of a single remediation
this now sounds like overall time of all remediation*s*

slintes (approved):
/retest
/hold cancel

PR #113, opened 2023-12-14
Shorten Timeouts for E2e Tests

razo7:
- Shorten Timeouts for e2e tests
- Use shorter human-readable time
- ~Don't look for the fence agent success message from the FAR log which is redundant, less strong, and won't be needed after #106~ Done in #106

### DIFF
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -37,11 +37,11 @@ const (
 
 	//TODO: try to minimize timeout
 	// eventually parameters
-	timeoutLogs     = 3 * time.Minute
-	timeoutReboot   = 6 * time.Minute  // fencing with fence_aws should be completed within 6 minutes
-	timeoutDeletion = 10 * time.Second // this timeout is used after all the other steps have been succesfult
-	pollDeletion    = 250 * time.Millisecond
-	pollInterval    = 10 * time.Second
+	timeoutLogs   = "3m0s"
+	timeoutReboot = "6m0s" // fencing with fence_aws should be completed within 6 minutes
+	timeoutShort  = "5s"   // this timeout is used after all the other steps have been succesfult
+	pollShort     = "250ms"
+	pollInterval  = "1s"
 )
 
 var remediationTimes []time.Duration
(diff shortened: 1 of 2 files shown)


slintes:
because of

>  won't be needed after ...

this should not be merged before the other PR, otherwise we miss a test...

/hold

razo7:
> this should not be merged before the other PR, otherwise we miss a test...

It is a step in a test, and not a full test but I agree that it is better to wait.
However, having the `checkFarLogs` step will fail the e2e due to the #112 change since `checkFarLogs` checks the logs of the [first FAR running pod](https://github.com/medik8s/fence-agents-remediation/blob/main/pkg/utils/pods.go#L35-L38) (e.g., see failing e2e tests at #107), when we have now two running pods at any time.

> checkFarLogs is already removed in https://github.com/medik8s/fence-agents-remediation/pull/106

Correct, so either we merge this PR or the other one, otherwise every e2e test could fail due to the #112 change and we can't merge the remaining PRs.
I lean to merge this small PR first rather than the other large one.

clobrano:
> I lean to merge this small PR first rather than the other large one. 

Please do not. It's been already a bit tricky to rebase #106 onto #108, I'd rather get the first in before moving on with further changes to e2e tests.

test/e2e/far_e2e_test.go
```diff
@@ -40,11 +40,11 @@ const (
...
 	// eventually parameters
-	timeoutLogs     = 3 * time.Minute
-	timeoutReboot   = 6 * time.Minute  // fencing with fence_aws should be completed within 6 minutes
-	timeoutDeletion = 10 * time.Second // this timeout is used after all the other steps have been succesfult
-	pollDeletion    = 250 * time.Millisecond
-	pollInterval    = 10 * time.Second
+	timeoutLogs   = "3m0s"
+	timeoutReboot = "6m0s" // fencing with fence_aws should be completed within 6 minutes
+	timeoutShort  = "5s"   // this timeout is used after all the other steps have been succesfult
+	pollShort     = "250ms"
```
mshitrit, line 46: why change the var names with suffix `Short` , I think in this context long/short time are very subjective (especially since the "long" time doesn't exist anymore as a baseline to compare to) terms so I prefer a more meaningful name.

razo7: I agree, it was just short from my POV.

test/e2e/far_e2e_test.go
```diff
@@ -40,11 +40,12 @@ const (
...
-	timeoutLogs     = 3 * time.Minute
-	timeoutReboot   = 6 * time.Minute  // fencing with fence_aws should be completed within 6 minutes
-	timeoutDeletion = 10 * time.Second // this timeout is used after all the other steps have been succesfult
-	pollDeletion    = 250 * time.Millisecond
-	pollInterval    = 10 * time.Second
+	timeoutLogs        = "3m0s"
+	timeoutTaint       = "2s"   // Timeout for checking the FAR taint
+	timeoutReboot      = "6m0s" // fencing with fence_aws should be completed within 6 minutes
+	timeoutAfterReboot = "5s"   // Timeout for verifying steps  after the node has been rebooted
+	pollShort          = "250ms"
```
mshitrit, line 47: I think you've missed that one

PR #116, opened 2024-01-01
Restrict Fence Agent Command

razo7:
Verify and restrict far CRs to specific (and available) agents that exist in the container/pod.

1. Print on start-up the amount and list of available agents.
2. Validate CR agent name with the available agents.
3. Add new unit-test for creating CR with unknown agent name

[ECOPROJECT-1753](https://issues.redhat.com//browse/ECOPROJECT-1753)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,5 +41,9 @@ RUN dnf install -y dnf-plugins-core \
     && dnf install -y fence-agents-all fence-agents-aws fence-agents-azure-arm fence-agents-gce \
     && dnf clean all -y
 
+# Store the avavliable agents names (fence_*) into readonly fence_agents_list file
+RUN ls /usr/sbin/fence* > fence_agents_list \
+    && chmod 444 fence_agents_list
+
 USER 65532:65532
 ENTRYPOINT ["/manager"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,7 @@ RUN dnf install -y dnf-plugins-core \
     && dnf clean all -y
 
 # Store the avavliable agents names (fence_*) into readonly fence_agents_list file
-RUN ls /usr/sbin/fence_* | grep -oP '(?<=/usr/sbin/)(fence_[^/]*)' > fence_agents_list \
+RUN ls basename /usr/sbin | grep fence_ > fence_agents_list \
     && chmod 444 fence_agents_list
 
 USER 65532:65532
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,7 @@ RUN dnf install -y dnf-plugins-core \
     && dnf clean all -y
 
 # Store the avavliable agents names (fence_*) into readonly fence_agents_list file
-RUN ls /usr/sbin/fence* > fence_agents_list \
+RUN ls /usr/sbin/fence_* | grep -oP '(?<=/usr/sbin/)(fence_[^/]*)' > fence_agents_list \
     && chmod 444 fence_agents_list
 
 USER 65532:65532
--- a/controllers/controllers_suite_test.go
+++ b/controllers/controllers_suite_test.go
@@ -134,15 +134,17 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(k8sClient).NotTo(BeNil())
 
-	executor, _ := cli.NewFakeExecuter(k8sClient, controlledRun)
+	fakeExecutor, _ := cli.NewFakeExecuter(k8sClient, controlledRun)
 
 	os.Setenv("DEPLOYMENT_NAMESPACE", defaultNamespace)
 
+	agentsList := []string{"fence_ipmilan"}
 	err = (&FenceAgentsRemediationReconciler{
-		Client:   k8sClient,
-		Log:      k8sManager.GetLogger().WithName("test far reconciler"),
-		Scheme:   k8sManager.GetScheme(),
-		Executor: executor,
+		Client:     k8sClient,
+		Log:        k8sManager.GetLogger().WithName("test far reconciler"),
+		Scheme:     k8sManager.GetScheme(),
+		Executor:   fakeExecutor,
+		AgentsList: agentsList,
 	}).SetupWithManager(k8sManager)
 	Expect(err).NotTo(HaveOccurred())
 
--- a/controllers/controllers_suite_test.go
+++ b/controllers/controllers_suite_test.go
@@ -140,11 +140,11 @@ var _ = BeforeSuite(func() {
 
 	agentsList := []string{"fence_ipmilan"}
 	err = (&FenceAgentsRemediationReconciler{
-		Client:     k8sClient,
-		Log:        k8sManager.GetLogger().WithName("test far reconciler"),
-		Scheme:     k8sManager.GetScheme(),
-		Executor:   fakeExecutor,
-		AgentsList: agentsList,
+		Client:          k8sClient,
+		Log:             k8sManager.GetLogger().WithName("test far reconciler"),
+		Scheme:          k8sManager.GetScheme(),
+		Executor:        fakeExecutor,
+		AgentsWhiteList: agentsList,
 	}).SetupWithManager(k8sManager)
 	Expect(err).NotTo(HaveOccurred())
 
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -177,12 +177,14 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 
 	// Validate whetehr the fence agent is supported
 	r.Log.Info("Check if the agent is supported")
-	if err := r.checkAgentSupport(far.Spec.Agent); err != nil {
-		updateErr := utils.UpdateConditions(utils.FenceAgentNotSupported, far, r.Log)
-		if updateErr != nil {
-			err = fmt.Errorf("validion error for fence agent: %w, and status error: %w", updateErr, err)
+	if isSupported := r.isAgentSupported(far.Spec.Agent); !isSupported {
+		unsuppportedErr := fmt.Errorf("fence agent name didn't match one of the available supported agents")
+		if err := utils.UpdateConditions(utils.FenceAgentNotSupported, far, r.Log); err != nil {
+			unsuppportedErr = fmt.Errorf("%w and it failed to update conditons for reason %s and error %w", unsuppportedErr, utils.FenceAgentNotSupported, err)
 		}
-		return emptyResult, err
+		r.Log.Error(unsuppportedErr, "Fence agent is not supported", "CR's Name", req.Name, "Fence Agent", far.Spec.Agent)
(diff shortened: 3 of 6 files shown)


mshitrit:
Following our [discussion](https://redhat-internal.slack.com/archives/C03M5GKJNBA/p1704094840480129) this PR should be the implementation of the **Second** option (also pasting here the relevant part).

```

- At the moment there is no validation for fence agent name so this allow a security breach for the user to execute unwanted commands.
- There are several options on how  to address this:
  - First (simple) option: is a hard coded list of fence agents that will be used for validation
  - Second (advanced) option: retrieving the fence agents name during runtime and use that data for validation.
- The main pro for using the Second option that it'll support  different setups of fence agents (mainly keeping in mind upstream / downstream differences ) where the con is that it is more complicated.

```

- I couldn't find the logic part that populate the valid fence agents :thinking: 
- I don't see a value of using a file, I think it is less secure and more complicated than the alternative.
is there a particular reason to use it that I am missing ?

Dockerfile
```diff
@@ -41,5 +41,9 @@ RUN dnf install -y dnf-plugins-core \
     && dnf install -y fence-agents-all fence-agents-aws fence-agents-azure-arm fence-agents-gce \
     && dnf clean all -y
 
+# Store the avavliable agents names (fence_*) into readonly fence_agents_list file
+RUN ls /usr/sbin/fence* > fence_agents_list \
```
mshitrit, line 45: fence_agents_list should probably contain a list of the names of the  fence agents and not list of paths of the fence agent

controllers/fenceagentsremediation_controller.go
```diff
@@ -265,6 +277,16 @@ func (r *FenceAgentsRemediationReconciler) updateStatus(ctx context.Context, far
 	return nil
 }
 
+// checkAgentSupport return error if the agent name from the CR is not matching one of the available agents
+func (r *FenceAgentsRemediationReconciler) checkAgentSupport(agent string) error {
```
mshitrit, line 281: I think this method should return a bool.
error is usually reserved for a use case when something unexpected went wrong.

controllers/fenceagentsremediation_controller.go
```diff
@@ -173,6 +174,17 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
 	}
+
+	// Validate whetehr the fence agent is supported
+	r.Log.Info("Check if the agent is supported")
+	if err := r.checkAgentSupport(far.Spec.Agent); err != nil {
+		updateErr := utils.UpdateConditions(utils.FenceAgentNotSupported, far, r.Log)
+		if updateErr != nil {
+			err = fmt.Errorf("validion error for fence agent: %w, and status error: %w", updateErr, err)
+		}
+		return emptyResult, err
```
mshitrit, line 185: IIUC no value in returning an error here since there isn't a chance that this will be resolved in another reconcile

razo7: > since there isn't a chance that this will be resolved in another reconcile

The only chance is by updating the fence agent's name, but still, we can potentially print once and not multiple times.

Dockerfile
razo7: Makes sense :)

main.go
```diff
@@ -143,3 +156,43 @@ func printVersion() {
...
 	setupLog.Info(fmt.Sprintf("Build Date: %s", version.BuildDate))
 }
+
+func printSupportedFenceAgents(agentList []string) {
+	setupLog.Info(fmt.Sprintf("Fence agents: %d", len(agentList)))
+	setupLog.Info(fmt.Sprintf("Available agents: %s", strings.Join(agentList, " ")))
+}
+
+// checkFenceAgentsFileExists check if the file exists, read it, and then remove redundant prefix
+func checkFenceAgentsFileExists(filePath string) ([]string, error) {
```
mshitrit, line 166: I think method name should be changed - it mainly extract the fence agents from the file.
IMO `getFenceAgents` (or something similar)  is a better name.

Another option is to break this method into two methods (one that checks for the file and one that retrieves the file)

I'm leaning into the first option of just changing the name but I think both options are fine.

main.go
```diff
@@ -143,3 +156,43 @@ func printVersion() {
...
+	if _, err := os.Stat(filePath); err != nil {
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("fence agents file not found: %v", err)
+		}
+		return nil, fmt.Errorf("failed to check for supported fence agents list: %v", err)
+	}
+
+	content, err := os.ReadFile(filePath)
```
mshitrit, line 174: Nit: I think file or fenceAgentFile is a better name than content

main.go
```diff
@@ -143,3 +156,43 @@ func printVersion() {
...
+	if err != nil {
+		return nil, fmt.Errorf("error reading file: %v", err)
+	}
+	contentStr := string(content)
```
mshitrit, line 178: missing `defer content.Close()` before this line

mshitrit (changes requested):

controllers/fenceagentsremediation_controller.go
mshitrit: Good point,
However IIUC it's pretty unlikely, since this name is retrivied from the CR which is created automatically by NHC and not user created.
it is much more likely that the fence agent name will be changed in the Template or Configuration (I can't remember which one is the origin :sweat_smile: )

main.go
```diff
@@ -143,3 +156,43 @@ func printVersion() {
 	setupLog.Info(fmt.Sprintf("Git Commit: %s", version.GitCommit))
 	setupLog.Info(fmt.Sprintf("Build Date: %s", version.BuildDate))
 }
+
+func printSupportedFenceAgents(agentList []string) {
+	setupLog.Info(fmt.Sprintf("Fence agents: %d", len(agentList)))
+	setupLog.Info(fmt.Sprintf("Available agents: %s", strings.Join(agentList, " ")))
+}
+
```
mshitrit, line 164: Nit: I think this method can be removed.
Either by merging both into a one line, or just removing the first line (which is redundant IMO).

controllers/fenceagentsremediation_controller.go
razo7: >error is usually reserved for a use case when something unexpected went wrong.

There are a few points in this function when an error might occur (file is not found, file is empty or we can't read the file). Do you suggest ignoring them or printing log error and exit error or (as it is right now) returning a unique error which will be checked and printed when `checkAgentSupport` is called

controllers/fenceagentsremediation_controller.go
razo7: >created automatically by NHC and not user created.

fine by me.

>Template or Configuration (I can't remember which one is the origin 😅 )

Just template.

controllers/fenceagentsremediation_controller.go
mshitrit: I think you might refer to `checkFenceAgentsFileExists` method in main.
At the moment as far as I can see this method only returns an error in one place which will occur if the agent searched is not in the list.

main.go
razo7: The first line could be nice information for a user to see how many options (of fence agents) FAR supports. It could be a nice indication for us to review this line whenever FAR supports a different number of agents instead of creating CR and verifying that.
I don't have a strong preference.

controllers/fenceagentsremediation_controller.go
razo7: >I think you might refer to checkFenceAgentsFileExists method in main.

Correct, I got confused with the names.

Dockerfile
```diff
@@ -41,5 +41,9 @@ RUN dnf install -y dnf-plugins-core \
     && dnf install -y fence-agents-all fence-agents-aws fence-agents-azure-arm fence-agents-gce \
     && dnf clean all -y
 
+# Store the avavliable agents names (fence_*) into readonly fence_agents_list file
+RUN ls /usr/sbin/fence_* | grep -oP '(?<=/usr/sbin/)(fence_[^/]*)' > fence_agents_list \
```
clobrano, line 45: `ls basename /usr/sbin | grep fence_` should be easier to understand

controllers/controllers_suite_test.go
```diff
@@ -134,15 +134,17 @@ var _ = BeforeSuite(func() {
...
 	err = (&FenceAgentsRemediationReconciler{
-		Client:   k8sClient,
-		Log:      k8sManager.GetLogger().WithName("test far reconciler"),
-		Scheme:   k8sManager.GetScheme(),
-		Executor: executor,
+		Client:     k8sClient,
+		Log:        k8sManager.GetLogger().WithName("test far reconciler"),
+		Scheme:     k8sManager.GetScheme(),
+		Executor:   fakeExecutor,
+		AgentsList: agentsList,
```
clobrano, line 147: nit: I think that `AgentsWhiteList` might be an even clearer naming

controllers/fenceagentsremediation_controller.go
```diff
@@ -173,6 +174,19 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Info("Finalizer was removed", "CR Name", req.Name)
 		return emptyResult, nil
 	}
+
+	// Validate whetehr the fence agent is supported
+	r.Log.Info("Check if the agent is supported")
+	if isSupported := r.isAgentSupported(far.Spec.Agent); !isSupported {
+		unsuppportedErr := fmt.Errorf("fence agent name didn't match one of the available supported agents")
```
clobrano, line 181: typo: s/unsuppported/unsupported

Also, might be clearer using a simpler error message like "unsupported fence agent", even with the name that failed

controllers/fenceagentsremediation_controller.go
```diff
@@ -265,6 +279,16 @@ func (r *FenceAgentsRemediationReconciler) updateStatus(ctx context.Context, far
 	return nil
 }
 
+// isAgentSupported return true if the agent name from the CR is matching one of the available agents, and false otherwise
```
clobrano, line 282: typo: s/return/returns, and I'd say "...the CR matches" more than "is matching"

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -96,7 +97,7 @@ var _ = Describe("FAR Controller", func() {
 		})
 
 		Context("buildFenceAgentParams", func() {
-			When("FAR include different action than reboot", func() {
+			When("far CR include different action than reboot", func() {
```
clobrano, line 100: Same observation of the other PR. Better keep capitalized FAR here IMHO

clobrano (changes requested):

controllers/fenceagentsremediation_controller.go
razo7: >even with the name that failed

I already add it in the [next log print](https://github.com/medik8s/fence-agents-remediation/pull/116/files#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R185) as ` "Fence Agent", far.Spec.Agent`

razo7:
/hold
We might want to to this check in Webhook

PR #117, opened 2024-01-02
Improve Logs from Updating Conditions

razo7:
The `UpdateConditions` function has an error only when an unknown agent has been found. Thus, there is no need to return and check it from reconcile, since it is not an error from trying to update it. 
FAR can simply log an informative error when it was called when unknown ConditionsChangeReason.
Moreover, the lastTrinsitionTime print is more human friendly

### DIFF
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -124,7 +124,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 	}
 	if !valid {
 		r.Log.Error(err, "Didn't find a node matching the CR's name", "CR's Name", req.Name)
-		err := utils.UpdateConditions(utils.RemediationFinishedNodeNotFound, far, r.Log)
+		utils.UpdateConditions(utils.RemediationFinishedNodeNotFound, far, r.Log)
 		return emptyResult, err
 	}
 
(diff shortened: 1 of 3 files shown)


controllers/fenceagentsremediation_controller.go
```diff
@@ -46,6 +46,7 @@ const (
 	// errors
 	errorMissingParams     = "nodeParameters or sharedParameters or both are missing, and they cannot be empty"
 	errorMissingNodeParams = "node parameter is required, and cannot be empty"
+	errFailUpdate          = "Failed to update conditons"
```
clobrano, line 49: typo: s/conditons/conditions

controllers/fenceagentsremediation_controller.go
```diff
@@ -123,16 +124,21 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		return emptyResult, err
 	}
 	if !valid {
-		r.Log.Error(err, "Didn't find a node matching the CR's name", "CR's Name", req.Name)
-		err := utils.UpdateConditions(utils.RemediationFinishedNodeNotFound, far, r.Log)
+		notFoundErr := fmt.Errorf("didn't find a node matching the CR's name")
+		if err := utils.UpdateConditions(utils.RemediationFinishedNodeNotFound, far, r.Log); err != nil {
+			notFoundErr = fmt.Errorf("failed to update conditons for reason %s, and error: %v", utils.RemediationFinishedNodeNotFound, err)
```
clobrano, line 129: Maybe you could wrap the previous error too
Also typo s/conditons/conditions

clobrano (changes requested):
I wonder if all the error messages after UpdateStatus failures could be put inside UpdateStatus function itself. I also added some other comments in the code.

razo7:
>I wonder if all the error messages after UpdateStatus failures could be put inside UpdateStatus function itself.

They can but this way we log an error in the function and Reconcile

clobrano:
> > I wonder if all the error messages after UpdateStatus failures could be put inside UpdateStatus function itself.
> 
> They can but this way we log an error in the function and Reconcile

Sorry, I meant inside `utils.UpdateConditions`.

IIUC the additional information this new logs provide is the CR's name, which is already accessible inside `utils.UpdateConditions`, so we could simply update the error logs that are inside `utils.UpdateConditions`

razo7:
SGTM

clobrano (approved):

PR #118, opened 2024-01-02
[WIP] Minimize FAR Image

razo7:
Building FAR using `debian-slim` for smaller image
Using half the previous image size is better for downloading the image and minimizing any potential risks from running a larger image (with some redundant files).

TODO: Update [e2e test environment](https://github.com/openshift/release/blob/master/ci-operator/config/medik8s/fence-agents-remediation/medik8s-fence-agents-remediation-main__4.15.yaml#L7) to build FAR using debian-slim and not CentOS

[ECOPROJECT-1757](https://issues.redhat.com//browse/ECOPROJECT-1757)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,16 +30,16 @@ COPY .git/ .git/
 # Build
 RUN ./hack/build.sh
 
-FROM quay.io/centos/centos:stream8
+FROM debian:stable-slim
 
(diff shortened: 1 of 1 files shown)


razo7:
It won't be investigated soon and it isn't highly important for now. Closing for now.

PR #119, opened 2024-01-04
Validate Agent Name by Format and Binary

razo7:
Agent field in far or farTemplate has not been verified.

1. We add format validation by Kubebuilder (should have `fence_` prefix)
2. We add Webhook validation for whether a fence agent file exists with this agent name (under `/usr/sbin` directory)

ECOPROJECT-1753


api/v1alpha1/fenceagentsremediation_webhook_test.go
```diff
@@ -0,0 +1,65 @@
...
+				_, err := far.ValidateCreate()
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring(validation.ErrorNotFoundAgent, invalidAgentName))
+			})
+		})
+	})
+
+	Context("updating FenceAgentsRemediation", func() {
+
+		oldFAR := getTestFAR(invalidAgentName)
```
clobrano, line 36: Shouldn't this variable go into a BeforeEach context?

pkg/validation/validation.go
```diff
@@ -0,0 +1,51 @@
...
+// isAgentSupported returns true if the agent name matches a binary, and false otherwise
+func (vfe *validateAgentExistence) isAgentSupported(agent string) (bool, error) {
+	directory := "/usr/sbin/"
+	// Create the full path by joining the directory and filename
+	fullPath := filepath.Join(directory, agent)
+
+	// Check if the file exists
+	_, err := vfe.isFileExist(fullPath)
+	if err == nil {
+		fmt.Printf("Agent %s was found at %s\n", agent, directory)
```
clobrano, line 44: Probably we can drop this `fmt.Printf`, I don't think they are visible in the logs

api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
```diff
@@ -0,0 +1,69 @@
...
+				_, err := farTemplate.ValidateCreate()
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring(validation.ErrorNotFoundAgent, invalidAgentName))
+			})
+		})
+	})
+
+	Context("updating FenceAgentsRemediationTemplate", func() {
+
+		oldFARTemplate := getTestFARTemplate(invalidAgentName)
```
clobrano, line 36: Should this go into a `BeforeEach`?

clobrano (changes requested):
I left some comments

clobrano:
Sorry, somehow my editor pushed the same review twices

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
```diff
@@ -0,0 +1,67 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/medik8s/fence-agents-remediation/pkg/validation"
+)
+
+var (
+	// webhookTemplateLog is for logging in this package.
+	webhookFARTemplateLog = logf.Log.WithName("fenceagentsremediationtemplate-resource")
+	// verify agent existence with os.Stat function
+	agentValidator = validation.NewAgentValidator(func(agent string) (fs.FileInfo, error) { return os.Stat(agent) })
```
slintes, line 36: - since FAR is partof FARTemplate, and we want to validate FAR's agend, I would prefer to put actual validation and everything around into the FAR webhook
- since validation moved into its own package (which is ok), the webhook should not care about how validation happens. What about having a default validator which uses the filesystem in the validation package?

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
```diff
@@ -0,0 +1,67 @@
...
+// TODO(user): EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
+// +kubebuilder:webhook:path=/validate-fence-agents-remediation-medik8s-io-v1alpha1-fenceagentsremediationtemplate,mutating=false,failurePolicy=fail,sideEffects=None,groups=fence-agents-remediation.medik8s.io,resources=fenceagentsremediationtemplates,verbs=create;update,versions=v1alpha1,name=vfenceagentsremediationtemplate.kb.io,admissionReviewVersions=v1
+var _ webhook.Validator = &FenceAgentsRemediationTemplate{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (farTemplate *FenceAgentsRemediationTemplate) ValidateCreate() (admission.Warnings, error) {
+	webhookFARTemplateLog.Info("validate create", "name", farTemplate.Name)
+	return agentValidator.ValidateAgentName(farTemplate.Spec.Template.Spec.Agent)
```
slintes, line 54: The validator should not know about webhook specifics, so it should not return admission.Warnings IMHO.
In order to prevent multiple conversions from validation result to webhook result, you can add a function which does that.

pkg/validation/validation.go
```diff
@@ -0,0 +1,51 @@
...
+	"os"
+	"path/filepath"
+
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const ErrorNotFoundAgent = "unsupported fence agent %s:"
+
+type validateAgentExistence struct {
+	isFileExist func(string) (fs.FileInfo, error)
```
slintes, line 15: - for better readabilty, what about introducing a funtion type instead of using the finction ignatire at multiple places?
- FileInfo is never used (which good, because:)
- name should be independent of the implementation, so maybe don't use anything about files at all?

pkg/validation/validation.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type AgentValidator interface {
+	ValidateAgentName(agent string) (admission.Warnings, error)
+}
+
+func NewAgentValidator(fileGetter func(string) (fs.FileInfo, error)) AgentValidator {
```
slintes, line 22: the validator interface should not care about files, that's just the default validation method.

pkg/validation/validation.go
```diff
@@ -0,0 +1,51 @@
...
+	return &validateAgentExistence{isFileExist: fileGetter}
+}
+
+func (vfe *validateAgentExistence) ValidateAgentName(agent string) (admission.Warnings, error) {
+	exist, err := vfe.isAgentSupported(agent)
```
slintes, line 27: - what's the value of calling another function?
- error handling is... strange? Probably because exists is always false when err != nil? That's confusing.

pkg/validation/validation.go
```diff
@@ -0,0 +1,51 @@
...
+	if !exist {
+		return admission.Warnings{"Check the agent spec field"}, fmt.Errorf(ErrorNotFoundAgent+" %w", agent, err)
```
slintes, line 29: - the validator should not care about webhook specifics IMHO
- why adding a warning? The error should contain a clear message about what's wrong. IMHO warnings should be about issues which don't fail validation but e.g. can be improved... :thinking:

slintes (changes requested):
I found it a bit complicated to read the changes. 
IMHO "separation of concerns" can be improved, by not using anything about files in the webhook, and not using anything about webhooks in the validator.

I wasn't sure if the result is that much better, so I gave it a try. Code might clarify my comments :D Feel free to pick those changes where you agree that it's better.

https://github.com/slintes/fence-agents-remediation/commit/e87ead509a4bb794f87e30054dc6676f36aa706d

pkg/validation/validation.go
razo7: I am pretty confident they do, but it might be better to properly use logs function rather than _Printf_

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
razo7: I agree, it felt a little bit "ugly" to pass the admission.Warnings in multiple places

pkg/validation/validation.go
slintes: I'm pretty confident that using Print is a no go 😉

clobrano: > I am pretty confident they do

maybe in the shell, because it both catches proper logs and stdout?

slintes: I think logs go to stdout as well by default, it's about the format of the log message which is missing when using Print

PR #120, opened 2024-01-07
Initialize Event Recorder for executer

razo7:
- Initialization was missing, and [`e.recorder`](https://github.com/medik8s/fence-agents-remediation/blob/main/pkg/cli/cliexecuter.go#L187) was nil.
- Fix unit-tests

Fix [ECOPROJECT-1816](https://issues.redhat.com//browse/ECOPROJECT-1816)

### DIFF
--- a/controllers/controllers_suite_test.go
+++ b/controllers/controllers_suite_test.go
@@ -48,13 +48,12 @@ import (
 const defaultNamespace = "default"
 
 var (
-	k8sClient             client.Client
-	k8sManager            manager.Manager
-	testEnv               *envtest.Environment
-	ctx                   context.Context
-	cancel                context.CancelFunc
-	fakeReconcileRecorder *record.FakeRecorder
-	fakeExecRecorder      *record.FakeRecorder
+	k8sClient    client.Client
+	k8sManager   manager.Manager
+	testEnv      *envtest.Environment
+	ctx          context.Context
+	cancel       context.CancelFunc
+	fakeRecorder *record.FakeRecorder
 
 	plogs *peekLogger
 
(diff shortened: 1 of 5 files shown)


mshitrit:
I think a UT capturing an event would be nice

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -51,6 +52,10 @@ const (
 	timeoutPreRemediation  = "1s" // this timeout is used for the other steps that occur before remediation is completed
 	timeoutPostRemediation = "2s" // this timeout is used for the other steps that occur after remediation is completed
 	pollInterval           = "200ms"
+
+	// eventSteps
+	eventExixst    = "Verifying that event %s was created from"
+	eventNotExixst = "Verifying that event %s was not created from"
```
mshitrit, line 58: Typo: Exixst

controllers/controllers_suite_test.go
```diff
@@ -48,12 +48,13 @@ import (
...
-	ctx          context.Context
-	cancel       context.CancelFunc
-	fakeRecorder *record.FakeRecorder
+	k8sClient             client.Client
+	k8sManager            manager.Manager
+	testEnv               *envtest.Environment
+	ctx                   context.Context
+	cancel                context.CancelFunc
+	fakeReconcileRecorder *record.FakeRecorder
+	fakeExecRecorder      *record.FakeRecorder
```
mshitrit, line 57: I think it's fine to use the same recorder in the test code, it'll simplify the code significantly and I don't think there is a major trade-off as far as test quality.

razo7: Indeed. I was not sure what is better

PR #121, opened 2024-01-08
Update Readme with the optional retry CR parameters

clobrano:


README.md
```diff
@@ -96,7 +96,19 @@ FAR CRs are deleted by NHC after it sees the Node is healthy again.
 
 #### Example CR
 
-The FAR, `FenceAgentsRemediation`, CR is created by the admin and is used to trigger the fence-agent on a specific node. The CR includes an *agent* field for the fence-agent name, *sharedparameters* field with all the shared, not specific to a node, parameters, and *nodeparameters* field to specify the parameters for the fenced node.
+The FAR CR (CustomResource), created by the admin, is used to trigger the fence-agent on a specific node.
+
+The CR includes the following parameters:
+* `agent` - the fence-agent name
+* `sharedparameters` - all the shared, not specific to a node, parameters
```
razo7, line 103:
```suggestion
* `sharedparameters` - common and shared parameters between the cluster nodes.
```

README.md
```diff
@@ -96,7 +96,19 @@ FAR CRs are deleted by NHC after it sees the Node is healthy again.
...
+* `nodeparameters` - the parameters for the fenced node
```
razo7, line 104:
```suggestion
* `nodeparameters` - specific cluster nodes information for the fenced node
```

README.md
```diff
@@ -96,7 +96,19 @@ FAR CRs are deleted by NHC after it sees the Node is healthy again.
...
+ 
+Optionally, the CR can include further parameters to specify how the operator should retry the fence-agent command in case of failure:
+
+* `.spec.retrycount` - the number of times to retry the fence-agent in case of failure. Default is 5.
+* `.spec.retryinterval` - the interval between retries in seconds. Default is "5s".
+* `.spec.timeout` - the timeout for the fence-agent in seconds. Default is "60s".
```
razo7, line 110: I am not sure the `.spec.` prefix is needed. 

Can you also add these fields in the below YAML example (114-135)

README.md
```diff
@@ -96,7 +96,19 @@ FAR CRs are deleted by NHC after it sees the Node is healthy again.
...
 
-The FAR, `FenceAgentsRemediation`, CR is created by the admin and is used to trigger the fence-agent on a specific node. The CR includes an *agent* field for the fence-agent name, *sharedparameters* field with all the shared, not specific to a node, parameters, and *nodeparameters* field to specify the parameters for the fenced node.
+The FAR CR (CustomResource), created by the admin, is used to trigger the fence-agent on a specific node.
+
+The CR includes the following parameters:
+* `agent` - the fence-agent name
+* `sharedparameters` - all the shared, not specific to a node, parameters
+* `nodeparameters` - the parameters for the fenced node
+ 
+Optionally, the CR can include further parameters to specify how the operator should retry the fence-agent command in case of failure:
```
razo7, line 106: I wouldn't say that the `the CR can include` as it will always include these fields (by default or by manually adding them).

I would even append these fields/parameters on the above bullet list and combine this line with [line 101](https://github.com/medik8s/fence-agents-remediation/pull/121/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R101)

clobrano: It makes sense, maybe just indicating that they are optional with default values

razo7: All the fields are optional, except _agent_, so they are not different than others. What is "unique" is that they have default values that other parameters don't have.

README.md
```diff
@@ -96,7 +96,17 @@ FAR CRs are deleted by NHC after it sees the Node is healthy again.
...
+The FAR CR (CustomResource), created by the admin, is used to trigger the fence-agent on a specific node.
+
+The CR includes the following parameters:
+
+* `agent` - the fence-agent name
+* `sharedparameters` - common cluster nodes information for executing the fence agent
+* `nodeparameters` - specific cluster nodes information for executing the fence agent
+* `retrycount` - the number of times to retry the fence-agent in case of failure. Optional parameter, default is 5.
+* `retryinterval` - the interval between retries in seconds. Optional parameter, default is "5s".
+* `timeout` - the timeout for the fence-agent in seconds. Optional parameter, default is "60s".
```
razo7, line 108: NIT: I am unsure if I would use the word `the` at the beginning of each parameter explanation. It is fine without it, but it is just a minor preference, so feel free to ignore my comment.

README.md
```diff
@@ -96,7 +96,17 @@ FAR CRs are deleted by NHC after it sees the Node is healthy again.
...
 #### Example CR
 
-The FAR, `FenceAgentsRemediation`, CR is created by the admin and is used to trigger the fence-agent on a specific node. The CR includes an *agent* field for the fence-agent name, *sharedparameters* field with all the shared, not specific to a node, parameters, and *nodeparameters* field to specify the parameters for the fenced node.
+The FAR CR (CustomResource), created by the admin, is used to trigger the fence-agent on a specific node.
+
+The CR includes the following parameters:
+
+* `agent` - the fence-agent name
+* `sharedparameters` - common cluster nodes information for executing the fence agent
+* `nodeparameters` - specific cluster nodes information for executing the fence agent
```
razo7, line 105: Mentioning the next three parameters as optional and not mentioning these two as optional means that they are not optional.
In practice they are [optional](https://github.com/medik8s/fence-agents-remediation/blob/main/api/v1alpha1/fenceagentsremediation_types.go#L80-L86), even though I don't see a case where a user won't use/set them, thus I am leaning towards setting them as non-optional parameters or simply remove the `omitempty`. WDYT?

razo7 (approved):
/lgtm
left a nit and a general question that came to me from your PR.
Please update PR description and link it to the old and relevant PR
/hold
Feel free to unhold if want to discard my NIT

README.md
clobrano: I guess they are optional just because it's up to the user to decide which specific parameters they want, but you're right that this way is confusing. I might rephrase the last three as "If unset, the default value is..." ?

README.md
clobrano: I prefer with the article honestly, but you're right that both ways sound correct

README.md
clobrano: > it's up to the user to decide which specific parameters they want

and IF they want it. I think it is ok to keep them optional

razo7: >"If unset, the default value is..." ?

Or just remove the _optional_ words. `Optional parameter, default is "5s".` -> `Default is "5s".`
I don't have a strong preference here.

> I think it is ok to keep them optional

Agreed. My question was more about the `sharedparameters` and `nodeparameters` which maybe should not be optional. Without them running the fence agent would fail as it misses some parameters.

clobrano: I was referring to sharedparameters and nodeparameters too :)

> Without them running the fence agent would fail as it misses some parameters

but I didn't realized that. They're surely not optional then

README.md
clobrano: I changed my mind, there would be a mix of descriptions some with the article and some without. Let's make them all without

README.md
razo7: >They're surely not optional then

Yes, I was just looking for another point of view to assert my assumption for taking an action :)

README.md
```diff
@@ -96,7 +96,17 @@ FAR CRs are deleted by NHC after it sees the Node is healthy again.
 
 #### Example CR
 
-The FAR, `FenceAgentsRemediation`, CR is created by the admin and is used to trigger the fence-agent on a specific node. The CR includes an *agent* field for the fence-agent name, *sharedparameters* field with all the shared, not specific to a node, parameters, and *nodeparameters* field to specify the parameters for the fenced node.
+The FAR CR (CustomResource), created by the admin, is used to trigger the fence-agent on a specific node.
+
+The CR includes the following parameters:
+
+* `agent` - fence-agent name
+* `sharedparameters` - common cluster nodes information for executing the fence agent
```
slintes, line 104: I'm not 100% sure if I have the meaning of these CR fields right in my head... but from here I don't really get it 🤔 

IIUC:

- `sharedparameters` - cluster wide parameters for executing the fence agent
- `nodeparameters` - node specific parameters for executing the fence agent

WDYT? 
@clobrano @razo7

razo7: It is kind of similar.
I think `common cluster nodes information` is just saying in different words that it is `cluster wide parameters` without repeating the word `parameters` for explaining `sharedparameters`.
I don't have a strong preference, even though your suggestion might be just more accurate.

clobrano: I like @slintes version too, thanks!

PR #124, opened 2024-01-18
[WIP] Support off action

k-keiichi-rh:
This PR is to support off action.

The following is the FAR workflow with off action:
1. FAR adds NoExecute taint to the failed node
2. FAR powers off the failed node via the Fence Agent
3. FAR deletes workloads in the failed node
4. [User Intervention] Admins turn the failed node on after they check the failed node has been recovered.
5. After the failed node becomes healthy, NHC deletes FenceAgentsRemediation CR, the NoExecute taint in Step 2 is removed, and the node becomes schedulable again

In step 4, if users want to do troubleshooting on the failed node, they need to manually add the proper taints before turning on the failed node. The document pr is tracked in https://issues.redhat.com/browse/ECOPROJECT-1756.

[ECOPROJECT-1471](https://issues.redhat.com/browse/ECOPROJECT-1471)

### DIFF
--- a/controllers/fenceagentsremediation_controller_test.go
+++ b/controllers/fenceagentsremediation_controller_test.go
@@ -99,6 +99,47 @@ var _ = Describe("FAR Controller", func() {
 		})
 
 		Context("buildFenceAgentParams", func() {
+			Context("build fence agent params", func() {
+				baseShareParam := map[v1alpha1.ParameterName]string{
+					"--username": "admin",
+					"--password": "password",
+					"--ip":       "192.168.111.1",
+					"--lanplus":  "",
+				}
+				testCases := []struct {
+					name   string
+					action string
+					expect error
+				}{
+					{"reboot action", "reboot", nil},
+					{"off action", "off", nil},
+					{"unsupported action", "cycle", errors.New(errorUnsupportedAction)},
+				}
+
+				for _, tc := range testCases {
+					When(fmt.Sprintf("FAR includes %s", tc.name), func() {
+						It("should return expected result", func() {
+							shareParam := baseShareParam
+							shareParam["--action"] = tc.action
+							far := getFenceAgentsRemediation(workerNode, fenceAgentIPMI, shareParam, testNodeParam)
+							shareString, err := buildFenceAgentParams(far)
+							if tc.expect == nil {
+								Expect(err).NotTo(HaveOccurred())
+								Expect(shareString).To(ConsistOf([]string{
+									"--lanplus",
+									"--password=password",
+									"--username=admin",
+									fmt.Sprintf("--action=%s", tc.action),
+									"--ip=192.168.111.1",
+									"--ipport=6233"}))
+							} else {
+								Expect(err).To(HaveOccurred())
+								Expect(err).To(Equal(tc.expect))
+							}
+						})
+					})
+				}
+			})
 			When("FAR include different action than reboot", func() {
 				It("should succeed with a warning", func() {
 					invalidValTestFAR := getFenceAgentsRemediation(workerNode, fenceAgentIPMI, invalidShareParam, testNodeParam)
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -48,10 +48,12 @@ const (
 	// errors
 	errorMissingParams     = "nodeParameters or sharedParameters or both are missing, and they cannot be empty"
 	errorMissingNodeParams = "node parameter is required, and cannot be empty"
+	errorUnsupportedAction = "FAR doesn't support any other action than reboot and off"
 
-	SuccessFAResponse    = "Success: Rebooted"
-	parameterActionName  = "--action"
-	parameterActionValue = "reboot"
+	SuccessFAResponse          = "Success: Rebooted"
+	parameterActionName        = "--action"
+	parameterRebootActionValue = "reboot"
+	parameterOffActionValue    = "off"
 )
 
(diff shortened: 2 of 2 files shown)


mshitrit:
Hi @k-keiichi-rh , since this PR is still a WIP I've converted it to a "draft" PR.
We usually create our PRs as such in order to save cloud resources (draft PR doesn't run e2e tests automatically).

We try to follow this process:
- create draft PR
- run a single e2e test 
- review the PR
- once the test passes and the PR is lgtm we convert it to a none draft PR and run the rest of the e2e tests.

Let me know if that makes sense.

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -99,6 +99,47 @@ var _ = Describe("FAR Controller", func() {
 		})
 
 		Context("buildFenceAgentParams", func() {
+			Context("build fence agent params", func() {
```
slintes, line 102: this Context has the same description as the above Context, can we have a better description please?

slintes (changes requested):

controllers/fenceagentsremediation_controller_test.go
k-keiichi-rh: Thanks. I will change the comment.

kattz-kawa:
Hi everyone,

I am currently working on implementing end-to-end (E2E) tests in collaboration with @k-keiichi-rh. We are focusing on automating the process to power on the test nodes after executing the E2E tests. To achieve this,
I am considering two implementation approaches for this and would like to know if either of these would be unacceptable to you all to avoid miscommunication and potential rework. Could you please share your thoughts on these approaches?

**Approach 1:** Importing [aws-sdk-go-v2](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws) and creating a wrapper for the 'PowerOnNode' function.

 Pros:

1. Easier AWS operations: By adding a library, we can perform AWS operations more easily.

 Cons:

1. Managing library: Adding this library increases maintenance costs, which need to be managed. Compatibility issues may arise when libraries are updated.

1. Specific usage: Although this library has many functions, we are currently using it only for powering on/off in this E2E test. If we plan to add more features to this E2E test using this library in the future, it might be fine. However, if we only use it for powering on/off, the library might be more than we need.

**Approach 2:** Passing valid arguments to '[execCommandOnPod](https://github.com/medik8s/fence-agents-remediation/blob/d14361739a53e1b23fecdd42cf682b137b0129b3/test/e2e/utils/command.go#L100)' to execute 'fence_aws' or 'fence_ipmilan' within the fence-agents-remediation-controller-manager pod information.
 
 Pros:

1. Using existing code: This approach ensures future maintainability by reusing existing code.

 Cons:

1. Minor testing interference: This is not a big concern. There might be some interaction with the FAR container during testing.

I think adding the library is easy, but it might cause extra work in the future. Since I'm not the maintainer, I want to respect the maintainer's opinion. The second approach can be done within the existing framework, making it easier to maintain. I believe we should choose the option that will be most beneficial for FAR in the future. Your feedback would be greatly appreciated. Thank you.

razo7:
> Hi everyone,
> 
> I am currently working on implementing end-to-end (E2E) tests in collaboration with @k-keiichi-rh. We are focusing on automating the process to power on the test nodes after executing the E2E tests. To achieve this, I am considering two implementation approaches for this and would like to know if either of these would be unacceptable to you all to avoid miscommunication and potential rework. Could you please share your thoughts on these approaches?
> 
> **Approach 1:** Importing [aws-sdk-go-v2](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws) and creating a wrapper for the 'PowerOnNode' function.
> 
> Pros:
> 
> 1. Easier AWS operations: By adding a library, we can perform AWS operations more easily.
> 
> Cons:
> 
> 1. Managing library: Adding this library increases maintenance costs, which need to be managed. Compatibility issues may arise when libraries are updated.
> 2. Specific usage: Although this library has many functions, we are currently using it only for powering on/off in this E2E test. If we plan to add more features to this E2E test using this library in the future, it might be fine. However, if we only use it for powering on/off, the library might be more than we need.
> 
> **Approach 2:** Passing valid arguments to '[execCommandOnPod](https://github.com/medik8s/fence-agents-remediation/blob/d14361739a53e1b23fecdd42cf682b137b0129b3/test/e2e/utils/command.go#L100)' to execute 'fence_aws' or 'fence_ipmilan' within the fence-agents-remediation-controller-manager pod information.
> 
> Pros:
> 
> 1. Using existing code: This approach ensures future maintainability by reusing existing code.
> 
> Cons:
> 
> 1. Minor testing interference: This is not a big concern. There might be some interaction with the FAR container during testing.
> 
> I think adding the library is easy, but it might cause extra work in the future. Since I'm not the maintainer, I want to respect the maintainer's opinion. The second approach can be done within the existing framework, making it easier to maintain. I believe we should choose the option that will be most beneficial for FAR in the future. Your feedback would be greatly appreciated. Thank you.

Thanks for sharing your thoughts @kattz-kawa.
I think that approach 2 seems better for now, and approach 1 should be considered if there is anything that can improve the new E2E. It seems like approach 2 is easier to implement, which is preferable IMHO to speed up merging this important feature.

One note: please see (many) changes on the usage of FAR parameters from https://github.com/medik8s/fence-agents-remediation/pull/171  when you rebase the branch.

kattz-kawa:
Thank you for your feedback. @razo7 
I will proceed with Approach 2 and consider the FAR parameter changes from #171.

PR #125, opened 2024-01-24
Fix Remediation Events Occurrences and Names

razo7:
- Change _RemediationStarted_ event location and occourences
- Test _RemoveFinalizer_ event

Fix https://issues.redhat.com/browse/ECOPROJECT-1819

### DIFF
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -107,7 +107,6 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "Failed to get FenceAgentsRemediation CR")
 		return emptyResult, err
 	}
-	commonEvents.RemediationStarted(r.Recorder, far)
 
 	// At the end of each Reconcile we try to update CR's status
 	defer func() {
@@ -148,8 +147,9 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		if err := r.Client.Update(context.Background(), far); err != nil {
 			return emptyResult, fmt.Errorf("failed to add finalizer to the CR - %w", err)
 		}
-		r.Log.Info("Finalizer was added", "CR Name", req.Name)
+		commonEvents.RemediationStarted(r.Recorder, far)
 
+		r.Log.Info("Finalizer was added", "CR Name", req.Name)
 		utils.UpdateConditions(utils.RemediationStarted, far, r.Log)
 		commonEvents.NormalEvent(r.Recorder, far, utils.EventReasonAddFinalizer, utils.EventMessageAddFinalizer)
 		return requeueImmediately, nil
--- a/controllers/fenceagentsremediation_controller_test.go
+++ b/controllers/fenceagentsremediation_controller_test.go
@@ -206,7 +206,7 @@ var _ = Describe("FAR Controller", func() {
 					return controllerutil.ContainsFinalizer(underTestFAR, v1alpha1.FARFinalizer)
 				}, timeoutPreRemediation, pollInterval).Should(BeFalse(), "finalizer shouldn't be added")
 				verifyEvent(corev1.EventTypeWarning, utils.EventReasonCrNodeNotFound, utils.EventMessageCrNodeNotFound)
-				verifyNoEvent(corev1.EventTypeNormal, utils.EventReasonAddFinalizer, utils.EventMessageAddFinalizer)
+				verifyNoEvent(corev1.EventTypeNormal, utils.EventReasonRemediationStarted, utils.EventMessageRemediationStarted)
 
 				// If finalizer is missing, then a taint shouldn't exist
 				By("Not having remediation taint")
@@ -489,7 +489,7 @@ func verifyPreRemediationSucceed(underTestFAR *v1alpha1.FenceAgentsRemediation, 
 	By("Searching for finalizer ")
 	Expect(k8sClient.Get(context.Background(), client.ObjectKey{Name: nodeName, Namespace: namespace}, underTestFAR)).To(Succeed())
 	Expect(controllerutil.ContainsFinalizer(underTestFAR, v1alpha1.FARFinalizer)).To(BeTrue())
-	verifyEvent(corev1.EventTypeNormal, utils.EventReasonAddFinalizer, utils.EventMessageAddFinalizer)
+	verifyEvent(corev1.EventTypeNormal, utils.EventReasonRemediationStarted, utils.EventMessageRemediationStarted)
 
 	By("Searching for remediation taint if we have a finalizer")
 	Eventually(func(g Gomega) {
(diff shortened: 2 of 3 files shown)


controllers/fenceagentsremediation_controller_test.go
```diff
@@ -582,15 +582,20 @@ func cleanupFar() func(ctx context.Context, far *v1alpha1.FenceAgentsRemediation
...
 			deleteErr := k8sClient.Get(ctx, client.ObjectKeyFromObject(far), cr)
 			if apierrors.IsNotFound(deleteErr) {
 				// when trying to create far CR with invalid name
 				log.Info("Cleanup: Got error 404", "name", cr.Name)
+				isCRNotFound = true
 				return nil
 			}
 			return deleteErr
 		}, pollInterval, timeoutPostRemediation).Should(BeNil(), "CR should be deleted")
+		if !isCRNotFound {
```
clobrano, line 596: I won't block the PR on this, but I found a bit tricky to read the test if some checks are "hidden" in clean up functions

razo7: ok, it can be moved to `DeferCleanup`

controllers/fenceagentsremediation_controller.go
```diff
@@ -178,7 +177,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 			return emptyResult, fmt.Errorf("failed to remove finalizer from CR - %w", err)
 		}
 		r.Log.Info("Finalizer was removed", "CR Name", req.Name)
-		commonEvents.NormalEvent(r.Recorder, far, utils.EventReasonRemoveFinalizer, utils.EventMessageRemoveFinalizer)
+		commonEvents.NormalEvent(r.Recorder, far, utils.EventReasonRemoveRemediationCR, utils.EventMessageRemoveRemediationCR)
```
mshitrit, line 180: Any particular reason for this change ? (EventReasonRemoveFinalizer -> EventReasonRemoveRemediationCR)
I think `EventReasonRemoveFinalizer` makes sense in this context.
Since FAR does not actually triggers the CR deletion I'd expect this sort of event to be triggered by NHC.

Also I can think of edge cases were the finalizer is removed without CR deletion being triggered.

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -149,7 +149,14 @@ var _ = Describe("FAR Controller", func() {
 			DeferCleanup(k8sClient.Delete, context.Background(), node)
 
 			Expect(k8sClient.Create(context.Background(), underTestFAR)).To(Succeed())
-			DeferCleanup(cleanupFar(), context.Background(), underTestFAR)
+			DeferCleanup(func() {
```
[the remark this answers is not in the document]
razo7: > have one cleanup method that runs after each test instead of spreading the cleanup around

Sounds nice and I like to combine logic to simplify a process. ATM I like that for every resource we create we add a cleaning in the next line. Combining them into one place would "break" the `BeforeEach` and `JustBeforeEach` structure. Therefore, I am leaning toward keeping it as it is and maybe change it in the future,

>Also not sure node cleanup is needed, I vaguely remember some issue with that.

In the past, I had some issues when I tried to create a node/resource where the resource version was identical. This could be avoided when the node/resource is deleted and recreated again. Therefore, I think it is needed and some delay between tests is needed so there won't be any race (trying to create a resource which is currently deleting). This could also be avoided if we delete we zero grace period.

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -149,7 +149,14 @@ var _ = Describe("FAR Controller", func() {
...
+				Expect(cleanupFar(context.Background(), underTestFAR)).To(Succeed())
+				deleteErr := k8sClient.Get(ctx, client.ObjectKeyFromObject(underTestFAR), &v1alpha1.FenceAgentsRemediation{})
```
clobrano, line 154: I do not think that this is much different that putting the `verifyEvent` directly inside the cleanup function.
Is this necessary for a reason? Could we instead explicitly call the appropriate `verifyEvent` inside the test body?

controllers/fenceagentsremediation_controller.go
mshitrit: >as it could expect to see the AddFinalizer event as well

I agree, but I think there indeed should be an `AddFinalizer` event

controllers/fenceagentsremediation_controller_test.go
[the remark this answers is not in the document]
mshitrit: >Therefore, I am leaning toward keeping it as it is and maybe change it in the future

Fair enough , I consider this a NIT as it's IMO a matter of personal preference. 

> I vaguely remember some issue with that.

BTW I looked it up and I was confusing it with the NS deletion issue Marc has found,  so no issue here

> This could be avoided when the node/resource is deleted and recreated again

Hmm is there something preventing us from using the same node for multiple tests ?

controllers/fenceagentsremediation_controller_test.go
razo7: I am confused. From your [last comment](https://github.com/medik8s/fence-agents-remediation/pull/125#discussion_r1465009887) I understood that you prefer that I won't "hide" a check (`verifyEvent(...utils.EventMessageRemoveRemediationCR)`) in a function, so I took it out of from the function.
>I do not think that this is much different that putting the verifyEvent directly inside the cleanup function.

I think that this way is clearer for showing that we test this _verifyEvent_ after cleaning the CR.

>Is this necessary for a reason? Could we instead explicitly call the appropriate verifyEvent inside the test body?

Normally we don't verify that the CR was deleted using this event, but it looks like a good sign. We can create a test that verifies that if a CR was created and when it is deleted then the finalizer was deleted as well. Would it make sense to follow this logic or leave it as a "step" in _DeferCleanup_ or even as one line in _cleanupFar_ as I introduced in the beginning?

clobrano: > I am confused.

I am sorry, I should have been more clear. I was concerned by the call to `verifyEvent` not being in the Test body as other `verifyEvent` calls are. So having it in the cleanup function or called after the cleanup function, but still not in the Test body looks the same to me.

> We can create a test that verifies that if a CR was created

that's too much logic in a test (that should be simple to read). My idea was that inside each Test body you should know if the CR is created or not, so you can explicitly call `verifyEvent` or not, without any specific check. However, I was not sure whether this is easy to do or not, that's why I asked if there was a specific reason not to do so.

controllers/fenceagentsremediation_controller_test.go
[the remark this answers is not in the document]
razo7: >Hmm is there something preventing us from using the same node for multiple tests ?

Having new and independent resources doesn't connect unit tests that modify the resource (node). Simply recreating the node is easy, fast, and disconnect tests.

controllers/fenceagentsremediation_controller_test.go
clobrano: we can deal with this in a next PR, not urgent to fix it here

clobrano:
as per my comment [here](https://github.com/medik8s/fence-agents-remediation/pull/125#discussion_r1470725640)

/lgtm

PR #128, opened 2024-02-13
 Log error in remediation

mshitrit:
[ECOPROJECT-1505](https://issues.redhat.com//browse/ECOPROJECT-1505)
- retry in case of update conflict and log as info instead of an error
- Adding a "generate all" make job

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -434,3 +434,13 @@ ocp-aws-credentials: ## Add CredentialsRequest for OCP on AWS
 test-e2e: ginkgo ## Run end to end (E2E) tests
 	@test -n "${KUBECONFIG}" -o -r ${HOME}/.kube/config || (echo "Failed to find kubeconfig in ~/.kube/config or no KUBECONFIG set"; exit 1)
 	$(GINKGO) -r --keep-going --require-suite --vv  ./test/e2e -coverprofile cover.out
+
+# Revert all version or build date related changes
+.PHONY: bundle-reset
+bundle-reset:
+	VERSION=0.0.1 $(MAKE) manifests bundle
+	# empty creation date
+	sed -r -i "s|createdAt: .*|createdAt: \"\"|;" ${BUNDLE_CSV}
+
+.PHONY: full-gen
+full-gen:  go-verify manifests  generate manifests fmt bundle fix-imports bundle-reset ## generates all automatically generated content
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -167,9 +167,18 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		}
 
 		// remove node's taints
-		if err := utils.RemoveTaint(r.Client, far.Name); err != nil && !apiErrors.IsNotFound(err) {
-			return emptyResult, err
+		taint := utils.CreateRemediationTaint()
+		if err := utils.RemoveTaint(r.Client, far.Name, taint); err != nil {
+			if apiErrors.IsConflict(err) {
+				r.Log.Info("Failed to remove taint from node due to node update, retrying... ,", "node name", node.Name, "taint key", taint.Key, "taint effect", taint.Effect)
+				return ctrl.Result{RequeueAfter: time.Second}, nil
+
+			} else if !apiErrors.IsNotFound(err) {
+				r.Log.Error(err, "Failed to remove taint from node,", "node name", node.Name, "taint key", taint.Key, "taint effect", taint.Effect)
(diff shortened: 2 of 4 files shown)


controllers/fenceagentsremediation_controller.go
```diff
@@ -167,9 +167,18 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		}
 
 		// remove node's taints
-		if err := utils.RemoveTaint(r.Client, far.Name); err != nil && !apiErrors.IsNotFound(err) {
-			return emptyResult, err
+		taint := utils.CreateRemediationTaint()
+		if err := utils.RemoveTaint(r.Client, far.Name, taint); err != nil {
+			if apiErrors.IsConflict(err) {
+				r.Log.Info("Failed to remove taint from node due to node update, retrying... ,", "node name", node.Name, "taint key", taint.Key, "taint effect", taint.Effect)
+				return ctrl.Result{RequeueAfter: time.Second}, nil
```
razo7, line 174: Why requeue after one second and not immediately afterwards since this conflict should not happen multiple times, right?

Makefile
```diff
@@ -434,3 +434,13 @@ ocp-aws-credentials: ## Add CredentialsRequest for OCP on AWS
...
+
+# Revert all version or build date related changes
+.PHONY: bundle-reset
+bundle-reset:
+	VERSION=0.0.1 $(MAKE) manifests bundle
+	# empty creation date
+	sed -r -i "s|createdAt: .*|createdAt: \"\"|;" ${BUNDLE_CSV}
+
+.PHONY: full-gen
+full-gen:  go-verify manifests  generate manifests fmt bundle fix-imports bundle-reset ## generates all automatically generated content
```
razo7, line 446: NIT:  redundant space in `manifests  generate`

Makefile
```diff
@@ -434,3 +434,13 @@ ocp-aws-credentials: ## Add CredentialsRequest for OCP on AWS
 test-e2e: ginkgo ## Run end to end (E2E) tests
 	@test -n "${KUBECONFIG}" -o -r ${HOME}/.kube/config || (echo "Failed to find kubeconfig in ~/.kube/config or no KUBECONFIG set"; exit 1)
 	$(GINKGO) -r --keep-going --require-suite --vv  ./test/e2e -coverprofile cover.out
+
+# Revert all version or build date related changes
+.PHONY: bundle-reset
+bundle-reset:
+	VERSION=0.0.1 $(MAKE) manifests bundle
+	# empty creation date
+	sed -r -i "s|createdAt: .*|createdAt: \"\"|;" ${BUNDLE_CSV}
```
razo7, line 443: Why is it needed? Running bundle won't initialize the `createdAt` field?

Makefile
razo7, line 446: When this target is useful? Running `make test` would do the same and also test the unit tests :)

controllers/fenceagentsremediation_controller.go
mshitrit: Giving the update a bit of time to finish, no point in stressing the system and spamming the log

Makefile
mshitrit: - test doesn't generate the bundle
- sometimes I wan't to just make sure my files are up to speed without running tests

Makefile
mshitrit: but running bundle-update would.
so this can be used to make sure we have a "clean slate".

razo7: bundle-update is called only when building the bundle container (`bundle-build`) or setting him up for community (`bundle-community`). For both of the places, it makes sense to "update" the bundle.
>sometimes I wan't to just make sure my files are up to speed without running tests

Considering your comment below, I see your motivation for having this target

controllers/fenceagentsremediation_controller.go
mshitrit: Well, in case we want to utilize the exponential back-off mechanism we must return an error which will also be logged as such.
IIUC it is something we want to avoid (logging an error) since this behavior is not uncommon and expected .

slintes: `Requeue: True` also results in rate limited requeue
https://github.com/kubernetes-sigs/controller-runtime/blob/56159419231e985c091ef3e7a8a3dee40ddf1d73/pkg/internal/controller/controller.go#L341

razo7: >Requeue: True also results in rate limited requeue

I agree this is what I originally meant. But IIRC Michael suspects that this error is recurrent (I didn't see a proof for that), it seems more like a temporary error to me.

PR #130, opened 2024-02-21
Remove catalog build and reference in release note

clobrano:
- Remove catalog build from Makefile
- Remove catalog's reference from the release note


clobrano:
Change plan. I'll update the target to create file-based catalogs instead

PR #132, opened 2024-03-12
Set Golang version in Dockerfile

razo7:
#### Why we need this PR

Fix the way we set the Golang version in the Dockerfile. 
See the below error from [post-submit](https://github.com/medik8s/fence-agents-remediation/blob/main/.github/workflows/post-submit.yaml) job that blocks the container build, of the pushed code (after merging PR or setting new tag), and eventually pushing it to Quay.

> .
> .
> .
> RUN export GO_VERSION=$(grep -E "go [[:digit:]]\.[[:digit:]][[:digit:]]" go.mod | awk ...
> go: downloading golang.org/dl v0.0.0-20240305174203-2885f567d808
> go: golang.org/dl/go1.21@latest: module golang.org/dl@latest found (v0.0.0-20240305174203-2885f567d808), but does not contain package golang.org/dl/go1.21

It tries to use Golang v1.21 ([latest installed Golang version](https://github.com/medik8s/fence-agents-remediation/blob/main/Dockerfile#L3)) for installing Golang v1.20.

#### Changes made

Directly install a desired Go version instead of first installing what the builder image provides, and then using Golang CLI to get the desired version. It stems from an error we have seen in building the contianer at the [post-submit](https://github.com/medik8s/fence-agents-remediation/blob/main/.github/workflows/post-submit.yaml) job

No issues, but an error from post-submit job

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Build the manager binary
 FROM quay.io/centos/centos:stream8 AS builder
-RUN dnf install -y golang jq git \
+RUN dnf install -y jq git \
     && dnf clean all -y
 
 WORKDIR /workspace
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,28 @@
 # Build the manager binary
 FROM quay.io/centos/centos:stream8 AS builder
-RUN dnf install -y golang git \
+RUN dnf install -y golang jq git \
     && dnf clean all -y
 
 WORKDIR /workspace
-# Copy the Go Modules manifests
+# Copy the Go Modules manifests for detecting Go version
 COPY go.mod go.mod
 COPY go.sum go.sum
 
-# Ensure correct Go version
-RUN export GO_VERSION=$(grep -E "go [[:digit:]]\.[[:digit:]][[:digit:]]" go.mod | awk '{print $2}') && \
-    go install golang.org/dl/go${GO_VERSION}@latest && \
-    ~/go/bin/go${GO_VERSION} download && \
-    /bin/cp -f ~/go/bin/go${GO_VERSION} /usr/bin/go && \
-    go version
+RUN \
+    # get Go version from mod file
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,19 +1,28 @@
 # Build the manager binary
 FROM quay.io/centos/centos:stream8 AS builder
-RUN dnf install -y golang git \
+RUN dnf install -y golang jq git \
```
slintes, line 3: remove golang

Makefile
```diff
@@ -216,7 +216,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
 .PHONY: docker-build
-docker-build: test-no-verify ## Build docker image with the manager.
+docker-build: ## Build docker image with the manager.
```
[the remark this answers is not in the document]
razo7: No, a mistake

Dockerfile
```diff
@@ -1,19 +1,15 @@
...
 
-# Ensure correct Go version
-RUN export GO_VERSION=$(grep -E "go [[:digit:]]\.[[:digit:]][[:digit:]]" go.mod | awk '{print $2}') && \
-    go install golang.org/dl/go${GO_VERSION}@latest && \
-    ~/go/bin/go${GO_VERSION} download && \
-    /bin/cp -f ~/go/bin/go${GO_VERSION} /usr/bin/go && \
-    go version
+# Install Golang based on the version from go.mod
+RUN GO_VERSION=$(grep -E "go [[:digit:]]\.[[:digit:]][[:digit:]]" go.mod | awk '{print $2}') \
+    && dnf install golang-1.21.7 git -y && go version \
```
beekhof, line 11: Why extract GO_VERSION if we always install `1.21.7`?

razo7: that was for testing :(

razo7: The idea of using the patch version in go.mod isn't supported in Golang v1.20 and below, so I will introduce it in #133

slintes (approved):

PR #133, opened 2024-03-15
Build container using x.y.z golang version and updating Golang version

razo7:
#### Why we need this PR
Simplify how the Dockerfile fetches the Golang version for building the binary.

#### Changes made

- Fetch the Golang version from the go directive in go.mod, and download it using the repos of the base image (CentOS stream 8) instead of using the official website to download and extract the newer Golang version
- Upgrade Golang 1.20-> 1.21.7 and use the patch version in go.mod

Follow up to #132, and should not be merged beforehand.


razo7:
There is a tradeoff for using a patch version (here and eventually aligning it in other places) for Golang builder, and the pros won't overcome the cons. It might be more relevant in the future, but not for now.

PR #135, opened 2024-03-19
Add support to multiple templates with the same kind

mshitrit:
#### Why we need this PR
This PR will enable the user to select several FAR Templates (of the same kind) when using NHC escalating remediation.
In a nutshell Far can be used several times as a remediator each time with a different fence agent.
This feature is dependent on the changes in NHC as well (which will be available from NHC 0.8.0).

#### Changes made
- Added a "Defaulting" (Mutating) Webhook to Far's Template which automaticlly adds to the template the "Support Same Kind" annotation which NHC detects
- Fetching node name from "Node Name" annotation in case it exist instead of always assuming that the Node's name is the CR's name
- Upgraded common version in order to use both mentioned annotations from common

[ECOPROJECT-1918](https://issues.redhat.com//browse/ECOPROJECT-1918)

### DIFF
--- a/api/v1alpha1/fenceagentsremediationtemplate_webhook.go
+++ b/api/v1alpha1/fenceagentsremediationtemplate_webhook.go
@@ -49,7 +49,7 @@ func (farTemplate *FenceAgentsRemediationTemplate) Default() {
 	if farTemplate.GetAnnotations() == nil {
 		farTemplate.Annotations = make(map[string]string)
 	}
-	if _, isSameKindSupported := farTemplate.GetAnnotations()[commonAnnotations.MultipleTemplatesSupportedAnnotation]; !isSameKindSupported {
+	if _, isSameKindAnnotationSet := farTemplate.GetAnnotations()[commonAnnotations.MultipleTemplatesSupportedAnnotation]; !isSameKindAnnotationSet {
 		farTemplate.Annotations[commonAnnotations.MultipleTemplatesSupportedAnnotation] = "true"
 	}
 }
--- a/api/v1alpha1/fenceagentsremediationtemplate_webhook.go
+++ b/api/v1alpha1/fenceagentsremediationtemplate_webhook.go
@@ -17,6 +17,8 @@ limitations under the License.
 package v1alpha1
 
 import (
+	commonAnnotations "github.com/medik8s/common/pkg/annotations"
+
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -37,8 +39,24 @@ func (r *FenceAgentsRemediationTemplate) SetupWebhookWithManager(mgr ctrl.Manage
 
 // TODO(user): EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 
+// +kubebuilder:webhook:path=/mutate-fence-agents-remediation-medik8s-io-v1alpha1-fenceagentsremediationtemplate,mutating=true,failurePolicy=fail,sideEffects=None,groups=fence-agents-remediation.medik8s.io,resources=fenceagentsremediationtemplates,verbs=create;update,versions=v1alpha1,name=mfenceagentsremediationtemplate.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Defaulter = &FenceAgentsRemediationTemplate{}
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+func (farTemplate *FenceAgentsRemediationTemplate) Default() {
+	webhookFARTemplateLog.Info("default", "name", farTemplate.Name)
+	if farTemplate.GetAnnotations() == nil {
+		farTemplate.Annotations = make(map[string]string)
+	}
+	if _, isSameKindSupported := farTemplate.GetAnnotations()[commonAnnotations.MultipleTemplatesSupportedAnnotation]; !isSameKindSupported {
+		farTemplate.Annotations[commonAnnotations.MultipleTemplatesSupportedAnnotation] = "true"
+	}
+}
+
 // TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
 // +kubebuilder:webhook:path=/validate-fence-agents-remediation-medik8s-io-v1alpha1-fenceagentsremediationtemplate,mutating=false,failurePolicy=fail,sideEffects=None,groups=fence-agents-remediation.medik8s.io,resources=fenceagentsremediationtemplates,verbs=create;update,versions=v1alpha1,name=vfenceagentsremediationtemplate.kb.io,admissionReviewVersions=v1
+
 var _ webhook.Validator = &FenceAgentsRemediationTemplate{}
 
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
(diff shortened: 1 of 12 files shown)


api/v1alpha1/fenceagentsremediationtemplate_webhook.go
```diff
@@ -37,8 +39,24 @@ func (r *FenceAgentsRemediationTemplate) SetupWebhookWithManager(mgr ctrl.Manage
 
 // TODO(user): EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 
+// +kubebuilder:webhook:path=/mutate-fence-agents-remediation-medik8s-io-v1alpha1-fenceagentsremediationtemplate,mutating=true,failurePolicy=fail,sideEffects=None,groups=fence-agents-remediation.medik8s.io,resources=fenceagentsremediationtemplates,verbs=create;update,versions=v1alpha1,name=mfenceagentsremediationtemplate.kb.io,admissionReviewVersions=v1
```
razo7, line 42: Does the following check happen on creation and update or only for creation? I am asking due to the verbs that are used in this line `verbs=create;update`

controllers/fenceagentsremediation_controller.go
```diff
@@ -287,6 +287,16 @@ func (r *FenceAgentsRemediationReconciler) updateStatus(ctx context.Context, far
 	return nil
 }
 
+func getNodeName(far *v1alpha1.FenceAgentsRemediation) string {
+	if far.GetAnnotations() == nil {
+		return far.GetName()
+	} else if nodeName, isNodeNameAnnotationExist := far.GetAnnotations()[commonAnnotations.NodeNameAnnotation]; isNodeNameAnnotationExist {
```
razo7, line 293: NIT: no need for `else` block after we return a value. Not sure which one way is more readable, but the else is redundant

controllers/fenceagentsremediation_controller.go
```diff
@@ -287,6 +287,16 @@ func (r *FenceAgentsRemediationReconciler) updateStatus(ctx context.Context, far
 	return nil
 }
 
+func getNodeName(far *v1alpha1.FenceAgentsRemediation) string {
```
razo7, line 290: Can you add a function declaration that will include we take the node name from the CR name or the annotation?

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
mshitrit: IIUC both.
I've done a test trying to remove the annotation and the update webhook triggering kept adding it.

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -163,12 +163,8 @@ var _ = Describe("FAR Controller", func() {
 		})
 
 		When("creating valid FAR CR", func() {
-			BeforeEach(func() {
-				node = utils.GetNode("", workerNode)
-				underTestFAR = getFenceAgentsRemediation(workerNode, fenceAgentIPMI, testShareParam, testNodeParam)
-			})
 
-			It("should have finalizer, taint, while the two VAs and one pod will be deleted", func() {
+			testFunc := func() {
```
razo7, line 167: Can we use a more meaningful variable name? e.g., `testValidCR`

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -194,7 +190,22 @@ var _ = Describe("FAR Controller", func() {
...
 			})
+
+			It("should have finalizer, taint, while the two VAs and one pod will be deleted", testFunc)
+
+			When("remediation is created from escalation remediation supporting same kind template", func() {
+				BeforeEach(func() {
+					underTestFAR.Name = fmt.Sprintf("%s-%s", workerNode, "pseudo-random-test-sufix")
+					underTestFAR.Annotations = map[string]string{"remediation.medik8s.io/node-name": workerNode}
+				})
+				It("should have finalizer, taint, while the two VAs and one pod will be deleted", testFunc)
```
razo7, line 206: It might be better to scaffold the test structure differently. I find it more readable to have a few contexts under `Context("Reconcile"` with `Context("creating valid FAR CR",...)` and maybe Context("creating invalid FAR CR",...)

Moving from 

	Context("Reconcile",..)->
		When("creating valid FAR CR",...)
			It(...)
			When("remediation is created from escalation ...", ...)
				It(...)
.
.
.

To maybe 

	Context("Reconcile",..)->
		Context("creating valid FAR CR",...)
			When("remediation isn't created from escalation ...", ...)
				It(...)
			When("remediation is created from escalation ...", ...)
				It(...)
 .
.
.

controllers/fenceagentsremediation_controller_test.go
mshitrit: No objection for a better name, but what this method tests is very generic (i.e fence command, pod deletion, conditions, events etc...) so I think a generic name is appropriate here.

I don't think `testValidCR` appropriately reflects what this method do, but I don't have a better name for a generic name,

controllers/fenceagentsremediation_controller_test.go
mshitrit: I've initially constructed it so in order to minimize the diff in order to make the review easier, but I don't mind doing some shuffling.

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
```diff
@@ -37,8 +39,24 @@ func (r *FenceAgentsRemediationTemplate) SetupWebhookWithManager(mgr ctrl.Manage
...
+
+// Default implements webhook.Defaulter so a webhook will be registered for the type
+func (farTemplate *FenceAgentsRemediationTemplate) Default() {
+	webhookFARTemplateLog.Info("default", "name", farTemplate.Name)
+	if farTemplate.GetAnnotations() == nil {
+		farTemplate.Annotations = make(map[string]string)
+	}
+	if _, isSameKindSupported := farTemplate.GetAnnotations()[commonAnnotations.MultipleTemplatesSupportedAnnotation]; !isSameKindSupported {
+		farTemplate.Annotations[commonAnnotations.MultipleTemplatesSupportedAnnotation] = "true"
+	}
```
clobrano, line 54: It might just be me, but reading the condition `!isSameKindSupported` makes me think that the condition is _not
supported_, instead than just the annotation unset. It might longer, but what do you think about
`isSameKindAnnotationSet` instead?

controllers/fenceagentsremediation_controller.go
```diff
@@ -120,7 +120,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 
 	// Validate FAR CR name to match a nodeName from the cluster
 	r.Log.Info("Check FAR CR's name")
-	node, err := utils.GetNodeWithName(r.Client, req.Name)
+	node, err := utils.GetNodeWithName(r.Client, getNodeName(far))
```
clobrano, line 123: I can't add the comment in the exact line, so writing here instead.
At line 129 the error log message should be updated to

from  "Didn't find a node matching the CR's name" (CR name is not expected to match node name anymore)
to something like "Could not find CR's target node"

clobrano (changes requested):

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
mshitrit: Good point, I agree :+1:

controllers/fenceagentsremediation_controller_test.go
razo7: How about `testSuccessfulRemediation`? Since it includes a few steps that are happening in a Successful Remediation by FAR

razo7:
>/test 4.14-openshift-e2e

 I would recommend testing on 4.15. since it is the latest version we have in CI, and it was already released :)
`/test 4.15-openshift-e2e`

clobrano (approved):

PR #136, opened 2024-03-27
Manual workflow for tagged versions

clobrano:
#### Why we need this PR

FAR CI uses the post-release workflow for building and pushing versioned bundle images, however, the workflow cannot guess the right PREVIOUS_VERSION (when it is necessary) and it would publish manifests with wrong value of the `.spec.replaces` field.

#### Changes made

- Move build and push images job from the `post-release` to the `release` workflow, which supports manual inputs such has PREVIOUS_VERSION
- Integrated the `release` workflow with the latest version of Medik8s shared workflow to build and push community bundles
- Set `spec.replaces` field only in OKD bundle

#### Test plan
Tested locally with nektos/act

### DIFF
--- a/.github/workflows/post-submit.yaml
+++ b/.github/workflows/post-submit.yaml
@@ -3,22 +3,25 @@ on:
   push:
     branches:
       - main
-    tags:
-      - 'v*'
 
 permissions:
   contents: write
   pull-requests: read
 
 jobs:
-  push-images:
-    name: Build and push images to quay.io/medik8s
+  push_to_registry:
+    name: Build and push unversioned images to quay.io/medik8s
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: go.mod
 
       - name: Log in to Quay.io
         uses: docker/login-action@v2
@@ -27,14 +30,8 @@ jobs:
           password: ${{ secrets.QUAY_PASSWORD }}
           registry: quay.io
 
-      - name: Build and push CSV 0.0.1 + latest images for PR merges to main
-        if: ${{ github.ref_type != 'tag' }}
+      - name: Build and push CSV version v0.0.1 with latest images
         run: export IMAGE_REGISTRY=quay.io/medik8s && make container-build-and-push-community
-
-      - name: Build and push versioned CSV and images for tags
-        if: ${{ github.ref_type == 'tag' }}
-        # remove leading 'v' from tag!
-        run: export VERSION=$(echo $GITHUB_REF_NAME | sed 's/v//') && make container-build-and-push-community
 
       - name: Create release with manifests
         if: ${{ github.ref_type == 'tag' }}
--- a/.github/workflows/post-submit.yaml
+++ b/.github/workflows/post-submit.yaml
@@ -14,17 +14,17 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
       - name: Set up Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
 
       - name: Log in to Quay.io
-        uses: docker/login-action@v2
+        uses: docker/login-action@v3
         with:
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,40 +1,110 @@
-name: Release bundles
+name: Release
 on:
   workflow_dispatch:
-    inputs:
-        version:
-          description: "The version to release, without the leading `v`"
-          required: true
-          type: string
-        previous_version:
-          description: "The previous version, used for the CVS's `replaces` field, without the leading `v`"
-          required: true
-          type: string
-        community:
-          description: "The community to release the bundle to"
-          required: true
-          type: choice
-          default: "K8S"
-          options:
-            - K8S
-            - OKD
+   inputs:
+     operation:
+       description: "The operation to perform."
+       required: true
+       type: choice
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -253,19 +253,24 @@ export ICON_BASE64 ?= ${DEFAULT_ICON_BASE64}
...
-verify-previous-version: ## Verifies that PREVIOUS_VERSION variable is set
-	@if [ $(VERSION) != $(DEFAULT_VERSION) ] && [ $(PREVIOUS_VERSION) = $(DEFAULT_VERSION) ]; then \
-  			echo "Error: PREVIOUS_VERSION must be set for the selected VERSION"; \
-    		exit 1; \
-    fi
+.PHONY: add-replaces-field
+add-replaces-field: ## Add replaces field to the CSV
+	# add replaces field when building versioned bundle
+	@if [ $(VERSION) != $(DEFAULT_VERSION) ]; then \
+		if [ $(PREVIOUS_VERSION) == $(DEFAULT_VERSION) ]; then \
```
razo7, line 266: Can we also check whether `VERSION != PREVIOUS_VERSION`, since the following values are valid according to this if, but they don't make sense for _VERSION_ being equal to _PREVIOUS_VERSION_.

DEFAULT_VERSION = 0.0.1
VERSION = 0.3.0
PREVIOUS_VERSION= 0.3.0

bundle/manifests/fence-agents-remediation.clusterserviceversion.yaml
```diff
@@ -159,7 +159,7 @@ spec:
     - Speed - FAR is rapid since it can reboot a node and receive an acknowledgment from the API call while other remediators might need to wait a safe time till they can expect the node to be rebooted.
     - Diversity - FAR includes several fence agents from a large known set of upstream fencing agents for bare metal servers, virtual machines, cloud platforms, etc.
     - Adjustability - FAR allows to set up different parameters for running the API call that remediates the node.
-  displayName: Fence Agents Remediation Operator
+  displayName: Fence Agents Remediation Operator - Community Edition
```
razo7, line 162: Is this intended? I would imagine that we want to append `- Community Edition` only when building the bundle for releasing it to the community.

.github/workflows/post-submit.yaml
```diff
@@ -3,39 +3,36 @@ on:
...
       - name: Checkout code
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
 
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: go.mod
+
```
razo7, line 25: Why is it needed now?

clobrano: I think it was always needed, just not used (and so potentially not using go version set in go.mod). I'll update to version 5

.github/workflows/release.yml
```diff
@@ -1,40 +1,110 @@
...
 jobs:
-  make_k8s_community_bundle:
-    if: inputs.community == 'K8S'
+  push_to_registry:
+    if: ${{ inputs.operation == 'build_and_push_images' }}
+    name: Build and push images to quay.io/medik8s
+    runs-on: ubuntu-22.04
+    env:
+      VERSION: ${{ inputs.version }}
+      #PREVIOUS_VERSION: ${{ inputs.previous_version }}
```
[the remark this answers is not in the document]
clobrano: to be uncommented :+1:

bundle/manifests/fence-agents-remediation.clusterserviceversion.yaml
clobrano: I think I run too fast here. The idea is to have the k8s bundle committed upstream, _however_, this will appear also downstream since FAR does not have (yet) the OCP bundle built from scratch downstream.
So, for now I need to revert this change, but in the future it will be right :)

razo7: >The idea is to have the k8s bundle committed upstream, however, this will appear also downstream since FAR does not have (yet) the OCP bundle built from scratch downstream.

I wasn't aware of that idea. I understand why it is needed for NHC, but not sure we have discussed it widely for the rest of the operators. It might be an improvement, not sure.

.github/workflows/post-submit.yaml
razo7: >I think it was always needed, 

Can you elaborate on why? afterward, we merely build (and push) the containers. Do we need Golang for that?

bundle/manifests/fence-agents-remediation.clusterserviceversion.yaml
clobrano: Why should be good for one operator and not for the others?

.github/workflows/release.yml
```diff
@@ -1,40 +1,110 @@
...
+
+      - name: Log in to Quay.io
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
+          registry: quay.io
+
+      - name: Build and push versioned CSV and images
+        run: VERSION=$VERSION PREVIOUS_VERSION=$PREVIOUS_VERSION SKIP_RANGE_LOWER=$SKIP_RANGE_LOWER make container-build-k8s container-push
```
clobrano, line 65: oops, wrong makefile target here. Need to fix this

.github/workflows/post-submit.yaml
clobrano: As part of container build target we also perform tests

bundle/manifests/fence-agents-remediation.clusterserviceversion.yaml
razo7: IIUC the idea for using that for NHC was due to three different bundle creation options: For k8s, for metrics, and for OCP. This scenario is unique since only NHC has a metrics bundle variant. Unless I am wrong here and it is planned to be created for each Medik8s operator.

clobrano: My understanding was to have the "pure" k8s bundle upstream regardless the other options, but I might be wrong and in any case downstream CI isn't ready for FAR to support it anyway :)

clobrano:
E2E tests do not test the changes of this PR. Can merge without them.

PR #142, opened 2024-04-10
Check stderr when running agent

razo7:
#### Why we need this PR

Considering a successful command execution when the return value is zero, no error, and stderr is empty as well.

#### Changes made

- Include a check for stderr being empty when considering a successful agent execution.

Fixes (somehow) [ECOPROJECT-1938](https://issues.redhat.com//browse/ECOPROJECT-1938)

### DIFF
--- a/pkg/cli/cliexecuter.go
+++ b/pkg/cli/cliexecuter.go
@@ -127,7 +127,7 @@ func (e *Executer) runWithRetry(ctx context.Context, uid types.UID, command []st
 			ctxWithTimeout, cancel := context.WithTimeout(ctx, timeout)
 			defer cancel()
 			stdout, stderr, faErr = e.runner(ctxWithTimeout, command)
-			if faErr == nil {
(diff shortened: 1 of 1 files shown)


pkg/cli/cliexecuter.go
```diff
@@ -127,7 +127,7 @@ func (e *Executer) runWithRetry(ctx context.Context, uid types.UID, command []st
 			ctxWithTimeout, cancel := context.WithTimeout(ctx, timeout)
 			defer cancel()
 			stdout, stderr, faErr = e.runner(ctxWithTimeout, command)
-			if faErr == nil {
+			if faErr == nil && stderr == "" {
```
slintes, line 130: stderr stream isn't an indication for an actual error!
e.g. curl uses stderr for verbose info, and stdout for the actual result

/hold

slintes (changes requested):

PR #144, opened 2024-04-17
Remove redundant VoulmeAttachment from testing

razo7:
#### Why we need this PR

VoulmeAttachment is no longer removed from the resource deletion strategy, thus it is no longer needed for testing. Follow up to #102 .

#### Changes made

Remove VoulmeAttachment creation and deletion in unit-test.


clobrano (approved):

razo7:
**_tide_** job is not responding, so I am pushing old commit as a new one to retrigger tide (with the many other tests)...

slintes:
/hold

need to fix CI config for having e2e tests required

PR #145, opened 2024-05-09
Update docs with the new remediationStrategy spec

clobrano:
#### Why we need this PR

FAR CR has a new spec field `remediationStrategy` (see #92)

#### Changes made

Add the documentation about the new FAR remediation's spec field `remediationStrategy`

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,10 +152,18 @@ metadata:
   name: fenceagentsremediationtemplate-default
   namespace: default
 spec:
-  template: {}
+  template:
+    spec:
+        remediationStrategy: <remediation_strategy>
 ```
 
 > *Note*: FenceAgentsRemediationTemplate CR must be created in the same namespace that the FAR operator has been installed.
+ 
+The `.spec.template.spec.remediation_strategy` field can either be `ResourceDeletion` or `OutOfServiceTaint`:
+
+- `ResourceDeletion`: This remediation strategy removes the pods on the node, rather than the removal of the node object. This strategy recovers workloads faster.
+- `OutOfServiceTaint`: This remediation strategy implicitly causes the removal of the pods and associated volume attachments on the node, rather than the removal of the node object. It achieves this by placing the `OutOfServiceTaint` taint on the node. The `OutOfServiceTaint` strategy also represents a non-graceful node shutdown. A non-graceful node shutdown occurs when a node is shutdown and not detected, instead of triggering an in-operating system shutdown.
+
 
 Configuring NodeHealthCheck to use the example `fenceagentsremediationtemplate-default` template above.
 
--- a/README.md
+++ b/README.md
@@ -152,18 +152,10 @@ metadata:
   name: fenceagentsremediationtemplate-default
   namespace: default
 spec:
-  template:
-    spec:
-        remediationStrategy: <remediation_strategy>
+  template: {}
 ```
 
 > *Note*: FenceAgentsRemediationTemplate CR must be created in the same namespace that the FAR operator has been installed.
- 
-The `.spec.template.spec.remediation_strategy` field can either be `ResourceDeletion` or `OutOfServiceTaint`:
-
-- `ResourceDeletion`: This remediation strategy removes the pods on the node, rather than the removal of the node object. This strategy recovers workloads faster.
-- `OutOfServiceTaint`: This remediation strategy implicitly causes the removal of the pods and associated volume attachments on the node, rather than the removal of the node object. It achieves this by placing the `OutOfServiceTaint` taint on the node. The `OutOfServiceTaint` strategy also represents a non-graceful node shutdown. A non-graceful node shutdown occurs when a node is shutdown and not detected, instead of triggering an in-operating system shutdown.
-
 
 Configuring NodeHealthCheck to use the example `fenceagentsremediationtemplate-default` template above.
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -152,10 +152,18 @@ metadata:
...
 spec:
-  template: {}
+  template:
+    spec:
+        remediationStrategy: <remediation_strategy>
 ```
 
 > *Note*: FenceAgentsRemediationTemplate CR must be created in the same namespace that the FAR operator has been installed.
+ 
+The `.spec.template.spec.remediation_strategy` field can either be `ResourceDeletion` or `OutOfServiceTaint`:
```
razo7, line 162: It can be the `.spec.template.spec.remediation_strategy` field of fartemplate CR or simply `.spec.remediation_strategy` of FAR CR. I prefer the latter and modify the README in the [FAR CR fields](https://github.com/medik8s/fence-agents-remediation/pull/145/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R197-R202) rather than the template spec of fartemplate

README.md
```diff
@@ -152,10 +152,18 @@ metadata:
...
+
+- `ResourceDeletion`: This remediation strategy removes the pods on the node, rather than the removal of the node object. This strategy recovers workloads faster.
```
razo7, line 164: Not sure if the rest is needed.
```suggestion
- `ResourceDeletion`: This remediation strategy (forcefully) removes the pods on the node.
```

razo7 (changes requested):
Please link to the PR which introduced the new field and mention that it is not **only** modifying README, as it should also update the CR sample as in https://github.com/medik8s/fence-agents-remediation/pull/122

README.md
```diff
@@ -152,10 +152,18 @@ metadata:
...
+    spec:
+        remediationStrategy: <remediation_strategy>
 ```
 
 > *Note*: FenceAgentsRemediationTemplate CR must be created in the same namespace that the FAR operator has been installed.
+ 
+The `.spec.remediation_strategy` field defines the _remediationStrategy_ that FAR supports and will be used for recovering workloads faster during node reboot in a [non-graceful](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#non-graceful-node-shutdown) manner. There are only two possible remediations - `ResourceDeletion` or `OutOfServiceTaint`:
+
+- `ResourceDeletion`: This remediation strategy (forcefully) removes the pods on the node.
+- `OutOfServiceTaint`: This remediation strategy implicitly causes the removal of the pods and associated volume attachments on the node, rather than the removal of the node object. It achieves this by placing the `OutOfServiceTaint` taint on the node. The `OutOfServiceTaint` strategy also represents a non-graceful node shutdown. A non-graceful node shutdown occurs when a node is shutdown and not detected, instead of triggering an in-operating system shutdown.
```
razo7, line 165: >The `OutOfServiceTaint` strategy also represents a non-graceful node shutdown. A non-graceful node shutdown occurs when a node is shutdown and not detected, instead of triggering an in-operating system shutdown.

I am not sure the above sentence is needed as it was already mentioned in line 162.


```suggestion
- `OutOfServiceTaint`: This remediation strategy implicitly causes the removal of the pods and associated volume attachments on the node. It achieves this by placing the [`OutOfServiceTaint` taint](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-kubernetes-io-out-of-service) on the node.
```

bundle/manifests/fence-agents-remediation.clusterserviceversion.yaml
```diff
@@ -11,6 +11,7 @@ metadata:
             "name": "worker-1"
           },
           "spec": {
+            "remediation_strategy": "OutOfServiceTaint",
```
razo7, line 14: Please also commit the changes under [`config/samples`](https://github.com/medik8s/fence-agents-remediation/blob/main/config/samples/fence-agents-remediation_v1alpha1_fenceagentsremediation.yaml#L5)

README.md
```diff
@@ -152,10 +152,18 @@ metadata:
...
 spec:
-  template: {}
+  template:
+    spec:
+        remediationStrategy: <remediation_strategy>
 ```
 
 > *Note*: FenceAgentsRemediationTemplate CR must be created in the same namespace that the FAR operator has been installed.
+ 
+The `.spec.remediation_strategy` field defines the _remediationStrategy_ that FAR supports and will be used for recovering workloads faster during node reboot in a [non-graceful](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#non-graceful-node-shutdown) manner. There are only two possible remediations - `ResourceDeletion` or `OutOfServiceTaint`:
```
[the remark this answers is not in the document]
clobrano: You're right, not sure what happened. I'll fix it

bundle/manifests/fence-agents-remediation.clusterserviceversion.yaml
slintes: the field name is `remediationStrategy` not `remediation_strategy`

razo7: how this file was created? by `make bundle`? 
CI should catch this kind of error if the bundle target hasn't been used

README.md
```diff
@@ -220,6 +223,7 @@ spec:
       worker-0: "6233"
       worker-1: "6234"
       worker-2: "6235"
+  remediationStrategy: OutOfServiceTaint
```
razo7, line 226: Although `OutOfServiceTaint` is preferable when it is enabled, but by default the `remediationStrategy` will be `ResourceDeletion` (and it is by default because due to issue of support to the _OutOfService_ taint)

bundle/manifests/fence-agents-remediation.clusterserviceversion.yaml
clobrano: PEBKAC, sorry. I forgot the link between the `config/sample` and what goes into `alm-examples`.

clobrano:
```
 level=fatal msg=error destroying bootstrap resources failed during the destroy bootstrap hook: failed to remove bootstrap SSH rule: bootstrap ssh rule was not removed within 5m0s: timed out waiting for the condition 
 ```
 
 The error doesn't seem our fault. Retrying
 
/test 4.16-openshift-e2e

File: README.md [hunk 7]
```diff
@@ -192,6 +192,9 @@ The CR includes the following parameters:
 * `retrycount` - number of times to retry the fence agent in case of failure. The default is 5.
 * `retryinterval` - interval between retries in seconds. The default is "5s".
 * `timeout` - timeout for the fence agent in seconds. The default is "60s".
+* `remediationStrategy` - either `ResourceDeletion` or `OutOfServiceTaint`:
+    * `ResourceDeletion`: This remediation strategy removes the pods on the node.
```
slintes, line 196: Maybe it's me only, but why don't call things how they are called already? 🤔 
Here: we do not `remove` anything, we `delete` them... that's the actual method name we use. That's the `D` in CRUD. 🤷🏼‍♂️ (Same below)

(and I don't block on this, just a general thought, which I had more often at various places)

slintes: /lgtm

in case you want to change...
/hold

clobrano: > why don't we call things how they are called already

for me, I think it's just an automatically mechanism to avoid repetitions in written documents, but indeed when explaining something this habit can be limited :)

slintes (approved):

razo7 (approved):

PR #150, opened 2024-07-01
Update to CentOS Stream9

slintes:
#### Why we need this PR
Container build fails because of invalid yum repo config.

#### Changes made
Update to CentOS Stream9, Stream8 is EOL

PR #151, opened 2024-07-23
Update Makefile

clobrano:
#### Why we need this PR

The current Makefile needs update for
- adding mandatory annotation to OKD bundle
- update golang version
- update tools

adding the mandatory annotation, I also noted an issue in CI which I decided to fix here, see the first change in the next section.

#### Changes made

- Fix tag version in release document created by CI
- Bump golang to version 1.22
- Add OCP annotation to OKD bundle
- Bump kustomize to version v5.4.2
- Fix verify-unchanged script log
- Bump controller-gen and envtest versions
- Bump ginkgo version to v2.19.0
- vendoring due to ginkgo version bump
- Bump goimports, opm and sort-imports versions
- Bump operator-sdk version to v1.35
- Bump envtest k8s version to 1.30


Makefile
```diff
@@ -257,13 +261,13 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 ## Some addition to bundle creation in the bundle
 DEFAULT_ICON_BASE64 := $(shell base64 --wrap=0 ./config/assets/medik8s_blue_icon.png)
 export ICON_BASE64 ?= ${DEFAULT_ICON_BASE64}
-export BUNDLE_CSV ?="./bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml"
+export CSV ?="./bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml"
```
razo7, line 264: Any reason to modify this var?  BUNDLE_CSV seems more appropriate as there are two CSVs in the repo. One is under _config_ and one is under _bundle_...

go.mod
```diff
@@ -1,12 +1,12 @@
 module github.com/medik8s/fence-agents-remediation
 
-go 1.20
+go 1.22
```
razo7, line 3: Update the Golang version in https://github.com/medik8s/fence-agents-remediation/blob/main/.ci-operator.yaml as well for OCP CI

Makefile
```diff
@@ -278,25 +282,38 @@ add-replaces-field: ## Add replaces field to the CSV
...
-bundle-community-rh: bundle-community  ## Generate bundle manifests and metadata customized to Red Hat community release
-	echo -e "\n  # Annotations for OCP\n  com.redhat.openshift.versions: \"v${OCP_VERSION}\"" >> bundle/metadata/annotations.yaml
+.PHONY: bundle-community-okd
+bundle-community-okd: bundle-community  ## Generate bundle manifests and metadata customized to Red Hat community release
 	$(MAKE) add-replaces-field
+	$(MAKE) add-ocp-annotations
+	echo -e "\n  # Annotations for OCP\n  com.redhat.openshift.versions: \"v${OCP_VERSION}\"" >> bundle/metadata/annotations.yaml
+
+.PHONY: add-ocp-annotations
+add-ocp-annotations: yq ## Add OCP annotations
```
razo7, line 303: Is there a reason to add it here (upstream) and not just downstream?

razo7 (commented):
Left small nits, and please update the PR title :)

Makefile
clobrano: I noticed that most of the other operators use `CSV`, and it is used in targets handling the bundle, so I felt OK to call it `CSV` (but mostly to simplify comparison against the other Makefiles for the same features).

Makefile
clobrano: OKD is also verifying their existence.

https://github.com/medik8s/self-node-remediation/pull/221

I will update the commit message

Makefile
razo7: SGTM

clobrano:
Dropping last commit, as it is already included in https://github.com/medik8s/fence-agents-remediation/pull/150

clobrano:
superseded by #152

PR #153, opened 2024-08-07
Remove Agent Packages

razo7:
Remove agents which use ssh and telnet

#### Changes made

Replace `fence-agents-all` package with only the packages for the agents that won't use telnet or SSH.


clobrano:
/lgtm
Giving a chance to get more reviews
/hold

PR #154, opened 2024-08-29
Error aggregation for Validation Webhooks

k-keiichi-rh:
#### Why we need this PR
Refactoring for Validation Webhooks with error aggregation
This is suggested [here](https://github.com/medik8s/fence-agents-remediation/pull/92#discussion_r1489509957).

#### Changes made
Run all validations and return an aggregated error

### DIFF
--- a/api/v1alpha1/fenceagentsremediation_webhook.go
+++ b/api/v1alpha1/fenceagentsremediation_webhook.go
@@ -19,9 +19,8 @@ package v1alpha1
 import (
 	"fmt"
 
-	"github.com/pkg/errors"
-
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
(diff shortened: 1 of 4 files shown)


api/v1alpha1/fenceagentsremediation_webhook.go
```diff
@@ -71,30 +70,35 @@ func (far *FenceAgentsRemediation) ValidateDelete() (admission.Warnings, error)
...
-	if _, err := validateAgentName(farSpec.Agent); err != nil {
-		return nil, err
-	}
-	return validateStrategy(farSpec.RemediationStrategy)
+	aggregated := errors.NewAggregate([]error{
+		validateAgentName(farSpec.Agent),
+		validateStrategy(farSpec.RemediationStrategy),
+	})
+
+	return nil, aggregated
```
razo7, line 78: NIT: unrelated to this change but I guess it would be clearer that we return an empty warning instead of a nil.
```suggestion
	return admission.Warnings{}, aggregated
```

k-keiichi-rh: Thank you for the suggestion. I will fix it as you mentioned.
After applying this change, I got the following error in the unit tests:
```
  > Enter [It] should be denied - /home/kii/git/kii/fence-agents-remediation/api/v1alpha1/fenceagentsremediation_webhook_test.go:50 @ 09/22/24 09:23:38.584
  2024-09-22T09:23:38-04:00     INFO    fenceagentsremediation-resource validate create {"name": "test-fence_ipmilan"}
  [FAILED] Unexpected non-nil/non-zero argument at index 0:
        <admission.Warnings>: admission.Warnings{}
  In [It] at: /home/kii/git/kii/fence-agents-remediation/vendor/github.com/onsi/gomega/internal/assertion.go:62 @ 09/22/24 09:23:38.584
```
I can simply avoid this error with the following:
[Before]
```
Expect(outOfServiceStrategy.ValidateCreate()).Error().To(MatchError(ContainSubstring(outOfServiceTaintUnsupportedMsg)))
```
[After]
```
_, err := outOfServiceStrategy.ValidateCreate()
Expect(err).To(MatchError(ContainSubstring(outOfServiceTaintUnsupportedMsg)))
```
But I think it's not a smart fix to avoid this error. Is there another good way to fix?

razo7: > Is there another good way to fix?

We can simply test whether the admission warning is empty
```
warnings, err := outOfServiceStrategy.ValidateCreate()
ExpectWithOffset(1, warnings).To(BeEmpty())
Expect(err).To(MatchError(ContainSubstring(outOfServiceTaintUnsupportedMsg)))
```

k-keiichi-rh: Thank you for the comment. I fixed it as you suggested.

razo7:
Waiting for https://github.com/openshift/release/pull/58825 to test it on OCP 4.18,

PR #155, opened 2024-08-29
Remove duplicated code in e2e tests

k-keiichi-rh:
#### Why we need this PR
Refactoring for the duplicated code in e2e tests
#### Changes made
Extract the common test routine and remove the duplication

### DIFF
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -60,7 +60,6 @@ var _ = Describe("FAR E2e", func() {
 		pod                              *corev1.Pod
 		startTime, nodeBootTimeBefore    time.Time
 		err                              error
-		remediationStrategy              v1alpha1.RemediationStrategyType
 	)
 	BeforeEach(func() {
 		// create FAR CR spec based on OCP platformn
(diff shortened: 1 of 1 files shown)


test/e2e/far_e2e_test.go
```diff
@@ -163,8 +126,18 @@ var _ = Describe("FAR E2e", func() {
 				remediationTimes = append(remediationTimes, time.Since(startTime))
 			})
 		})
+	}
+
+	Context("stress cluster with OutOfServiceTaint remediation strategy", func() {
```
razo7, line 131: The context doesn't match the RemediationStrategy.

```suggestion
	Context("stress cluster with ResourceDeletion remediation strategy", func() {
```

test/e2e/far_e2e_test.go
```diff
@@ -86,48 +85,12 @@ var _ = Describe("FAR E2e", func() {
...
-				remediationTimes = append(remediationTimes, time.Since(startTime))
-			})
-			It("should successfully remediate the second node", func() {
-				checkRemediation(nodeName, nodeBootTimeBefore, pod, remediationStrategy)
-				remediationTimes = append(remediationTimes, time.Since(startTime))
-			})
-		})
-	})
-	Context("stress cluster with OutOfServiceTaint remediation strategy", func() {
+	runFARTests := func(remediationStrategy v1alpha1.RemediationStrategyType, skipCondition func() bool) {
```
clobrano, line 89: Maybe I'm missing something, but can we just use a normal function here?

```go
func runFARTest(...) bool {
...
}

Comment (slintes) [reply test/e2e/far_e2e_test.go:89]: I've seen and used this pattern especially in tests more often. Advantages IMHO:

- the function is "physically" located nearer to the code where it's used, which improves readability IMHO
- the function has direct access to vars of the context, so less args are needed
- often these functions are very special for the tests of a specific context. Making them a regular `func` would make them available for all tests though, which isn't needed

Comment (clobrano) [reply test/e2e/far_e2e_test.go:89]: Fair enough, could we make it stand up a bit more in the context then? Even some space around might be enough

File: test/e2e/far_e2e_test.go (hunk 1 above)
Comment (k-keiichi-rh) [reply test/e2e/far_e2e_test.go:131]: Sorry. Fixed it.

File: test/e2e/far_e2e_test.go (hunk 2 above)
Comment (k-keiichi-rh) [reply test/e2e/far_e2e_test.go:89]: @slintes Thank you for the explanation. This is everything I thought.
In the current code, the test routine I wrote is duplicated and it will be troubled for maintenance in the future.
To remove the duplicated code, it needs access to vars of the context. This is the reason why I used this pattern.
@clobrano 
Thank you for the comment. I added the following comment. Is that fine for you?
```
+       // runFARTests is a utility function to run FAR tests.
+       // It accepts a remediation strategy and a condition to determine if the tests should be skipped.
+       runFARTests := func(remediationStrategy v1alpha1.RemediationStrategyType, skipCondition func() bool) {
```

clobrano: Yes, thank you

razo7:
If the PR changes suit your needs and it is ready to be approved and reviewed, then please remove `[WIP]` from the PR title as it adds a blocker label `do-not-merge/work-in-progress` @k-keiichi-rh

k-keiichi-rh:
As far as I checked, I have no concerns for this patch.
I appreciate it if you could review the patch. Thanks.

razo7:
Looking at the logs it works as designed.
Thank you @k-keiichi-rh for this improvement to the e2e test code.
/lgtm

PR #156, opened 2024-09-18
Build manager binary in a modular way

razo7:
#### Why we need this PR

Allow default values for building Golang manager binary using build.sh, and support for other flags.

#### Changes made

- Build manager binary in a modular way

Relates to [ECOPROJECT-1972](https://issues.redhat.com//browse/ECOPROJECT-1972)

#### Test plan

Similar to https://github.com/medik8s/node-maintenance-operator/pull/135

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -1,5 +1,4 @@
-#!/bin/bash
-set -ex
+#!/bin/bash -ex
 
 GIT_VERSION=$(git describe --always --tags || true)
 VERSION=${CI_VERSION:-${GIT_VERSION}}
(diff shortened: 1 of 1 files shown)


razo7:
The suggested changes should not affect nor interfere with how the newly created binary without seeing new Golang flags for the build.
However, the e2e test in 4.17 is still flaky. Initially, it was a deployment/environment issue and now the reboot simply doesn't succeed. We will do another run, prior to a closer inspection.
/retest

slintes (approved):

PR #159, opened 2024-11-06
Remediate Ready Worker Node in E2E

razo7:
#### Why we need this PR

1. Fix a trailing error when a node becomes unresponsive from the `resourceDeletion` strategy and it is selected for remediation for `outOfService` strategy. Then there is a failure of getting the node boot time in the BeforeEach prior to CR creation https://github.com/medik8s/fence-agents-remediation/blob/main/test/e2e/far_e2e_test.go#L114

```
  [FAILED] failed to get boot time of the node
  Unexpected error:
      <wait.errInterrupted>: 
      timed out waiting for the condition
      {
          cause: <*errors.errorString | 0xc0003cd250>{
              s: "timed out waiting for the condition",
          },
      }
  occurred
```
2. Reboot doesn't succeed on the second node remediation (per strategy), as it can't get the boot time of the node https://github.com/medik8s/fence-agents-remediation/blob/main/test/e2e/far_e2e_test.go#L430 passing a timeout.

```
Timeout for node reboot has passed, even though FAR CR has been created 
The function passed to Eventually returned the following error: <wait.errInterrupted>: timed out waiting for the condition { cause: <*errors.errorString | 0xc0003cd250> { s: "timed out waiting for the condition", }, }
```
3. The tested pods for checking the boot time have been evicted due to [FAR remediation taint](https://github.com/medik8s/fence-agents-remediation/blob/main/api/v1alpha1/fenceagentsremediation_types.go#L30
), while they should tolerate it. 

#### Changes made

1. Improve the function `getAvailableWorkerNodes` to `getReadyWorkerNodes` by getting a list of all the ready worker nodes. Otherwise, we will try to remediate the node and fail to fetch its boot time at BeforeEach. 
2. Function `getNodeRoleFromMachine` will add the node role to the list of nodes before every e2e test
3. Fix listing machines with no associated node - too long and messy return message.
4. Extend RetryCount for (AWS) agent.
5. Move cluster information from BeforeEach to an independent test, and skip subsequent tests on failure.
6. Tested pods toleration should use the [Exist operator over Equal](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#concepts) as the toleration value is irrelevant

[ECOPROJECT-2337](https://issues.redhat.com//browse/ECOPROJECT-2337)

### DIFF
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -55,10 +55,6 @@ var _ = Describe("FAR E2e", func() {
 		fenceAgent, nodeIdentifierPrefix string
 		testShareParam                   map[v1alpha1.ParameterName]string
 		testNodeParam                    map[v1alpha1.ParameterName]map[v1alpha1.NodeName]string
-		selectedNode                     *corev1.Node
-		nodeName                         string
-		pod                              *corev1.Pod
-		startTime, nodeBootTimeBefore    time.Time
 		err                              error
 	)
 	BeforeEach(func() {
@@ -89,14 +85,20 @@ var _ = Describe("FAR E2e", func() {
 	// runFARTests is a utility function to run FAR tests.
 	// It accepts a remediation strategy and a condition to determine if the tests should be skipped.
 	runFARTests := func(remediationStrategy v1alpha1.RemediationStrategyType, skipCondition func() bool) {
-		var availableWorkerNodes *corev1.NodeList
+		var (
+			availableWorkerNodes          *corev1.NodeList
+			selectedNode                  *corev1.Node
+			nodeName                      string
+			pod                           *corev1.Pod
+			startTime, nodeBootTimeBefore time.Time
+		)
 		BeforeEach(func() {
 			if skipCondition() {
 				Skip("Skip this block due to unsupported condition")
 			}
 
 			if availableWorkerNodes == nil {
-				availableWorkerNodes = getAvailableWorkerNodes()
+				availableWorkerNodes = getReadyWorkerNodes()
 			}
 			selectedNode = pickRemediatedNode(availableWorkerNodes)
 			nodeName = selectedNode.Name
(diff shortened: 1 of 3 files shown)


test/e2e/far_e2e_test.go
```diff
@@ -48,55 +47,66 @@ const (
...
-			nodeIdentifierPrefix = nodeIdentifierPrefixAWS
-			By("running fence_aws")
-		case configv1.BareMetalPlatformType:
-			fenceAgent = fenceAgentIPMI
-			nodeIdentifierPrefix = nodeIdentifierPrefixIPMI
-			By("running fence_ipmilan")
-		default:
-			Skip("FAR haven't been tested on this kind of cluster (non AWS or BareMetal)")
-		}
+	When("trying to identify cluster platform", func() {
```
slintes, line 62: Why did you switch from BeforeEach to When/It/stopTesting?

test/e2e/far_e2e_test.go
```diff
@@ -240,24 +250,36 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
...
+	for _, node := range availableWorkerNodes.Items {
+		for _, condition := range node.Status.Conditions {
+			if condition.Type == corev1.NodeReady && condition.Status == corev1.ConditionTrue {
+				readyWorkerNodes.Items = append(readyWorkerNodes.Items, node)
+				break // "Ready" was found
+			}
+		}
+	}
+	if len(readyWorkerNodes.Items) < 1 {
+		Skip("There isn't an available (and ready) worker node in the cluster")
```
slintes, line 273: this should result in a failure IMHO

test/e2e/far_e2e_test.go
```diff
@@ -240,24 +250,36 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
...
 	}
-	return availableNodes
+	return readyWorkerNodes
```
slintes, line 275: nit: why returning a list when we only add max one to it?

test/e2e/far_e2e_test.go
```diff
@@ -240,24 +250,36 @@ func buildNodeParameters(clusterPlatformType configv1.PlatformType) (map[v1alpha
...
 }
 
 // pickRemediatedNode randomly returns a next remediated node from the current available nodes,
 // and then the node is removed from the list of available nodes
 func pickRemediatedNode(availableNodes *corev1.NodeList) *corev1.Node {
 	if len(availableNodes.Items) < 1 {
-		Fail("No available node found for remediation")
+		Skip("There isn't an available (and ready) worker node in the cluster")
```
slintes, line 282: why switch from Fail to Skip? This is hiding that we don't run tests...

test/e2e/far_e2e_test.go
```diff
@@ -303,7 +325,7 @@ func createFAR(nodeName string, agent string, sharedParameters map[v1alpha1.Para
 			SharedParameters:    sharedParameters,
 			NodeParameters:      nodeParameters,
 			RemediationStrategy: strategy,
-			RetryCount:          5,
+			RetryCount:          10,
```
slintes, line 328: do we need this? IIUC the issue was starting a test pod, not rebooting the node?

slintes (changes requested):
general remark: we should fail tests when we don't run them, otherwise the risk to just be happy about fast and green tests, while something significant went wrong, is too high IMHO

test/e2e/far_e2e_test.go
razo7: I thought that it was redundant to fetch and print the cluster information (e.g., Platform) and agent parameter values per every e2e test (4 times). Testing and fetching this information in the beginning with independent tests saves (minimal to zero) time, and removes compatibility issues with the agent per every test.

test/e2e/far_e2e_test.go
razo7: Yes, the test pod wasn't the only case.

test/e2e/far_e2e_test.go
razo7: >when we only add max one to it?

Can you elaborate? The function returns the list of Ready worker nodes. Assuming there are 3 worker nodes in the cluster, then the returned list can be at most 3 nodes.

slintes: ah sorry, my fault, the `break` in the loop refers to the conditions only

slintes:
/lgtm

Feel free to address my suggestion or unhold :)
/hold

slintes (approved):

razo7:
It would be nice if we can merge https://github.com/openshift/release/pull/58825 and then test the PR with OCP 4.18

razo7:
https://github.com/openshift/release/pull/58825 tests are failing, and only when this PR will be merged then the tests will begin to be green again.

/retest

PR #168, opened 2025-01-12
Remove finalizer on NHC Timeout deleted FAR CR

razo7:
Create three functions to reconcile logic into smaller usable functions: 

- Remove FAR CR finalizer on deletion (_removeFarFinalizer_)
- Remove FAR taint on deletion (_removeFarTaints_)
- Stop agent execution (_stopAgentAndGetCrStatus_) 

These functions will be used when FAR CR is deleted (on success) or NHC time-out annotation exists and FAR CR is deleted (on failure/timed-out). The timed out is set mostly on escalating remediations.

#### Why we need this PR

FAR should allow the removal of redundant remediations - remediation of a node that no longer exists since MDR deleted the node and triggered the provisioning of a new one.

#### Changes made

When FAR detects a remediation that is both deleted and marked as Timeout by NHC it will remove the added taints and finalizer.

[ECOPROJECT-2489](https://issues.redhat.com//browse/ECOPROJECT-2489)


controllers/fenceagentsremediation_controller.go
```diff
@@ -132,13 +132,13 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		return emptyResult, err
 	}
 
-	// Check NHC timeout annotation
-	if isTimedOutByNHC(far) {
+	// Check NHC timeout annotation and stop the agent, since remediation is no longer relevant (most likely because fixed by a different remediator)
+	if r.Executor.Exists(far.GetUID()) && isTimedOutByNHC(far) {
```
slintes, line 136: you add a check here, and few lines below you requeue.... I think that's a significant unwanted change in execution flow, what happens on next reconcile?

PR #171, opened 2025-03-05
Create Credentials Secret

mshitrit:
#### Why we need this PR

- Don't use plaintext values of credentials inside CR spec. Fetch them from a secret.

#### Changes made
- Update FenceAgentsRemediation CRD with `CredentialParameters []ParameterName` to list the parameters that we gather their values from a secret.

#### Test plan
- New unit tests for buildFenceAgentParams function


mshitrit:
This is a duplicate of #158

controllers/fenceagentsremediation_controller.go
```diff
@@ -337,52 +343,119 @@ func getNodeName(far *v1alpha1.FenceAgentsRemediation) string {
...
+func resolveParameterValueFromSecret(ctx context.Context, c client.Client, far *v1alpha1.FenceAgentsRemediation, paramName v1alpha1.ParameterName) (string, error) {
+	var secret *corev1.Secret
+	var err error
+	logger := ctrl.Log.WithName("resolve-value-from-secret")
+	secretName := far.Name
+	// fetch secret/node name from remediation's annotation if present
+	if annotatedName, exist := far.Annotations["remediation.medik8s.io/node-name"]; exist {
+		secretName = annotatedName
+	}
+	secret, err = fetchSecret(ctx, c, client.ObjectKey{Name: secretName, Namespace: far.Namespace})
```
slintes, line 365: so IIUC:
- we have a secret with the same name as the node
- we look for a single entry in it, coming from a single value given in CredentialParameters

that triggers some questions...
- how to provide not node specific shared credentials?
- why do we need to select specific data from the secret instead of using all?
- why calling this for each single parameter instead of once for all?

Sorry, but IMHO we should go back to the design phase...

mshitrit: >how to provide not node specific shared credentials?

IIUC this is handled by using a default secret [here](https://github.com/medik8s/fence-agents-remediation/pull/171/files#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R371), which would apply in case a node secret isn't found. I agree though that performance can be improved here.

>why calling this for each single parameter instead of once for all?

I agree, but seems like a simple enough performance improvement, hardly something that require a new design IMO 

>why do we need to select specific data from the secret instead of using all?

I'm probably missing something trivial but I'm not sure I understand the distinction between this and # 2 :sweat_smile: any chance you can elaborate a bit more or give an example ?

slintes: >> how to provide not node specific shared credentials?

> IIUC this is handled by using a default secret 

I don't see how to use 2 secrets, one for node specific params, one for shared. It always looks for the node specific one, and always falls back to the default if the node specific one isn't found...

> why do we need to select specific data from the secret instead of using all?

what I mean with this:
Assume we have secret filled with `ip`, `username` and `password`. Why would a user need to define which of these to use? Isn't it a fair assumption that all data in the secret is relevant?

mshitrit: >I don't see how to use 2 secrets, one for node specific params, one for shared. It always looks for the node specific one, and always falls back to the default if the node specific one isn't found...

I agree it would be a nice improvement, but it seems like a simple implementation change of checking for the value in both secrets .

>Why would a user need to define which of these to use? 

IIUC ATM (before this PR) we already use Key value pairs in FAR CR for mapping those values .
I think that changing that is probably out of scope for this PR, which purpose is to solve the security issue.

>Isn't it a fair assumption that all data in the secret is relevant?

Maybe it is, but I think that removing those params from FAR CR will reduce the user visibility to the necessary credentials params as they would be "hided" in the secret.
TBH I am really not sure on which approach is better, but I think that it's ok to leave it out of scope for now.

controllers/fenceagentsremediation_controller.go
```diff
@@ -46,12 +48,14 @@ import (
...
-	errorMissingParams     = "nodeParameters or sharedParameters or both are missing, and they cannot be empty"
-	errorMissingNodeParams = "node parameter is required, and cannot be empty"
+	errorMissingParams      = "nodeParameters or sharedParameters or both are missing, and they cannot be empty"
+	errorMissingNodeParams  = "node parameter is required, and cannot be empty"
+	errorFailFetchingSecret = "failed to fetch secret `%s` at namespace `%s`: %w"
 
 	SuccessFAResponse    = "Success: Rebooted"
 	parameterActionName  = "--action"
 	parameterActionValue = "reboot"
+	SharedSecretName     = "shared-secret"
```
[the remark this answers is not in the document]
mshitrit: Hmm maybe "fence-agents-remediation-shared-secret" ? 
Let me know if you have something else in mind.

controllers/fenceagentsremediation_controller.go
```diff
@@ -337,52 +385,90 @@ func getNodeName(far *v1alpha1.FenceAgentsRemediation) string {
...
-			fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, nodeVal)
+		if nodeVal, isFound := nodeMap[v1alpha1.NodeName(nodeName)]; isFound {
+			if _, exist := fenceAgentParamNames[paramName]; !exist {
+				fenceAgentParamNames[paramName] = true
+				fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, nodeVal)
+			}
 		} else {
 			err := errors.New(errorMissingNodeParams)
 			logger.Error(err, "Missing matching nodeParam and CR's name")
 			return nil, err
```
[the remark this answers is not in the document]
mshitrit: IIUC this check fail in case there are already node params, but not for the specific node under remediation.
So the assumption is that if there are node params they need to be defined for all of the nodes.

controllers/fenceagentsremediation_controller.go
```diff
@@ -337,52 +385,90 @@ func getNodeName(far *v1alpha1.FenceAgentsRemediation) string {
...
+		fenceAgentParams = appendParamToSlice(fenceAgentParams, parameterActionName, parameterActionValue)
+	}
 	return fenceAgentParams, nil
 }
 
 // appendParamToSlice appends parameters in a key-value manner, when value can be empty
 func appendParamToSlice(fenceAgentParams []string, paramName v1alpha1.ParameterName, paramVal string) []string {
+	stringParam := string(paramName)
+	if !strings.HasPrefix(stringParam, "--") {
+		stringParam = "--" + stringParam
```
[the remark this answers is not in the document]
mshitrit: Hmm good question.
IIUC the answer is yes, I assume this is the reason Or made the change.

I was able to find this quote:
>When called by other pieces of software (fenced, fence_node), fencing agents take arguments from standard input (as opposed to the command line) in the following format:
```
argument=value
#this line is ignored
argument2=value2
```
[source](https://github.com/ClusterLabs/fence-agents/blob/main/doc/FenceAgentAPI.md#what-is-given-to-fencing-agents-by-fenced--stonithd) 

IIUC this indicates that when fence agents are invoked by cluster management tools like fenced or fence_node, parameters are passed through standard input in a key-value format, without the -- prefix. This internal mechanism differs from direct command-line invocation, where parameters typically use the --option=value syntax.

controllers/fenceagentsremediation_controller.go
```diff
@@ -337,52 +385,90 @@ func getNodeName(far *v1alpha1.FenceAgentsRemediation) string {
...
+	fenceAgentParamNames := make(map[v1alpha1.ParameterName]bool)
+
+	// append shared parameters
 	for paramName, paramVal := range far.Spec.SharedParameters {
-		if paramName != parameterActionName {
-			fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, paramVal)
-		} else if paramVal != parameterActionValue {
-			// --action attribute was selected but it is different than reboot
+		//Verify action must be reboot
+		if paramName == parameterActionName && paramVal != parameterActionValue {
```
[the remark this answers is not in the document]
mshitrit: Not sure I 100% follow.
- Regarding "already exists" check: IIUC the purpose of this check is to make sure that the same param isn't defined in multiple groups (for example both in SharedParameters and NodeParameters).  I agree it would make sense to apply that for params defined in the secret (let me if I understood your intention correctly)
- Regarding "action param": IIUC the purpose here is to add action in case it isn't defined, this isn't a check that's done for each param group so I don't think using params from the secret should affect that logic

controllers/fenceagentsremediation_controller.go
```diff
@@ -337,52 +385,90 @@ func getNodeName(far *v1alpha1.FenceAgentsRemediation) string {
...
+}
+
 // buildFenceAgentParams collects the FAR's parameters for the node based on FAR CR, and if the CR is missing parameters
-// or the CR's name don't match nodeParameter name or it has an action which is different than reboot, then return an error
-func buildFenceAgentParams(far *v1alpha1.FenceAgentsRemediation) ([]string, error) {
+// or the CR's name don't match nodeParameter name, or it has an action which is different from reboot, then return an error
+func buildFenceAgentParams(far *v1alpha1.FenceAgentsRemediation, secretParams map[string]string) ([]string, error) {
 	logger := ctrl.Log.WithName("build-fa-parameters")
 	if far.Spec.NodeParameters == nil || far.Spec.SharedParameters == nil {
 		err := errors.New(errorMissingParams)
```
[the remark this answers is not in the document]
mshitrit: I thought [that](https://github.com/medik8s/fence-agents-remediation/pull/171/files#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R450-R454) bit covers it.
Did you had something else in mind ?

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
mshitrit: @beekhof since you are very familiar with fence agents you can probably provide a quick answer here.
TL:DR
We've noticed Or updated FAR code to always append fence agents params with "--" and wondering if that's a valid assumption.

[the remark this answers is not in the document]
mshitrit: @razo7 welcome back :-) , we can also use your input on that.

README.md
```diff
@@ -187,6 +187,7 @@ The FAR CR, `FenceAgentsRemediation`, is created by the admin and is used to tri
 The CR includes the following parameters:
 
 * `agent` - fence agent name. File name which is validated (by kubebuilder and Webhook) against a list of supported agents in the FAR pod.
+* `credentialarameters` - credential parameters for accessing the node to be remediated.
```
beekhof, line 190: Out of date

README.md
```diff
@@ -209,9 +210,10 @@ spec:
   retrycount: 5
   retryinterval: "5s"
   timeout: "60s"
+  credentialparameters:
+    --password
```
beekhof, line 214: Out of date

README.md
```diff
@@ -209,9 +210,10 @@ spec:
...
   sharedparameters:
     --username: "admin"
-    --password: "password"
```
beekhof, line 214: Out of date

controllers/fenceagentsremediation_controller.go
```diff
@@ -325,6 +335,41 @@ func (r *FenceAgentsRemediationReconciler) updateStatus(ctx context.Context, far
 	return nil
 }
 
+func (r *FenceAgentsRemediationReconciler) collectSecretParams(far *v1alpha1.FenceAgentsRemediation, ctx context.Context) (map[string]string, error) {
+	secretParams := map[string]string{}
+	nodeName := getNodeName(far)
+	nodeSecretName := fmt.Sprintf("fence-agents-credentials-node-%s", nodeName)
```
beekhof, line 341: This string is not referenced *anywhere*, meaning no-one can use this feature.

Please make the user specify this name, rather than to have them scour yet-to-be created documentation to find out about it.

README.md
mshitrit: IIUC you are referring to ` --password: "password"` which is already removed in this PR

api/v1alpha1/fenceagentsremediation_types.go
```diff
@@ -81,7 +81,8 @@ type FenceAgentsRemediationSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec
 	Timeout metav1.Duration `json:"timeout,omitempty"`
 
-	// SharedParameters are passed to the fencing agent regardless of which node is about to be fenced (i.e., they are common for all the nodes)
+	// SharedParameters are passed to the fencing agent regardless of which node is about to be fenced
+	// (i.e., they are common for all the nodes)
 	//+operator-sdk:csv:customresourcedefinitions:type=spec
 	SharedParameters map[ParameterName]string `json:"sharedparameters,omitempty"`
```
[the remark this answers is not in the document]
razo7: I think this comment can be ignored as the `CredentialParameters` field is declared after `NodeParameters` and not before `NodeParameters`.

controllers/fenceagentsremediation_controller.go
mshitrit: Yep here it is

>FAR CRD includes a new spec field, CredentialsSecretName, for the secret name, which is impractical when there are many secrets.

I was actually considering taking a hybrid approach  adding the Shared Secret to the spec and just the Node Name Secret prefix to the Spec (suffix is the node name).

Updated the design [here](https://docs.google.com/document/d/1O6EBqvTivEc9YjhFk5gB9a6BamRyMPe2NYAVluooZjk/edit?tab=t.0#heading=h.1p69gnjejvtd)  as a 3rd option.

controllers/fenceagentsremediation_controller.go
```diff
@@ -337,52 +382,114 @@ func getNodeName(far *v1alpha1.FenceAgentsRemediation) string {
...
-	nodeName := v1alpha1.NodeName(getNodeName(far))
 	for paramName, nodeMap := range far.Spec.NodeParameters {
-		if nodeVal, isFound := nodeMap[nodeName]; isFound {
-			fenceAgentParams = appendParamToSlice(fenceAgentParams, paramName, nodeVal)
+		if nodeVal, isFound := nodeMap[v1alpha1.NodeName(nodeName)]; isFound {
+			//Verify action must be reboot
+			if err := validateRebootAction(paramName, nodeVal, logger); err != nil {
+				return nil, err
+			}
+			// For node params we don't enforce uniqueness but use other value if defined, TODO explain why ?
```
razo7, line 436: IIUC, we don't enforce uniqueness, and we allow creating a fence agent command with the same parameter. I think that the node params should override shared params (and we can add a log for having the node param priority)

razo7: Furthermore, non-secret node params (i.e., `NodeParameters`) can't be duplicated with any secret param (i.e., `secretParams`).

razo7: We can change the flow of building the command from SharedParams, SecretParams, and then NodeParams to NodeParams, SharedParams, and then SecretParams.
This way, the priority of the parameters should be:

1.  NodeParams - specific to a node and unique
2. SharedParams - shared in the cluster and not confidential.
3. SecretParams (from node) -  confidential for specific node and should not be explicit as NodeParams.
4. SecretParams (from default secret) - shared and used by default when not set by CR or node-specific secret

controllers/fenceagentsremediation_controller.go
mshitrit: I've taken a crack at the first approach.
If you decide you have a different strong preference  let me know.

controllers/fenceagentsremediation_controller.go
mshitrit: >But secured params should not override public params of the same category, or the other way around. That would very likely be a configuration mistake, not?

Yep agreed.
In that case we'll get a validation error.

controllers/fenceagentsremediation_controller.go
```diff
@@ -46,11 +49,14 @@ import (
 
 const (
 	// errors
-	errorMissingParams     = "nodeParameters or sharedParameters or both are missing, and they cannot be empty"
-	errorMissingNodeParams = "node parameter is required, and cannot be empty"
+	errorMissingParams             = "nodeParameters or sharedParameters or both are missing, and they cannot be empty"
+	errorMissingNodeParams         = "node parameter is required, and cannot be empty"
+	errorParamDefinedMultipleTimes = "invalid multiple definition of FAR  param"
```
razo7, line 54: NIT: redundant space
```suggestion
	errorParamDefinedMultipleTimes = "invalid multiple definition of FAR param"
```

api/v1alpha1/fenceagentsremediation_types.go
```diff
@@ -98,6 +99,20 @@ type FenceAgentsRemediationSpec struct {
 	// +kubebuilder:validation:Enum=ResourceDeletion;OutOfServiceTaint
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	RemediationStrategy RemediationStrategyType `json:"remediationStrategy,omitempty"`
+
+	// NodeSecretPrefix is the prefix for the name of the Secret (followed by the node name) which will contain params needed for FAR in order to remediate a particular node.
+	// Using this Secret is optional.
+	// +kubebuilder:default:="fence-agents-credentials-node-"
```
razo7, line 105: Do we want to have CRD validation for this format?

controllers/fenceagentsremediation_controller.go
```diff
@@ -226,14 +233,19 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
-		faParams, err := buildFenceAgentParams(far)
+		secretParams, err := r.collectSecretParams(far, ctx)
 		if err != nil {
-			r.Log.Error(err, "Invalid shared or node parameters from CR", "Node Name", node.Name, "CR Name", req.Name)
+			r.Log.Error(err, "Failed collecting secrets data", "Node Name", node.Name, "CR Name", req.Name)
+			return emptyResult, err
+		}
+		faParams, err := buildFenceAgentParams(far, secretParams)
+		if err != nil {
+			r.Log.Error(err, "Invalid credential/shared/node parameter from CR", "Node Name", node.Name, "CR Name", req.Name)
```
razo7, line 243: NIT: For better logging, we can replace `credential` by `secret` as there are no more `credentialParam` in this PR. It would be nicer if it followed the order of parameter priority as mentioned at https://github.com/medik8s/fence-agents-remediation/pull/171/files#r2089036850

controllers/fenceagentsremediation_controller.go
```diff
@@ -325,6 +337,47 @@ func (r *FenceAgentsRemediationReconciler) updateStatus(ctx context.Context, far
...
+			for secretKey, secretVal := range nodeSecret.Data {
+				secretParams[secretKey] = string(secretVal)
+				r.Log.Info("found a value from secret", "secret name", nodeSecretName, "parameter name", secretKey)
+			}
+		}
+	}
+
+	return secretParams, nil
+}
+
```
razo7, line 380: IMHO it would be nicer to split the function logic into two as the manipulation on the secret is quite identical between the shard secret and node secret
```suggestion

// readSecret reads and adds the secret params if they are available, otherwise returns an error
func (r *FenceAgentsRemediationReconciler) readSecret(secretParams map[string]string, secretName, namespace string , ctx context.Context) (map[string]string, error) {
	secret, err := r.getSecret(ctx, client.ObjectKey{Name: scretName, Namespace: namespace})
	if err != nil && !apiErrors.IsNotFound(err) {
		r.Log.Error(err, "failed to fetch secret", "secret name", secretName, "namespace", namespace)
		return nil, fmt.Errorf(errorFailFetchingSecret, scretName, namespace, err)
	}

	// fill secret params from secret
	if secret != nil {
		for secretKey, secretVal := range secret.Data {
			secretParams[secretKey] = string(secretVal)
			r.Log.Info("found a value from secret", "secret name", secretName, "parameter name", secretKey)
		}
	}
	return secretParams, nil
}

// collectSecretParams collects the parameters from the shared secret and the node secret
func (r *FenceAgentsRemediationReconciler) collectSecretParams(far *v1alpha1.FenceAgentsRemediation, ctx context.Context) (map[string]string, error) {
	secretParams := map[string]string{}
	var err error
	
	// collect secret params from shared secret
	if len(far.Spec.SharedSecretName) > 0 {
		secretParams, err = r.readSecret(secretParams, far.Spec.SharedSecretName, far.Namespace, ctx)
		if err != nil {
			return nil, err
		}
	}
	// collect secret params from the node's secret
	if len(far.Spec.NodeSecretPrefix) > 0 {
		nodeSecretName := fmt.Sprintf("%s%s", far.Spec.NodeSecretPrefix, getNodeName(far))
		secretParams, err = r.readSecret(secretParams, nodeSecretName, far.Namespace, ctx)
		if err != nil {
			return nil, err
		}
	}

	return secretParams, nil
}
```

api/v1alpha1/fenceagentsremediation_types.go
mshitrit: I think not, it's just a default value.
I don't think we should limit the user to this format.

controllers/fenceagentsremediation_controller.go
```diff
@@ -325,6 +337,47 @@ func (r *FenceAgentsRemediationReconciler) updateStatus(ctx context.Context, far
...
+		sharedSecret, err := r.getSecret(ctx, client.ObjectKey{Name: sharedSecretName, Namespace: far.Namespace})
+		if err != nil && !apiErrors.IsNotFound(err) {
+			r.Log.Error(err, "failed to fetch secret", "secret name", sharedSecretName, "namespace", far.Namespace)
+			return nil, fmt.Errorf(errorFailFetchingSecret, sharedSecretName, far.Namespace, err)
+		}
+
+		//Fill secret params from shared secret
+		if sharedSecret != nil {
+			for secretKey, secretVal := range sharedSecret.Data {
+				secretParams[secretKey] = string(secretVal)
```
razo7, line 354: Does it include any [decryption](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#ensure-all-secrets-are-encrypted)? @mshitrit 
IIUC, we store the plaintext values from the secret in secretParams.

mshitrit: k8s automatically decrypts Secrets at the API server level so it's not something we need to manage, and since we don't create the Secrets IIUC it's up to the user to decide whether to encrypt the data or not.

api/v1alpha1/fenceagentsremediation_types.go
razo7: ok

controllers/fenceagentsremediation_controller.go
razo7: Looks better with the new changes, and we can remove the `, TODO explain why ?` comment or just add the above priority of parameters in comments or (better IMO) in readme.

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
mshitrit: Thanks !
I guess that only leaves the discussion on what to do in case "--" is missing.
One approach is appending it, another one might be validation error.
Since I think it's unrelated to this PR topic I suggest will deal with that separately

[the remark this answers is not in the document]
razo7: SGTM

controllers/fenceagentsremediation_controller.go
mshitrit: Ah Sorry, I  got it now.
Initially I thought you were referring to the order [here](https://github.com/medik8s/fence-agents-remediation/pull/171/files#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R248)

controllers/fenceagentsremediation_controller.go
beekhof: I was thinking of the admin specifying the full secret name, not just a prefix

beekhof: This gives them the ability to put everything in one secret, have one secret per field, per node, whatever suits their style

controllers/fenceagentsremediation_controller.go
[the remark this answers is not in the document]
beekhof: I have a vague recollection that there might have been a couple of fields for which the stdin equivalent of `--name` was `othername` at some point. No idea if that's still true.  Definitely there are some that are `--name` instead of `--name=value`, and there are short forms (eg. `-n value`), but on the command line never just `name`.

controllers/fenceagentsremediation_controller.go
mshitrit: I've updated my suggestion in the [design](https://docs.google.com/document/d/1O6EBqvTivEc9YjhFk5gB9a6BamRyMPe2NYAVluooZjk/edit?tab=t.0#heading=h.1p69gnjejvtd) at the 4th option.

@slintes let me know if that makes sense.

mshitrit: Yeah he was trying to capture maximum user flexibility option, where the trade-off is simplicity.
I've talked with him and he is fine with proceeding with the current design (my apologizes for forgetting to update).

controllers/fenceagentsremediation_controller.go
```diff
@@ -226,14 +233,19 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		}
 
 		r.Log.Info("Build fence agent command line", "Fence Agent", far.Spec.Agent, "Node Name", node.Name)
-		faParams, err := buildFenceAgentParams(far)
+		secretParams, err := r.collectRemediationSecretParams(far, ctx)
```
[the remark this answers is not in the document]
mshitrit: Regarding readability I have no particular preference.
In that particular case `collectRemediationSecretParams` is a method and `buildFenceAgentParams` is a function.
In order to move `collectRemediationSecretParams` inside `buildFenceAgentParams` it would mean changing `buildFenceAgentParams` into a method which will require a unit test refactor as well.
I prefer to avoid this in this PR, in case it's still something you feel strongly about let me know.

controllers/fenceagentsremediation_controller.go
mshitrit: In the current implementation:
a,f = allowed
b,c,d,e = not allowed

IIUC we agree on that implementation.
 
Current unit tests cover: a, b, c, d, e

Even though `f` is allowed I think there is a bug, where secure node params override all of the secure shared params (instead of just the duplicate ones).

 What's left: 
 - add unit test (which should confirm the bug I'm suspecting)
 - fix the bug
 
 Let me know if that makes sense.

config/rbac/role.yaml
```diff
@@ -22,6 +22,14 @@ rules:
   - list
   - update
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - secrets
```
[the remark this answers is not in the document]
mshitrit: I think it's a good idea.
Replacing 
`// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch`
with 
`// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch,namespace=openshift-workload-availability
`
Should do the trick.

The problem with that approach is that I'd be hard-coding the namespace, which isn't ideal (not sure if it might also affect our e2e tests).

Any idea how to work around this ?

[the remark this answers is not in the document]
mshitrit: >At other places system is used as placeholder for the namespace

That's a cool idea, even though we haven't used it in any marker I've searched online and saw that it was applied, I'll try it.

>Did you check what happens in the generated CSV? The secret rule should go from clusterPermissions to permissions without any namespace specified, then we are fine.

As far as I can tell the namespace is specified
```
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: manager-role
  namespace: openshift-workload-availability
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - get
  - list
  - watch
  ```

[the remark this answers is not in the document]
Comment (mshitrit) [reply config/rbac/role.yaml:28]: I guess we are fine than, will test on my cluster though just to be on the safe side.

File: controllers/fenceagentsremediation_controller.go [hunk 20]
```diff
@@ -325,6 +337,54 @@ func (r *FenceAgentsRemediationReconciler) updateStatus(ctx context.Context, far
...
+	if len(far.Spec.SharedSecretName) > 0 {
+		secretParams, err = r.collectSecretParams(far.Spec.SharedSecretName, far.Namespace, ctx)
+		if err != nil {
+			return nil, err
+		}
+	}
+	// collect secret params from the node's secret
+	nodeSecretName := r.getNodeSecretName(far)
+	if len(nodeSecretName) > 0 {
+		secretParams, err = r.collectSecretParams(nodeSecretName, far.Namespace, ctx)
```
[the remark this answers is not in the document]
mshitrit: >Maybe it makes sense to put all this parameter handling code into its own package

I think it's a bit of an overkill ATM but I've made some other suggested refactoring so it can be simply done in the future in case FAR code continue to evolve (i.e unit tests, turning `collectRemediationSecretParams` into method and moving `collectSecretParams` into `collectRemediationSecretParams` ).

clobrano (changes requested):

controllers/fenceagentsremediation_controller.go
```diff
@@ -46,12 +48,14 @@ import (
 
 const (
 	// errors
-	errorMissingParams     = "nodeParameters or sharedParameters or both are missing, and they cannot be empty"
-	errorMissingNodeParams = "node parameter is required, and cannot be empty"
+	errorMissingParams      = "nodeParameters or sharedParameters or both are missing, and they cannot be empty"
+	errorMissingNodeParams  = "node parameter is required, and cannot be empty"
+	errorFailFetchingSecret = "failed to fetch secret key `%s` from secret `%s` at namespace `%s`: %w"
```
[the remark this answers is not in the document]
mshitrit: It was already updated, this comment is in on an outdated code

config/rbac/role.yaml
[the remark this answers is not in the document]
mshitrit: That was my initial implementation, however Gemini suggested against it.
Rational was that we risk a potential future overwrite when generating the manifests.
It seems relevant since opposed to the Role, the required role binding weren't added automatically by applying the marker and I had to manually add them.

Let me know if you still prefer this path, (also copying below the original message)

```
However, there's a critical caveat when working with kubebuilder:

Risk of Overwrite: kubebuilder (via controller-gen) is designed to manage and potentially overwrite the YAML files in your config/rbac/ directory. If controller-gen decides it needs to regenerate role_binding.yaml (even if it's currently only putting the ClusterRoleBinding there), it might overwrite your manual additions.
Given that your role_binding.yaml currently contains a ClusterRoleBinding (which typically goes in cluster_role_binding.yaml by default kubebuilder convention), it suggests your RBAC setup might be a bit customized or from an older template.

Recommendation:

The safest and most robust approach is still to create a new, dedicated YAML file for your manually managed RoleBinding.
```

controllers/fenceagentsremediation_controller.go
```diff
@@ -226,14 +231,14 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		}
 
 		r.Log.Info("Build fence agent command line", "Fence Agent", far.Spec.Agent, "Node Name", node.Name)
-		faParams, err := buildFenceAgentParams(far)
+		faParams, err := r.buildFenceAgentParams(ctx, far)
 		if err != nil {
-			r.Log.Error(err, "Invalid shared or node parameters from CR", "Node Name", node.Name, "CR Name", req.Name)
+			r.Log.Error(err, "Invalid parameter from CR", "Node Name", node.Name, "CR Name", req.Name)
```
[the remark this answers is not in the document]
mshitrit: SGTM, also removed the confusing log since in any case all the scenarios are logged in the internal methods.

test/e2e/far_e2e_test.go
```diff
@@ -463,3 +425,89 @@ func checkRemediation(nodeName string, nodeBootTimeBefore time.Time, pod *corev1
...
+		username, password, err := e2eUtils.GetSecretData(clientSet, secretBMHName, "openshift-machine-api", "username", "password")
+		if err != nil {
+			log.Info("Can't get BMH credentials")
+			return nil, err
+		}
+		secrets["--username"] = username
+		secrets["--password"] = password
+	}
+	return secrets, nil
+}
```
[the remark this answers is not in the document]
mshitrit: This code isn't part of this PR, it shows as such due to refactoring that moved it around

bundle/manifests/fence-agents-remediation-manager-role_rbac.authorization.k8s.io_v1_role.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
mshitrit: Yep good catch.
It looks like it was generated from an earlier iteration (before I added manually the Role-binding).
(You should see it generated if you'll remove the role-binding)

mshitrit:
> I'm pretty sure I mentioned it at least once during reviews, but I can't find it: IMHO it would make a lot of sense to have a validating webhook, which validates fencing parameters a) according to our own rules and b) by running an actual status command with the given fence agent and parameters. It's too easy to mess parameters up, and noticing it when fencing is needed is too late. I know I probably can't convince anyone to add it in this PR, but please track this as a follow up.

I can't remember discussing it, but I agree it makes a lot of sense.
[Ticket](https://issues.redhat.com/browse/RHWA-65) created

clobrano (approved):

PR #172, opened 2025-03-12
[WIP]Test for checking if AWS EC2 instances remain up after the e2e test finished

Shai1-Levi:
Enforce that sentence will start with capital letter

#### Why we need this PR

Test

#### Changes made

Test


Shai1-Levi:
This PR was a test and it's not needed

PR #174, opened 2025-07-02
Adding template to shared parameters

mshitrit:
#### Why we need this PR
In order to avoid hard coding node details, for some fence agents a template mechanism could eliminate the need to hard code node specifics, i.e.

#### Changes made
- Adding the option to use the template "NodeName" in shared parameters
- In case this template exist switch it with actual node name during runtime
- Add a webhook validation to make sure the template is set properly in the Spec in order to avoid later runtime issues
- Adding Unit tests

[RHWA-97](https://issues.redhat.com//browse/RHWA-97)

## Summary by CodeRabbit

* **New Features**
  * Added support for Go template syntax in shared parameters and shared Secret parameters, allowing dynamic substitution of node names using `{{.NodeName}}`.
  * Documentation updated with usage instructions and examples for template support.

* **Bug Fixes**
  * Improved error handling to prevent fence agent execution if template syntax is invalid, with errors logged and command execution skipped.

* **Tests**
  * Expanded unit and integration test coverage for template processing, including correct substitution and error scenarios in both parameters and secrets.

### DIFF
--- a/api/v1alpha1/fenceagentsremediationtemplate_webhook.go
+++ b/api/v1alpha1/fenceagentsremediationtemplate_webhook.go
@@ -17,18 +17,13 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"fmt"
-
 	commonAnnotations "github.com/medik8s/common/pkg/annotations"
 
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
-
-	"github.com/medik8s/fence-agents-remediation/pkg/template"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


mshitrit:
> pretty straight forward, but what about secure shared params?

My understanding was that we only only plan yo apply this on shared param, but yeah I agree that it would make sense.

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
```diff
@@ -59,20 +64,55 @@ func (farTemplate *FenceAgentsRemediationTemplate) Default() {
...
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type
 func (farTemplate *FenceAgentsRemediationTemplate) ValidateDelete() (admission.Warnings, error) {
 	webhookFARTemplateLog.Info("validate delete", "name", farTemplate.Name)
 	return nil, nil
 }
+
+// validateFARTemplateSpec validates the underlying FenceAgentsRemediationSpec
+func validateFARTemplateSpec(spec *FenceAgentsRemediationSpec) error {
```
clobrano, line 114: IIUC, `validateFAR` is wrapped inside the new `validateFARTemplateSpec` only to ignore the warning. Couldn't we just do the same inside the old `validateFAR`?

api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
```diff
@@ -116,6 +116,69 @@ var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
...
+					Name:      "invalid-template",
+					Namespace: "test-namespace",
+				},
+				Spec: FenceAgentsRemediationTemplateSpec{
+					Template: FenceAgentsRemediationTemplateResource{
+						Spec: FenceAgentsRemediationSpec{
+							Agent: validAgentName,
+							SharedParameters: map[ParameterName]string{
+								"--systems-uri": "/redfish/v1/Systems/{{.NodeName", // Missing closing brace
+								"--hostname":    "{{.InvalidField}}",               // Invalid field
```
clobrano, line 134: Nit: we could explain that the template has the right format, but "unsupported" name

api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
```diff
@@ -116,6 +116,69 @@ var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
...
+								"--port":        "{{.NodeName}}.com",               // Valid template
+								"--invalid":     "/path/{{.NodeName",               // Another missing closing brace
```
clobrano, line 136: Here we could test a different error here. E.g. missing opening brackets, or spelling error ({{.NodeNme}})

pkg/template/template.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// TemplateData contains the data available for template substitution
+type TemplateData struct {
+	NodeName string `json:"nodeName"`
+}
+
+// ProcessParameterValue processes a parameter value string with Go template syntax
+// It replaces placeholders like {{.NodeName}} with actual values
+func ProcessParameterValue(paramValue string, nodeName string) (string, error) {
```
clobrano, line 32: nit: the function's name seems a bit too generic. Usually function handling templates are named like "RenderTemplate", we could call it "RenderParameterTemplate" for example.

clobrano (commented):
I left some comments, but nothing blocking

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
mshitrit: I'm not sure I follow.
`validateFAR` is used both to validate the remediation and the template. 
When it validates the remediation the warning isn't ignored and when it validates the template it is.
do you suggest to ignore the warning for both cases ?

clobrano: It isn't clear to me why the warnings are ignored in `validateFARTemplateSpec`. Also, `validateFAR`, does not explicitly ignores warnings, but it just returns an empty object as well, which is kind of the same, IIUC.

If `validateFARTemplateSpec` were to return both warning (as empty object) and error, wouldn't be simpler?

fabiand:
Do we also take care of the case where now only shared parameters are provided, no node specific ones anymore?

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
mshitrit: Got it, SGTM :+1:

mshitrit:
> Do we also take care of the case where now only shared parameters are provided, no node specific ones anymore?

I agree that this PR improves on this use case.
Basically now  because of the templating a single shared param may have different values (depending on the node name) were before that wasn't an option.

api/v1alpha1/fenceagentsremediationtemplate_webhook.go
clobrano: LGTM as well :+1:

controllers/fenceagentsremediation_controller.go
```diff
@@ -367,6 +372,19 @@ func (r *FenceAgentsRemediationReconciler) collectRemediationSecretParams(ctx co
 	return secretParams, nil
 }
 
+func (r *FenceAgentsRemediationReconciler) processSecretParams(secretParams map[string]string, secretName, nodeName string) (map[string]string, error) {
+	processedSecretParams := map[string]string{}
+	for key, val := range secretParams {
+		processedParamVal, err := template.RenderParameterTemplate(val, nodeName)
+		if err != nil {
+			r.Log.Error(err, "Failed to process parameter value stored in secret", "param key", key, "secret name", secretName)
```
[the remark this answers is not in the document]
mshitrit: I don't think so.
In this case the node name is the remediated node (which is logged many times) and not the node that is connected to the secret (since we are processing a shared secret and not a node secret).

clobrano:
/lgtm

giving a chance to close the other threads (if any left)
/hold

mshitrit:
> > > Do we also take care of the case where now only shared parameters are provided, no node specific ones anymore?
> > 
> > 
> > I agree that this PR improves on this use case. Basically now because of the templating a single shared param may have different values (depending on the node name) were before that wasn't an option.
> 
> I think the question is: does FAR work with shared parameters only?

I've added a test to verify the template can be created without node params (i.e only shared params) so it isn't blocked by FAR
//cc @fabiand

PR #175, opened 2025-07-07
[WIP] Support off action

kattz-kawa:
This new draft PR follows up on #124 and was created because a rebase was needed.  
The content and purpose of this PR are the same as #124.

[ECOPROJECT-1471](https://issues.redhat.com/browse/ECOPROJECT-1471)

### DIFF
--- a/controllers/fenceagentsremediation_controller.go
+++ b/controllers/fenceagentsremediation_controller.go
@@ -52,10 +52,12 @@ const (
 	errorParamDefinedMultipleTimes = "invalid multiple definition of FAR param"
 	errorFailGettingSecret         = "failed to get secret `%s` at namespace `%s`: %w"
 
-	SuccessFAResponse    = "Success: Rebooted"
-	parameterActionName  = "--" + actionName
-	actionName           = "action"
-	parameterActionValue = "reboot"
+	SuccessFAResponse           = "Success: Rebooted"
+	parameterActionName         = "--" + actionName
+	actionName                  = "action"
+	parameterDefaultActionValue = "reboot"
+	parameterRebootActionValue  = "reboot"
+	parameterOffActionValue     = "off"
 )
 
 // FenceAgentsRemediationReconciler reconciles a FenceAgentsRemediation object
@@ -234,6 +236,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		faParams, isRetryRequired, err := r.buildFenceAgentParams(ctx, far)
 		if err != nil {
 			if !isRetryRequired {
+				commonEvents.WarningEvent(r.Recorder, far, utils.EventReasonCrParamInvalid, utils.EventMessageCrParamInvalid)
 				return emptyResult, nil
 			}
 			return emptyResult, err
@@ -409,7 +412,7 @@ func (r *FenceAgentsRemediationReconciler) getSecret(ctx context.Context, secret
 }
 
 // buildFenceAgentParams collects the FAR's parameters for the node based on FAR CR, and if the CR is missing parameters
-// or the CR's name don't match nodeParameter name, or it has an action which is different from reboot, then return an error
+// or the CR's name don't match nodeParameter name, or it has an action which is different from reboot and off, then return an error
 func (r *FenceAgentsRemediationReconciler) buildFenceAgentParams(ctx context.Context, far *v1alpha1.FenceAgentsRemediation) (map[v1alpha1.ParameterName]string, bool, error) {
 	nodeName := getNodeName(far)
 	secretParams, err := r.collectRemediationSecretParams(ctx, far)
@@ -422,8 +425,8 @@ func (r *FenceAgentsRemediationReconciler) buildFenceAgentParams(ctx context.Con
 
 	// append shared parameters
 	for paramName, paramVal := range far.Spec.SharedParameters {
-		// Verify action must be reboot
-		if err := validateRebootAction(paramName, paramVal, r.Log); err != nil {
+		// Verify action must be reboot or off
+		if err := validateFenceAction(paramName, paramVal, r.Log); err != nil {
 			return nil, false, err
 		}
 		// Verify param isn't already defined
@@ -437,7 +440,7 @@ func (r *FenceAgentsRemediationReconciler) buildFenceAgentParams(ctx context.Con
 	for paramName, nodeMap := range far.Spec.NodeParameters {
 		if nodeVal, isFound := nodeMap[v1alpha1.NodeName(nodeName)]; isFound {
 			// Verify action must be reboot
-			if err := validateRebootAction(paramName, nodeVal, r.Log); err != nil {
+			if err := validateFenceAction(paramName, nodeVal, r.Log); err != nil {
 				return nil, false, err
 			}
 			// For node params we don't enforce uniqueness node param value will override shared param
@@ -455,7 +458,7 @@ func (r *FenceAgentsRemediationReconciler) buildFenceAgentParams(ctx context.Con
 	for secretKey, secretVal := range secretParams {
 		secretParam := v1alpha1.ParameterName(secretKey)
 		// Verify action must be reboot
-		if err := validateRebootAction(secretParam, secretVal, r.Log); err != nil {
+		if err := validateFenceAction(secretParam, secretVal, r.Log); err != nil {
 			return nil, false, err
 		}
 		if err := validateUniqueParam(fenceAgentParams, secretParam, r.Log); err != nil {
@@ -473,16 +476,17 @@ func (r *FenceAgentsRemediationReconciler) buildFenceAgentParams(ctx context.Con
 	// Add the reboot action with its default value - https://github.com/ClusterLabs/fence-agents/blob/main/lib/fencing.py.py#L103
 	if _, exist := fenceAgentParams[parameterActionName]; !exist {
 		r.Log.Info("`action` parameter is missing, so we add it with the default value of `reboot`")
-		fenceAgentParams[parameterActionName] = parameterActionValue
+		fenceAgentParams[parameterActionName] = parameterDefaultActionValue
 	}
 
 	return fenceAgentParams, false, nil
 }
 
-func validateRebootAction(paramName v1alpha1.ParameterName, paramVal string, logger logr.Logger) error {
-	if (paramName == actionName || paramName == parameterActionName) && paramVal != parameterActionValue {
-		// --action parameter with a different value from reboot is not supported
-		err := errors.New("FAR doesn't support any other action than reboot")
+func validateFenceAction(paramName v1alpha1.ParameterName, paramVal string, logger logr.Logger) error {
+	if (paramName == actionName || paramName == parameterActionName) &&
+		(paramVal != parameterRebootActionValue && paramVal != parameterOffActionValue) {
+		// --action parameter with a different value from reboot or off is not supported
+		err := errors.New("FAR doesn't support any other action than reboot and off")
 		logger.Error(err, "can't build CR with this action attribute", "action", paramVal)
 		return err
 	}
--- a/test/e2e/far_e2e_test.go
+++ b/test/e2e/far_e2e_test.go
@@ -29,7 +29,6 @@ import (
 const (
 	fenceAgentAWS            = "fence_aws"
 	fenceAgentIPMI           = "fence_ipmilan"
-	fenceAgentAction         = "reboot"
 	nodeIdentifierPrefixAWS  = "--plug"
 	nodeIdentifierPrefixIPMI = "--ipport"
 	testContainerName        = "test-container"
@@ -40,6 +39,7 @@ const (
 	timeoutTaint       = "2s"   // Timeout for checking the FAR taint
 	timeoutReboot      = "6m0s" // fencing with fence_aws should be completed within 6 minutes
 	timeoutAfterReboot = "5s"   // Timeout for verifying steps  after the node has been rebooted
+	timeoutPowerOff    = "6m0s"
 	pollTaint          = "100ms"
 	pollReboot         = "1s"
 	pollAfterReboot    = "250ms"
@@ -47,6 +47,7 @@ const (
 )
 
 var (
+	fenceAgentDefaultAction          string = "reboot"
 	stopTesting                      bool
 	remediationTimes                 []time.Duration
 	fenceAgent, nodeIdentifierPrefix string
@@ -59,7 +60,7 @@ var _ = Describe("FAR E2e", func() {
 		testNodeParam  map[v1alpha1.ParameterName]map[v1alpha1.NodeName]string
 	)
 	BeforeEach(func() {
-		testShareParam = buildSharedParameters(clusterPlatform, fenceAgentAction)
+		testShareParam = buildSharedParameters(clusterPlatform, fenceAgentDefaultAction)
 		var err error
 		testNodeParam, err = buildNodeParameters()
 		Expect(err).ToNot(HaveOccurred(), "can't get node information")
@@ -67,7 +68,7 @@ var _ = Describe("FAR E2e", func() {
 
 	// runFARTests is a utility function to run FAR tests.
 	// It accepts a remediation strategy and a condition to determine if the tests should be skipped.
-	runFARTests := func(remediationStrategy v1alpha1.RemediationStrategyType, skipCondition func() bool) {
+	runFARTests := func(remediationStrategy v1alpha1.RemediationStrategyType, testAction string, skipCondition func() bool) {
 		var (
 			availableWorkerNodes          *corev1.NodeList
 			selectedNode                  *corev1.Node
@@ -97,6 +98,7 @@ var _ = Describe("FAR E2e", func() {
 			var err error
 			// save the node's boot time prior to the fence agent call
 			nodeBootTimeBefore, err = e2eUtils.GetBootTime(clientSet, nodeName, testNsName, log)
+
 			Expect(err).ToNot(HaveOccurred(), "failed to get boot time of the node")
 
 		})
@@ -107,17 +109,39 @@ var _ = Describe("FAR E2e", func() {
 
 			// set the node as "unhealthy" by disabling kubelet
 			makeNodeUnready(selectedNode)
+			testShareParam["--action"] = testAction
 
 			startTime = time.Now()
 			far := createFAR(nodeName, fenceAgent, testShareParam, testNodeParam, remediationStrategy)
+			
 			DeferCleanup(deleteFAR, far)
+			// The node needs to be powered on after 'off' remediation
+			if testAction == "off" {
+				ctx := &e2eUtils.FenceCommandContext{
+					ClientSet: clientSet,
+					ControllerNS: operatorNsName,
+					Agent: fenceAgent,
+					SharedParams: testShareParam,
+					NodeParams: testNodeParam,
+					Logger: log,
+				}
+				DeferCleanup(e2eUtils.PowerOnNode, ctx, selectedNode.Name)
+			}
 		})
-		When("running FAR to reboot a node with secrets in shared parameters (legacy)", func() {
+		When("running FAR to remediate a node with secrets in shared parameters (legacy)", func() {
 			BeforeEach(func() {
 				testShareParam = addSecretsToSharedParams(testShareParam)
 			})
 			It("should successfully remediate the node", func() {
-				checkRemediation(nodeName, nodeBootTimeBefore, pod, remediationStrategy)
+				ctx := &e2eUtils.FenceCommandContext{
+					ClientSet: clientSet,
+					ControllerNS: operatorNsName,
+					Agent: fenceAgent,
+					SharedParams: testShareParam,
+					NodeParams: testNodeParam,
+					Logger: log,
+				}
+				checkRemediation(nodeName, nodeBootTimeBefore, pod, remediationStrategy, testAction, ctx)
 				remediationTimes = append(remediationTimes, time.Since(startTime))
 			})
 		})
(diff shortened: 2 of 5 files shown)


kattz-kawa:
@mshitrit, Thank you for the feedback. That makes a lot of sense.

In this case, I wasn’t able to push to the original branch, which is why I opened this new PR to continue the work on behalf of @k-keiichi-rh.  
I understand that keeping the original context is important, and I’ll keep that in mind for future contributions.

kattz-kawa:
I saw the rebase request. I've enabled "Allow edits by maintainers" on this PR, so please feel free to make any necessary changes or rebase directly if that's easier. Let me know if you'd prefer me to handle it instead. Thanks!

mshitrit:
@coderabbitai full review

controllers/fenceagentsremediation_controller.go
```diff
@@ -234,6 +236,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		faParams, isRetryRequired, err := r.buildFenceAgentParams(ctx, far)
 		if err != nil {
 			if !isRetryRequired {
+				commonEvents.WarningEvent(r.Recorder, far, utils.EventReasonCrParamInvalid, utils.EventMessageCrParamInvalid)
```
mshitrit, line 239: This might be a bit confusing since there are currently 3 use cases which can cause this:
1. Duplication of a param (which I think this event matches reasonably)
2. Invalid **value** of the action param (which I'm not sure is a great match)
3. missing params (which I think this event does not match at all)

test/e2e/utils/command.go
```diff
@@ -205,3 +221,134 @@ func GetPod(nodeName, containerName string) *corev1.Pod {
 		},
 	}
 }
+
+// GetPowerStatus returns the node's power status, otherwise it fails and returns an error
+func GetPowerStatus(ctx *FenceCommandContext, targetNodeName string) (string, error) {
+	log := ctx.Logger
+
+	far_controller, err := GetActiveFARControllerManagerPod(ctx.ClientSet, ctx.ControllerNS, ctx.Logger)
```
mshitrit, line 229: far_controller does not match camelCase naming convention

test/e2e/far_e2e_test.go
```diff
@@ -403,13 +465,18 @@ func verifyStatusCondition(nodeName, conditionType string, conditionStatus *meta
 }
 
 // checkRemediation verify whether the node was remediated
-func checkRemediation(nodeName string, nodeBootTimeBefore time.Time, pod *corev1.Pod, strategy v1alpha1.RemediationStrategyType) {
+func checkRemediation(nodeName string, nodeBootTimeBefore time.Time, pod *corev1.Pod, strategy v1alpha1.RemediationStrategyType, testAction string, ctx *e2eUtils.FenceCommandContext) {
```
mshitrit, line 468: is there any reason `ctx *e2eUtils.FenceCommandContext` is needed here ?
Can't we just pool this information from the FAR CR ?

test/e2e/utils/command.go
```diff
@@ -205,3 +221,134 @@ func GetPod(nodeName, containerName string) *corev1.Pod {
...
+		return "on", nil
+	} else if strings.Contains(output, "Status: OFF") {
+		return "off", nil
+	} else {
+		return "unknown", fmt.Errorf("unable to determine power status from output: %s", output)
+	}
+}
+
+// PowerOnNode executes the power-on command to turn on the node during the cleanup phase for the off-action test.
+func PowerOnNode(ctx *FenceCommandContext, targetNodeName string) (bool, error) {
```
mshitrit, line 259: Hmm I wonder if this is working on our AWS CI :thinking: , have you had a chance to verify  ?

test/e2e/utils/command.go
```diff
@@ -26,8 +26,23 @@ import (
...
 const (
 	containerTestName = "test-command"
+	// NOTE: These constants are duplicated from far_e2e_test.go for simplicity.
+	fenceAgentAWS = "fence_aws"
+	fenceAgentIPMI = "fence_ipmilan"
+	
 )
 
+// FenceCommandContext holds the information needed to run fence commands on the FAR controller pod
+type FenceCommandContext struct {
```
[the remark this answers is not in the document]
kattz-kawa: I introduced 'FenceCommandContext' struct to avoid passing multiple common/shared dependencies repeatedly across test helpers like `'wasNodePoweredOff'`, `'GetPowerStatus'`, and `'PowerOnNode'`.
This should help reduce repeated code, improve readability, and make the test code easier to maintain.

That said, maybe I'm missing a simpler way to achieve the same goal. If there's a cleaner pattern, like leveraging `context.Context` more effectively or pulling more from the FAR CR, I'd love to learn from that.

Let me know if you think this approach makes sense, or if you'd prefer a different direction.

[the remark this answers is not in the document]
mshitrit: Sure, I'll try to explain myself better.

My point is that it seems like `FenceCommandContext` is a placeholder for the data which is fetched from `FenceAgentsRemediation`.
Why not just use `FenceAgentsRemediation` ?

In case there is still data that isn't part of  `FenceAgentsRemediation` you can store it on  `context.Context` .
[Here](https://github.com/medik8s/node-healthcheck-operator/blob/main/controllers/nodehealthcheck_controller.go#L161) is an example.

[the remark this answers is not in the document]
kattz-kawa: Thanks for the clarification! That makes sense.  
I'll refactor accordingly and remove the `FenceCommandContext` struct.  
Appreciate the example, it really helped me understand the intended direction.

kattz-kawa:
> Sorry, but that is something you need to do yourself. Just rebase on your machine and force push the rebased code to your PR branch.

Got it, thanks for confirming.
I'll take care of the rebase and force-push the updated branch shortly.

controllers/fenceagentsremediation_controller.go
kattz-kawa: Thank you for pointing that out. I’ve updated the code.

test/e2e/utils/command.go
kattz-kawa: I’ve renamed far_controller to farController.

test/e2e/far_e2e_test.go
kattz-kawa: Originally, I used `ctx` to get the power status through the fencing command, because the FAR CR did not include reliable and direct power source information.
However, I found that `fence_aws --action=status` returns `Command Status: 2` when the power is OFF, which made it difficult to use.
So I changed the implementation of `GetPowerStatus` to retrieve the power status using the `machine-api` instead.
As a result, `ctx` is no longer needed, and I’ve removed it.

test/e2e/utils/command.go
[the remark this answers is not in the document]
kattz-kawa: I’ve updated the implementation. I removed the custom struct, as it is no longer necessary.

controllers/fenceagentsremediation_controller.go
mshitrit: Hmm I think I'm missing something, I could see a change.
Let me know if I'm missing something obvious.

1. Invalid value of the action param would still trigger this event, see [here](https://github.com/medik8s/fence-agents-remediation/pull/175/files#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R455-R457)
2. Same for missing params. (see [here](https://github.com/medik8s/fence-agents-remediation/pull/175/files#diff-05147ee3f0d225b5a60754b8b9bff2c569deb0a9d7eaddae33af680291d493a1R496-R500))

test/e2e/utils/command.go
```diff
@@ -205,3 +223,106 @@ func GetPod(nodeName, containerName string) *corev1.Pod {
 		},
 	}
 }
+
+// PowerOnNode executes the power-on command to turn on the node during the cleanup phase for the off-action test.
+func PowerOnNode(ctx context.Context, far *v1alpha1.FenceAgentsRemediation, targetNodeName string) (bool, error) {
+	clientset := ctx.Value(ClientSetContextKey).(*kubernetes.Clientset)
```
mshitrit, line 229: I think it's fine to pass `clientset` as a param to this method, it'll also keep consistency with the rest of the methods in this file.

test/e2e/utils/command.go
```diff
@@ -205,3 +223,106 @@ func GetPod(nodeName, containerName string) *corev1.Pod {
...
+	controllerNS := ctx.Value(ControllerNSContextKey).(string)
```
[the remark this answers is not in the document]
kattz-kawa: I've updated the code.

test/e2e/utils/command.go
kattz-kawa: I was a bit too focused on using `ctx` and didn’t consider the simpler approach.
Thanks again for pointing it out, I've updated the code accordingly.

test/e2e/far_e2e_test.go
```diff
@@ -40,13 +39,15 @@ const (
...
 	timeoutAfterReboot = "5s"   // Timeout for verifying steps  after the node has been rebooted
+	timeoutPowerOff    = "10m0s"
 	pollTaint          = "100ms"
 	pollReboot         = "1s"
 	pollAfterReboot    = "250ms"
 	skipOOSREnvVarName = "SKIP_OOST_REMEDIATION_VERIFICATION"
 )
 
 var (
+	fenceAgentDefaultAction          string = "reboot"
```
mshitrit, line 50: Nit: this can be const

test/e2e/utils/command.go
```diff
@@ -26,6 +26,9 @@ import (
 
 const (
 	containerTestName = "test-command"
+	// NOTE: These constants are duplicated from far_e2e_test.go for simplicity.
```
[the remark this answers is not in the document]
mshitrit: >I wasn’t sure if importing from the test file would be appropriate

I agree that production code shouldn't use test code variables, but in this case it's all test code so in theory I don't see any issue here.

However IIRC go will block you from importing variables form `_test`  since it woudn't make that distinction and built to prevent mix of test code variables into production code.

IMO the simplest way to solve this would be to remove the variables in `far_e2e_test.go` and instead import them from `command.go`

test/e2e/far_e2e_test.go
```diff
@@ -376,6 +397,23 @@ func wasNodeRebooted(nodeName string, nodeBootTimeBefore time.Time) {
 	log.Info("successful reboot", "node", nodeName, "offset between last boot", nodeBootTimeAfter.Sub(nodeBootTimeBefore), "new boot time", nodeBootTimeAfter)
 }
 
+// wasNodePoweredOff checks if the node remains powered off
+func wasNodePoweredOff(nodeName string) {
```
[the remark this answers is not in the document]
kattz-kawa: Thanks for the suggestion.

Just to mention: I named it `wasNodePoweredOff` to keep consistent with the existing `wasNodeRebooted`, which follows a similar pattern and behavior. I totally understand that `verifyNodePoweredOff` might be clearer in isolation, though.

If you think renaming it would improve readability, I will update it, just let me know what you prefer.

controllers/fenceagentsremediation_controller.go
```diff
@@ -462,7 +469,7 @@ func (r *FenceAgentsRemediationReconciler) buildFenceAgentParams(ctx context.Con
 	for paramName, nodeMap := range far.Spec.NodeParameters {
 		if nodeVal, isFound := nodeMap[v1alpha1.NodeName(nodeName)]; isFound {
 			// Verify action must be reboot
```
clobrano, line 471: This should be "reboot or off"

controllers/fenceagentsremediation_controller.go
```diff
@@ -480,7 +487,7 @@ func (r *FenceAgentsRemediationReconciler) buildFenceAgentParams(ctx context.Con
 	for secretKey, secretVal := range secretParams {
 		secretParam := v1alpha1.ParameterName(secretKey)
 		// Verify action must be reboot
```
clobrano, line 489: This should be "reboot or off"

test/e2e/utils/command.go
```diff
@@ -205,3 +216,101 @@ func GetPod(nodeName, containerName string) *corev1.Pod {
...
+			fmt.Sprintf("--action=%s", action),
+			fmt.Sprintf("--access-key=%s", accessKey),
+			fmt.Sprintf("--secret-key=%s", secretKey),
+			fmt.Sprintf("--region=%s", sharedparms["--region"]),
+			fmt.Sprintf("--plug=%s", nodeparms[v1alpha1.ParameterName("--plug")][v1alpha1.NodeName(targetNodeName)]),
+			"--retry-on=10",
+		}
+
+	case fenceAgentIPMI:
+		var username, password string
```
clobrano, line 269: Please, keep the "TODO" comments we have around this logic in main

test/e2e/utils/command.go
```diff
@@ -205,3 +216,101 @@ func GetPod(nodeName, containerName string) *corev1.Pod {
...
+	var err error
+	var agent string = far.Spec.Agent
+
+	nodeparms := far.Spec.NodeParameters
+	sharedparms := far.Spec.SharedParameters
+
+	switch agent {
+	case fenceAgentAWS:
+		var accessKey, secretKey string
+		accessKey, secretKey, err = GetSecretData(c, "aws-cloud-fencing-credentials-secret", "openshift-operators", "aws_access_key_id", "aws_secret_access_key")
```
clobrano, line 257: Please, use constants for this values. Same below for BMC.

clobrano (changes requested):
I left some comments

test/e2e/far_e2e_test.go
kattz-kawa: I’ve moved it.

test/e2e/utils/command.go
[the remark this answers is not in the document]
kattz-kawa: I’ve removed the variables in `far_e2e_test.go` and defined in `command.go`

controllers/fenceagentsremediation_controller.go
kattz-kawa: Thanks, I've updated it to 'reboot or off'.

controllers/fenceagentsremediation_controller.go
kattz-kawa: Thanks, I've updated it to 'reboot or off'.

test/e2e/utils/command.go
kattz-kawa: Done, TODOs are kept.

test/e2e/far_e2e_test.go
```diff
@@ -107,17 +109,23 @@ var _ = Describe("FAR E2e", func() {
...
 			makeNodeUnready(selectedNode)
+			testShareParam["--action"] = testAction
 
 			startTime = time.Now()
 			far := createFAR(nodeName, fenceAgent, testShareParam, testNodeParam, remediationStrategy)
+
 			DeferCleanup(deleteFAR, far)
+			// The node needs to be powered on after 'off' remediation
+			if testAction == "off" {
+				DeferCleanup(e2eUtils.PowerOnNode, clientSet, far, selectedNode.Name, log)
```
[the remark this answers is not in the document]
mshitrit: Nit: event though this should be triggered fine due to Defer stacking (i.e `waitForNodeHealthyCondition` will be triggered after `PowerOnNode`) IMO this is still quite confusing.
Since this is the only place `PowerOnNode` I think it'll be simpler just to add the `waitForNodeHealthyCondition` inside the `PowerOnNode` and have only one `DeferCleanup` triggered.

test/e2e/utils/command.go
```diff
@@ -205,3 +216,101 @@ func GetPod(nodeName, containerName string) *corev1.Pod {
...
+		log.Error(err, "Unsupported agent")
+		return false, err
+	}
+
+	bytes, err := waitForPodOutput(c, farController, cmd, "manager")
+
+	log.Info("Output of power-on command", "stdout", string(bytes))
+
+	if err != nil {
+		log.Error(err, "Power-on Failed")
```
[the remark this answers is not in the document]
mshitrit: Nits:
- Since not params are used in the error text I think `errors.New` is better than `fmt.Errorf`
- error is expected to start with lower case

test/e2e/far_e2e_test.go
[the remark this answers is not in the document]
kattz-kawa: I've updated both.

test/e2e/far_e2e_test.go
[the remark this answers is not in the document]
kattz-kawa: Thanks, I introduced `powerOnNodeAndWaitUntilReady` to improve the readability of the execution flow between `PowerOnNode` and `waitForNodeHealthyCondition`.

test/e2e/utils/command.go
mshitrit: ATM e2e test [fails](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/medik8s_fence-agents-remediation/175/pull-ci-medik8s-fence-agents-remediation-main-4.19-openshift-e2e/1952745856759238656), that's something that will need fixing before we can merge.

kattz-kawa: Do you think it would be reasonable to extend `verifyNodePoweredOff` timeout (`timeoutPowerOff`) as a fix?

I reviewed the must-gather logs and found that the test failure appears to be caused by a delay in the power-off operation of fence_aws, which is triggered by the far-controller. The command seems to have been retried several times before the node was actually powered off. Due to this delay, the `verifyNodePoweredOff` function hit its timeout, resulting in the test failure.

The simplest fix would be to increase the timeout value. However, I'm still unclear why the fence_aws command needed multiple retries in the CI environment.

Before proceeding with any changes, I’d appreciate your thoughts. Do you think extending the timeout is the right approach?

kattz-kawa:
> don't forget to update the README please :)

@slintes
I'm working on updating the README. Rather than modifying all existing mentions of "reboot", I'm planning to add a **separate section specifically for the newly supported `off` action**, explaining its purpose and usage. I think this approach will keep the documentation clearer and easier to maintain. Does that sound good to you? 

In parallel, I'm also addressing your comment by adding the timeout and removing unnecessary variables. 
Let me know if you have any preferences or suggestions. Thank you.

controllers/fenceagentsremediation_controller.go
```diff
@@ -52,11 +52,14 @@ const (
...
-	SuccessFAResponse    = "Success: Rebooted"
-	parameterActionName  = "--" + actionName
-	actionName           = "action"
-	parameterActionValue = "reboot"
+	errorUnsupportedAction         = "FAR doesn't support any other action than reboot and off"
+
+	SuccessFAResponse           = "Success: Rebooted"
+	parameterActionName         = "--" + actionName
+	actionName                  = "action"
+	parameterDefaultActionValue = "reboot"
```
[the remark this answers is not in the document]
kattz-kawa: I've updated that code.

clobrano (approved):
I think I am the last one

/lgtm

razo7:
Very good PR @kattz-kawa, and I have allowed all the testing.
I have some thoughts on small NITs, but that should not block the progress of this PR.

Please remove [WIP] in the PR's title for merging the PR :)

README.md
```diff
@@ -10,19 +10,32 @@ FAR uses a fence agent to fence a Kubernetes node. Generally, fencing is the pro
...
+
+### Power-off-based remediation
+
+If you choose the `off` action, FAR powers off the unhealthy node using the fence agent. Unlike `reboot`, the node does not automatically return to service. It remains powered off until manually recovered by an administrator. The taint and workload eviction behavior remain the same as in the `reboot` case, ensuring that the node is isolated and workloads are safely rescheduled elsewhere.
 
 FAR includes the `FenceAgentsRemediationTemplate` (or `fartemplate`) Custom Resource Definition ([CRD](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition)) for how to create a FenceAgentsRemediation CR. The template has the same fields as far CR (e.g., agent name) and it is used for automatically creating remediation CR by another operator/mechanism (e.g., [NHC](#far-with-nhc)). The other operator is responsible of creating (and eventually deleting) the FenceAgentsRemediation CR with the name of the unhealthy node, even though FAR can be used manually without fartemplate and an additional operator (see [standalone FAR](#standalone-far)).
```
razo7, line 40: NIT: The following line could be interpreted under the new `### Power-off-based remediation` subsection.

It might be better to have a new subsection for  `fartemplate` so this line won't be considered as only associated with power-off remediation.

kattz-kawa:
@slintes 
> Patching the machine for forcing a reconcile... hm, not sure, might be worth a try.

I’ve just committed an implementation that patches the Machine to force a reconcile. 
Could you please take a look and review it?

In my environment, all tests are passing, and the remediation finish time has improved. Here are results:

```
2025-08-20T08:28:00.672750166+09:00     INFO    Remediation was finished        {"remediation time": "3m0.79322408s"}
2025-08-20T08:28:00.672815001+09:00     INFO    Remediation was finished        {"remediation time": "1m59.614899288s"}
2025-08-20T08:28:00.672827852+09:00     INFO    Remediation was finished        {"remediation time": "3m10.676062952s"}
2025-08-20T08:28:00.672849274+09:00     INFO    Remediation was finished        {"remediation time": "1m51.687023791s"}
2025-08-20T08:28:00.67285381+09:00      INFO    Remediation was finished        {"remediation time": "1m36.581419788s"}
2025-08-20T08:28:00.672858347+09:00     INFO    Remediation was finished        {"remediation time": "1m2.099814578s"}
2025-08-20T08:28:00.672862562+09:00     INFO    Remediation was finished        {"remediation time": "1m38.030299522s"}
2025-08-20T08:28:00.672867499+09:00     INFO    Remediation was finished        {"remediation time": "58.323772848s"}
(snip)
[AfterSuite] PASSED [0.000 seconds]
------------------------------

Ran 8 of 8 Specs in 1577.180 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

test/e2e/far_e2e_test.go
```diff
@@ -403,13 +448,22 @@ func verifyStatusCondition(nodeName, conditionType string, conditionStatus *meta
...
+	if testAction == "reboot" {
+		By("Getting new node's boot time")
+		verifyNodeRebooted(nodeName, nodeBootTimeBefore)
+	} else if testAction == "off" {
+		By("Check if the node powered off")
+		verifyNodePoweredOff(nodeName)
+		// After detecting the node power-off, the Pod may not be immediately deleted.
+		// Calling checkPodDeleted too early can lead to fail test.
+		// Therefore, we wait until the Pod is removed before verifying its deletion status.
+		waitForPodDeletion(pod)
```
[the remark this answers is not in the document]
razo7: >This looks fishy to me. Why would there be a difference to the reboot case? I don't remember that the test ever failed at checkPodDeleted(). @razo7 do you?

I don't remember as well

>However, I'm wondering why checkPodDeleted uses Consistently, and not Eventually? 🤔 @razo7 do you remember why?

No, and it makes sense to use Eventually

test/e2e/utils/cluster.go
```diff
@@ -120,3 +132,95 @@ func GetBMHNodeInfoList(machineClient *machineclient.Clientset) (map[v1alpha1.No
...
+	return "", fmt.Errorf("no machine found for node name %q", nodeName)
+}
+
+// TriggerReconcileMachine forces a reconciliation of the target Machine object
+// by updating its metadata.labels with a unique value.
+//
+// GetPowerStatus retrieves the node's power status by querying the corresponding Machine object.
+// To ensure the status is current, this function first triggers a reconciliation of the Machine resource.
+// This helps speed up synchronization with the machine-api controller before reading the providerStatus.
+
```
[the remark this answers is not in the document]
kattz-kawa: I've updated.

test/e2e/utils/cluster.go
```diff
@@ -120,3 +132,95 @@ func GetBMHNodeInfoList(machineClient *machineclient.Clientset) (map[v1alpha1.No
...
+// Using the machine resource provides a simpler and safer way to check power status.
+func GetPowerStatus(machineClient *machineclient.Clientset, targetNodename string) (string, error) {
+	//  oc get machine -n openshift-machine-api MACHINE_NAME -o jsonpath='{.status.providerStatus.instanceState}'
+
+	machinename, err := FindMachineByNodeName(machineClient, targetNodename)
+	if err != nil {
+		return "", fmt.Errorf("failed to get machine name %s: %w", targetNodename, err)
+	}
+	// Trigger reconcile for status
+	TriggerReconcileMachine(machineClient, machinename)
```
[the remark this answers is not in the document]
kattz-kawa: I've updated.

PR #177, opened 2025-07-24
Adding validations to FAR Template webhook 

mshitrit:
#### Why we need this PR
Currently, misconfigurations in fence agent parameters (e.g., incorrect IP addresses, invalid credentials, or improper fence_agent arguments) are easily made and can go unnoticed until a fencing operation is actually required. Discovering such issues during a critical event when fencing is desperately needed is too late and can lead to severe cluster instability or outages.

To mitigate this risk, we propose implementing a validating admission webhook for fencing parameters

#### Changes made
Validating Params, Secrets and status upon template creation/change.

[RHWA-65](https://issues.redhat.com//browse/RHWA-65)

## Summary by CodeRabbit

* **New Features**
  * Node-name derivation and centralized fence-agent parameter assembly with secret templating
  * Action defaulting to "reboot" when unspecified
  * Admission validation routed through a dedicated validator

* **Bug Fixes**
  * Stricter action validation with clearer errors
  * Detects duplicate or missing parameter definitions and missing secrets
  * Improved error aggregation and reporting for template rendering

* **Tests**
  * Expanded unit and e2e tests covering templates, actions, secrets, validator flows

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


mshitrit:
@coderabbitai full review

api/v1alpha1/fenceagentsremediationtemplate_validator.go
```diff
@@ -0,0 +1,316 @@
...
+
+	// Collect all unique node names from NodeParameters and NodeSecretNames
+	nodeNames := getNodeNamesFromSpec(spec)
+
+	skipStatusValidation := false
+	// If no node-specific parameters, validate with shared parameters only, use a dummy placeholder for node name
+	if len(nodeNames) == 0 {
+		webhookTemplateValidatorLog.Info("validateFenceAgentTemplate no nodes found")
+		nodeNames["temp-validation"] = true
+		// Status validation will NOT occur for shared params with a node template (because we want to avoid getting all the nodes from the API server)
```
clobrano, line 149: If we have no node-specific parameters, can we consider picking 1 Node only and perform validation on that one?

api/v1alpha1/fenceagentsremediationtemplate_validator.go
```diff
@@ -0,0 +1,316 @@
...
+
+	// Validate shared parameters
+	for paramName, paramVal := range sharedParameters {
+		// Verify action must be reboot
+		if err := validateActionParameter(string(paramName), paramVal); err != nil {
+			return err
+		}
+		// Verify param isn't already defined
+		if existingParams[paramName] {
+			err := errors.New(errorParamDefinedMultipleTimes)
```
clobrano, line 221: We might want to add the duplicated parameter name also in the error, as it will appear in the user's terminal and not only in the logs.

api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
```diff
@@ -1,20 +1,94 @@
...
 
+// mockClient for testing
+type mockClient struct {
+	client.Client
+}
+
+// Implement Get method to handle secret retrieval in tests
+func (m *mockClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+	// Return a pre-built secret for testing duplicate parameters
+	if key.Name == "test-node-secret" && key.Namespace == "test-namespace" {
```
clobrano, line 26: If the intention is to explicitly create a failure case, it would be probably better to give a distinguished name to the `key`, something referring to the wanted error, like "test-node-secret-ip-conflict" for example.

api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
```diff
@@ -1,20 +1,94 @@
...
+
+func (m *MockCommandExecutor) RunCommand(ctx context.Context, name string, args ...string) (string, string, error) {
+	fullCmd := append([]string{name}, args...)
+	m.Commands = append(m.Commands, fullCmd)
+
+	// Match command pattern and return predefined response
+	cmdStr := strings.Join(fullCmd, " ")
+	if response, exists := m.Responses[cmdStr]; exists {
+		return response.Stdout, response.Stderr, response.Err
+	}
```
clobrano, line 63: Nice idea!

controllers/fenceagentsremediation_controller.go
```diff
@@ -241,7 +228,7 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		}
 
 		cmd := append([]string{far.Spec.Agent}, mapToSliceConvert(faParams)...)
-		r.Log.Info("Execute the fence agent", "Fence Agent", far.Spec.Agent, "Node Name", node.Name, "FAR uid", far.GetUID(), "Parameters", maps.Keys(faParams))
+		r.Log.Info("Execute the fence agent", "Fence Agent", far.Spec.Agent, "Node Name", node.Name, "FAR uid", far.GetUID(), "Parameters", slices.Collect(maps.Keys(faParams)))
```
clobrano, line 231: Is it ok to use a feature of golang 1.23?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -648,6 +650,52 @@ var _ = Describe("FAR Controller", func() {
 				verifyEvent(corev1.EventTypeNormal, utils.EventReasonRemoveOutOfServiceTaint, utils.EventMessageRemoveOutOfServiceTaint)
 			})
 		})
+
+		When("A FAR remediation that was timed out by NHC is deleted", func() {
```
clobrano, line 654: Is this part of this PR?

clobrano (changes requested):
I left some comments, but overall great PR :+1:

api/v1alpha1/fenceagentsremediationtemplate_validator.go
mshitrit: If we do that we'll need to get it from the API server instead of the template.
I'm not sure how I feel about it.
It seems like it exceeds template validation and I also worry about picking random node which might never intended to be remediated.

controllers/fenceagentsremediation_controller_test.go
mshitrit: Good catch. Those test belong to another PR.

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,225 @@
...
+)
+
+var (
+	// paramsLog is for logging in this package.
+	paramsLog = logf.Log.WithName("fenceagentsremediation-params")
+)
+
+// BuildFenceAgentParams collects the FAR's parameters for the node based on FAR CR, and if the CR is missing parameters
+// or the CR's name don't match nodeParameter name, or it has an action which is different from reboot, then return an error
+func BuildFenceAgentParams(ctx context.Context, k8sClient client.Client, far *FenceAgentsRemediation) (map[ParameterName]string, bool, error) {
```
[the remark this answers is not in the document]
mshitrit: In my original dev cycle I've put it in a different package (`pkg/validation`).
The downside of that approach that it would prevent us using API resource because it'll cause cyclic dependency.

In practice it means that  `BuildFenceAgentParams` can't receive `far *FenceAgentsRemediation`  and need to receive bunch of other params that are extracted from it instead.

IMO both solution aren't ideal but this one is slightly better, I don't have a strong preference though.

Let me know if:
a. You think of a better solution I didn't consider
b. Still have a strong preference to move the code

api/v1alpha1/fenceagentsremediationtemplate_validator.go
```diff
@@ -0,0 +1,316 @@
...
+	// Parameter validation constants shouldn't exceed 13 seconds ocp cap (https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/architecture/admission-plug-ins)
+	parameterValidationTimeout = 3 * time.Second
+)
+
+var (
+	// webhookTemplateValidatorLog is for logging in this package.
+	webhookTemplateValidatorLog = logf.Log.WithName("fenceagentsremediationtemplate-validator")
+)
+
+// +kubebuilder:webhook:path=/validate-fence-agents-remediation-medik8s-io-v1alpha1-fenceagentsremediationtemplate,mutating=false,failurePolicy=fail,sideEffects=None,timeoutSeconds=13,groups=fence-agents-remediation.medik8s.io,resources=fenceagentsremediationtemplates,verbs=create;update,versions=v1alpha1,name=vfenceagentsremediationtemplate.kb.io,admissionReviewVersions=v1
```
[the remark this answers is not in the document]
mshitrit: I tried to explain it [here](https://github.com/medik8s/fence-agents-remediation/pull/177/files#diff-b1de6b717be2792b4f8ac3b59df5ae8432925aca13980f2f332199d043069994R39-R40)

Probably a good idea to put a comment here as well

api/v1alpha1/fenceagentsremediationtemplate_validator.go
```diff
@@ -0,0 +1,316 @@
...
+	var warnings []string
+	spec := &r.Spec.Template.Spec
+
+	// Check if template has any parameters at all
+	hasSharedParams := len(spec.SharedParameters) > 0
+	hasNodeParams := len(spec.NodeParameters) > 0
+	hasSecrets := spec.SharedSecretName != nil || spec.NodeSecretNames != nil
+
+	// If template has no parameters or secrets, skip parameter validation
+	// Templates are allowed to be empty - parameters can be added later
```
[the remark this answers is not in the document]
mshitrit: Yep I agree.
I'm also leaning into failing the validation here.

api/v1alpha1/fenceagentsremediationtemplate_validator.go
```diff
@@ -0,0 +1,316 @@
...
+	//TODO mshitrit should we allow this ?
+	if !hasSharedParams && !hasNodeParams && !hasSecrets {
+		webhookTemplateValidatorLog.Info("validateFenceAgentTemplate return no params")
+		return warnings, nil
+	}
+
+	// Collect all unique node names from NodeParameters and NodeSecretNames
+	nodeNames := getNodeNamesFromSpec(spec)
+
+	skipStatusValidation := false
```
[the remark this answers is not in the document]
mshitrit: This use case is just saying that even though nodeparams weren't found, we should continue the validation on the account of shared param may exist.

api/v1alpha1/fenceagentsremediationtemplate_validator.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+	}
+	for nodeName, _ := range spec.NodeSecretNames {
+		nodeNames[string(nodeName)] = true
+	}
+	return nodeNames
+}
+
+// validateFenceAgentParams validates all fence agent parameters without building the map
+func validateFenceAgentParams(
```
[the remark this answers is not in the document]
mshitrit: The idea was to separate the build method from the validation method.
The caveat is that in order to validate we need to iterate on the elements to be verified.
The iteration process would also exist in the build logic which I assume is the duplication you are pointing at.

I'm fine with merging those together.

api/v1alpha1/fenceagentsremediationtemplate_validator.go
clobrano: If there is a selection of nodes not indented for remediation, there wouldn't be a list of nodes intended to remediate in the template?

api/v1alpha1/fenceagentsremediationtemplate_validator.go
```diff
@@ -0,0 +1,316 @@
...
+	result := &ParameterValidationResult{
+		IsSuccessful: true,
+		Message:      "",
+	}
+
+	// Build command with status action
+	command := []string{agent, parameterActionName, parameterActionStatusValue}
+
+	// Add parameters (excluding action parameters to avoid conflicts)
+	for paramName, paramValue := range parameters {
```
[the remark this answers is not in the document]
mshitrit: I think in this case it's a bit more tricky.
The premise is that we have some build/validation logic which is shared both when creating a remediation and when webhook validation is triggered.
However this specific logic (i.e checking fence agent status) is meant only for the webhook validation.

mshitrit:
> I would really like to see an e2e test for this :)

I think we get a decent cover from the current e2e test, as this code triggered from the template creation.

pkg/executor/executor.go
```diff
@@ -0,0 +1,25 @@
+package executor
```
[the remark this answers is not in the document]
mshitrit: I definitely understand the motivation.

I think the problem is that we would greatly increase the chance of a webhook timeout by doing so (particularly due to the 13s time limit).

Running into the limit is mostly an issue because the current implementation assumes status failure is non blocking (issue a warning) and a webhook timeout will cause a blocking error.

IMO it makes sense for the status check to not be blocking but we can discuss this if you disagree.

Let me know your thoughts.

clobrano (commented):
I left one small comment, otherwise lgtm for me

clobrano (approved):
/lgtm

Giving the chance to close any other open thread

/hold

mshitrit:
>Giving the chance to close any other open thread

//cc @slintes

api/v1alpha1/fenceagentsremediationtemplate_validator.go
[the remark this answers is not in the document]
mshitrit: >Somehow at least one parameter has to contain something node specific

Sure, this validation makes sense.

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,461 @@
...
+)
+
+const (
+	parameterActionRebootValue     = "reboot"
+	actionName                     = "action"
+	parameterActionName            = "--" + actionName
+	parameterActionStatusValue     = "status"
+	errorParamDefinedMultipleTimes = "invalid multiple definition of FAR parameter, parameter name: %s"
+	errorMissingParams             = "nodeParameters or sharedParameters or both are missing, and they cannot be empty"
+	// Parameter validation constants shouldn't exceed 13 seconds ocp cap (https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/architecture/admission-plug-ins)
```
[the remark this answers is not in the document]
mshitrit: The idea is that we have a total of 13 seconds for all of the validations.
Each status validation is limited to max 3 seconds.
I'll try to improve the phrasing to reflect that.

mshitrit:
> Thinking more about that 13s timeout, that's a serious issue: assuming a status command needs half a second (random assumption, probably many agents are faster, but I guess some can be even slower), we manage to check <26 nodes only... thoughts?
> 
> Not sure if running the checks concurrently will introduce another set of issues, depending on fence agent type.... 🤔

I have a pretty simple idea.
In a nutshell if we run into a risk of a timeout, skip the rest of the status validations and issue warning.
That way we make sure we don't block the user.

To clarify I'll use an example:

**Let's say we have 10+ nodes we are validating**
**Each status validation is limited to 3 seconds**

**Flow**
- First node validation takes 2 second to validate (11 sec left out of original 13)
- 2nd Node takes 0.5 seconds to validate (10.5 sec left)
- 3rd Node takes 2.5 sec .. (8 sec left)
- 4th Node to 7th takes 1 sec each (4 sec left)
- Since 4 sec >=  3 limit we can run another validation
- 8th Node takes 2.5 sec (1.5 sec left)
- Since 1.5 < 3 9th+ node aren't validated and we return a warning. 

We can also make sure that the status validations runs after all other validations in order to make sure those are the only ones we'll skip.

WDYT ?

mshitrit:
> you just need to add a test which uses wrong credentials and ensure that validation fails?

Assuming the focus is the Status check, the validation will not fail since it's non blocking (only issues a warning).

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,382 @@
...
+func (v *customValidator) validateFenceAgentTemplate(ctx context.Context, r *FenceAgentsRemediationTemplate) error {
+	spec := &r.Spec.Template.Spec
+
+	// Check if template has any parameters at all
+	hasSharedParams := len(spec.SharedParameters) > 0
+	hasNodeParams := len(spec.NodeParameters) > 0
+	hasSecrets := spec.SharedSecretName != nil || spec.NodeSecretNames != nil
+
+	// If template has no parameters or secrets, template is considered invalid
+	if !hasSharedParams && !hasNodeParams && !hasSecrets {
```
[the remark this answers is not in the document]
mshitrit: Not sure I follow.
This is how I understand this code:
- this validation is basically verifying that we fail in case the required fencing params do not exist.
- Those params (or at least how to get them in case they are in secrets) are defined in the FAR Template.
- If no params exist we consider the template as invalid and force the user to update it.
(I think we've discussed it [here](https://github.com/medik8s/fence-agents-remediation/pull/177/#discussion_r2232836597)).

Are you suggesting that we should wait before a remediation is created before validating this ?

controllers/fenceagentsremediation_controller_test.go
```diff
@@ -649,51 +649,6 @@ var _ = Describe("FAR Controller", func() {
 				verifyEvent(corev1.EventTypeNormal, utils.EventReasonRemoveOutOfServiceTaint, utils.EventMessageRemoveOutOfServiceTaint)
 			})
 		})
-
-		When("A FAR remediation that was timed out by NHC is deleted", func() {
```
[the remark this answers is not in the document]
mshitrit: I think you give me too much credit here.
Rebase was indeed used, I suspect the mistake that got this code here was much dumber.

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,382 @@
...
+		err := errors.New(errorMissingParams)
+		paramsLog.Error(err, "Missing parameters")
+		return err
+	}
+
+	// Collect all unique node names from NodeParameters and NodeSecretNames
+	nodeNames := getNodeNamesFromSpec(spec)
+
+	// If no node-specific parameters, validate with shared parameters only, use a dummy placeholder for node name
+	if len(nodeNames) == 0 {
```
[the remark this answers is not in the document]
mshitrit: I agree.
It's validated later on in [here](https://github.com/medik8s/fence-agents-remediation/pull/177/files#diff-63460000082bd7d3eccded16862e1f23d389e2c6da1aedaa1af395f9c8e0c103R231-R236).
Since in order to validate this, we need to get the Secrets it made sense to do it a later context  when we fetch the secrets in any case, in order to save code duplication.

api/v1alpha1/fenceagentsremediation_params.go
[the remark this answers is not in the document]
mshitrit: >just like it was before?

Maybe I'm missing something but I think this validation (`validateFenceAgentTemplate`) is new to this PR and the older validation (`validateFAR`) remained in `fenceagentsremediation_webhook.go`

>why not adding this new validation to the FAR validation

The way I see it, even though the spec exist both in the FART and in FAR, validations that are specific for FAR CR create/update belongs in  `fenceagentsremediation_webhook.go` and those that are relevant for FART CR create/update belongs in  `fenceagentsremediationtemplate_webhook.go`.

In this specific case `validateFenceAgentTemplate` isn't part of `fenceagentsremediationtemplate_webhook.go` because  I separated the param logic (collecting and validation) in to `fenceagentsremediation_params.go`  

I'm still not 100% clear on what you are suggesting
Do you think that:  
- `validateFenceAgentTemplate` should move to `fenceagentsremediationtemplate_webhook.go`  
- `validateFenceAgentTemplate` should move to `fenceagentsremediation_webhook.go`
- `fenceagentsremediation_params.go`  deleted and merged into `fenceagentsremediationtemplate_webhook.go`  
- something else ?

clobrano (approved):
/lgtm

It seems there are other threads to close

/hold

mshitrit:
failure is unrelated to code change, retesting
/test 4.19-openshift-e2e

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,389 @@
...
+func (v *customValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	metaObj, _ := obj.(metav1.Object)
+	paramsLog.Info("validate delete", "name", metaObj.GetName())
+	return nil, nil
+}
+
+func (v *customValidator) validate(ctx context.Context, new runtime.Object) (admission.Warnings, error) {
+	spec := v.getSpec(new)
+
+	// Skipping validation because must be either a FenceAgentsRemediationTemplate or a FenceAgentsRemediation
```
slintes: I don't understand this comment?

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,389 @@
...
+		spec = far.Spec
+	}
+	return &spec
+}
+
+// validateFenceAgentForNodes validates fence agent parameters for all the nodes defined in the spec
+// by creating temporary FAR CRs and using BuildFenceAgentParams
+func validateFenceAgentForNodes(ctx context.Context, k8sClient client.Client, namespace string, spec *FenceAgentsRemediationSpec) error {
+
+	// Check if template has any parameters at all
```
slintes: nit: it's not about FAR templates only

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,389 @@
...
+	for _, nodeMap := range spec.NodeParameters {
+		for nodeName := range nodeMap {
+			nodeNamesMap[string(nodeName)] = true
+		}
+	}
+	for nodeName := range spec.NodeSecretNames {
+		nodeNamesMap[string(nodeName)] = true
+	}
+
+	var nodeNames []string
```
slintes: return slices.Collect(maps.Keys(nodeNamesMap))

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,389 @@
...
+			return fenceAgentParams, err
+		}
+		isNodeTemplateExistInSharedParams = isNodeTemplateExistInSharedParams || paramVal != processedParamVal
+		fenceAgentParams[paramName] = processedParamVal
+	}
+
+	// Validate and add node parameters (these can override shared parameters)
+	for paramName, nodeMap := range far.Spec.NodeParameters {
+		if nodeVal, isFound := nodeMap[NodeName(nodeName)]; isFound {
+			// Verify action must be reboot
```
slintes: outdated

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,389 @@
...
+	return nodeNames
+}
+
+// validateFenceAgentParams builds the fence agent parameters map with validation
+func validateFenceAgentParams(far *FenceAgentsRemediation, secretParams SecretParams) (map[ParameterName]string, error) {
+	nodeName := GetNodeName(far)
+	fenceAgentParams := make(map[ParameterName]string)
+
+	// Track parameter names for uniqueness validation
+	existingParams := make(map[ParameterName]bool)
```
slintes: why do we need this? AFAICS it will always contain the same names as fenceAgentParams...?

api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
```diff
@@ -29,17 +68,107 @@ var _ = Describe("FenceAgentsRemediationTemplate Validation", func() {
...
+					},
+				}
+
+				warnings, err := validator.ValidateCreate(ctx, farTemplate)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(warnings).To(BeEmpty())
+			})
+		})
+
+		When("template has no shared parameters and no node parameters", func() {
```
razo7:
```suggestion
		When("farTemplate has no shared parameters and no node parameters", func() {
```

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,416 @@
...
+)
+
+const (
+	parameterRebootActionValue     = "reboot"
+	parameterOffActionValue        = "off"
+	actionName                     = "action"
+	parameterActionName            = "--" + actionName
+	errorParamDefinedMultipleTimes = "invalid multiple definition of FAR parameter, parameter name: %s"
+	errorMissingParams             = "invalid template: mandatory parameters are missing"
+	ErrorUnsupportedAction         = "FAR doesn't support any other action than reboot or off"
```
razo7: NIT: Maybe add an apostrophe to distinguish that this word (reboot/off) is an actual field value?

```suggestion
	ErrorUnsupportedAction         = "FAR doesn't support any other action than `reboot` or `off`"
```

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,416 @@
...
+func (v *customValidator) validateFenceAgentForNodes(ctx context.Context, namespace string, spec *FenceAgentsRemediationSpec) error {
+
+	// Check if spec has any parameters at all
+	hasSharedParams := len(spec.SharedParameters) > 0
+	hasNodeParams := len(spec.NodeParameters) > 0
+	hasSecrets := spec.SharedSecretName != nil || spec.NodeSecretNames != nil
+
+	// If template has no parameters or secrets, template is considered invalid
+	if !hasSharedParams && !hasNodeParams && !hasSecrets {
+		err := errors.New(errorMissingParams)
```
razo7: IIUC validateFenceAgentForNodes will be used for far and farTemplate, so we should differentiate here the logic, since NodeTemplate is only used or relevant for farTemplate.
  between the echecks not be related only to the template (`invalid template: mandatory parameters are missing`)

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	return fenceAgentParams, nil
+}
+
+// validateFenceAction validates that action parameters are set correctly
+func validateFenceAction(paramName, paramVal string) error {
+	if (paramName == actionName || paramName == parameterActionName) &&
+		(paramVal != "" && paramVal != parameterRebootActionValue && paramVal != parameterOffActionValue) {
+		// --action parameter with a different value from reboot is not supported
```
razo7:
```suggestion
		// --action parameter with a different value than `reboot` or `off` is not supported
```

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,416 @@
...
+	// Validate and add secret parameters
+	for secretKey, secretVal := range secretParams.params {
+		secretParam := ParameterName(secretKey)
+		// Verify action must be reboot or off
+		if err := validateFenceAction(string(secretParam), secretVal); err != nil {
+			return nil, err
+		}
+		if _, exist := fenceAgentParams[secretParam]; exist {
+			err := fmt.Errorf(errorParamDefinedMultipleTimes, secretParam)
+			paramsLog.Error(err, "can't build fence agents parameters a parameter is defined multiple times", "parameter name", secretParam)
```
razo7: IMHO, a reasoning word is missing. Maybe store it as a var in the beginning, like `errorParamDefinedMultipleTimes`
```suggestion
			paramsLog.Error(err, "can't build fence agents parameters, when a parameter is defined multiple times", "parameter name", secretParam)
```

test/e2e/far_e2e_test.go
```diff
@@ -142,31 +151,78 @@ var _ = Describe("FAR E2e", func() {
...
 		})
+
+		When("Trying to create or update to an invalid FAR/T CR", func() {
+			BeforeEach(func() {
+				// since we test here the webhook we are skipping remediation creation in order to make sure failed nodes wouldn't affect the tested webhooks
+				skipRemediationCreation = true
+				testShareParam = addSecretsToSharedParams(testShareParam)
+			})
+			It("it should fail", func() {
+				// eventually block used to avoid update conflict
```
razo7: eventually?
I don't think this comment is related here.

razo7 (commented):
I left many NITs to review, even though the PR overall looks good 😸

api/v1alpha1/fenceagentsremediationtemplate_webhook_test.go
```diff
@@ -1,26 +1,65 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// getFuncNodeSecretIpConflict returns the default Get function behavior for secrets
+func getFuncNodeSecretIpConflict() func(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+	return func(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+		// Default behavior - Return a pre-built secret for testing duplicate parameters
+		if key.Name == "test-node-secret-ip-conflict" && key.Namespace == "test-namespace" {
```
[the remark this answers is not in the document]
mshitrit: good point, but not really part of this PR . I'll update

api/v1alpha1/fenceagentsremediation_params.go
mshitrit: Fair enough, since basically we check the spec here that is shared for both Template and CR I've updated the error to reflect that.

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,416 @@
...
+		paramsLog.Error(err, "Missing parameters")
+		return nil, err
+	}
+
+	return fenceAgentParams, nil
+}
+
+// validateFenceAction validates that action parameters are set correctly
+func validateFenceAction(paramName, paramVal string) error {
+	if (paramName == actionName || paramName == parameterActionName) &&
```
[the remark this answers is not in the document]
mshitrit: This was discussed and implemented in another PR, IIRC it was a concern Marc brought up.

api/v1alpha1/fenceagentsremediation_params.go
```diff
@@ -0,0 +1,416 @@
...
+			secretParams[paramName] = processedParamVal
+		}
+	}
+
+	// collect secret params from the node's secret
+	nodeSecretName, isFound := nodeSecretNames[NodeName(nodeName)]
+	if isFound {
+		nodeSecretParams, err := collectSecretParams(ctx, k8sClient, nodeSecretName, namespace, false) // false = isSharedSecret
+		if err != nil {
+			return SecretParams{nil, hasNodeTemplate}, err
```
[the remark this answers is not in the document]
mshitrit: No, `hasNodeTemplate` refers to first part where we checked the shared params, you are referning to an error that might have happened during the collection of the node params.

controllers/fenceagentsremediation_controller.go
```diff
@@ -128,13 +114,13 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 
 	// Validate FAR CR name to match a nodeName from the cluster
 	r.Log.Info("Check FAR CR's name")
-	node, err := utils.GetNodeWithName(r.Client, getNodeName(far))
+	node, err := utils.GetNodeWithName(r.Client, v1alpha1.GetNodeName(far))
```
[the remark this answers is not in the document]
mshitrit: Maybe, but ATM `v1alpha1`  is the convention we are using in all of our operators. I'm not thrilled to change it as part of this PR.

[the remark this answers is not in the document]
razo7: >  ATM v1alpha1 is the convention we are using in all of our operators.

Yes, for API stuff like in Spec, Status, or taints, but not for interior functions. I understand the compromise here with the cyclic dependency from another thread that Marc opened, but it just seems too odd for me. We can indeed discuss it in another PR

razo7:
/lgtm
Giving others one more chance to review
/hold

PR #178, opened 2025-07-31
Add podAntiAffinity to controller-manager to improve HA

kattz-kawa:
#### Why we need this PR
In the current deployment configuration, multiple replicas of the controller-manager can be scheduled on the same node. This creates a single point of failure. This setup does not fully utilize the benefits of having multiple replicas. With better placement, we can use the replicas more effectively.

#### Changes made

- Added `podAntiAffinity` rules to controller-manager in "config/manager/manager.yaml"
- Ensures that pods with the same labels (`control-plane=controller-manager`, `app.kubernetes.io/name=fence-agents-remediation-operator`) are not scheduled on the same node
- Uses `requiredDuringSchedulingIgnoredDuringExecution` with `topologyKey: kubernetes.io/hostname`

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -69,3 +69,17 @@ spec:
             memory: 64Mi
       serviceAccountName: controller-manager
       terminationGracePeriodSeconds: 10
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - labelSelector:
+                matchExpressions:
+                  - key: control-plane
+                    operator: In
+                    values:
+                      - controller-manager
+                  - key: app.kubernetes.io/name
+                    operator: In
+                    values:
+                      - fence-agents-remediation-operator
+              topologyKey: "kubernetes.io/hostname"


clobrano:
What is your reasoning to use `requiredDuringSchedulingIgnoredDuringExecution`?

In cases we only have 1 node matching the rule, the other replica will stay pending with no real reason, IIUC. With `preferredDuringSchedulingIgnoredDuringExecution`, we should reach the same goal when we have multiple nodes to schedule the replica on, and we can handle the previous use case as well.

clobrano:
According to https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/, `preferredDuringSchedulingIgnoredDuringExecution` only fails if "a matching node is not available", so the scenario you described is covered. 
We need to decide if we prefer two replicas on the _only available_ node, when it happens, or one replica in forever pending state in the same scenario.

mshitrit:
>Which issue(s) this PR fixes

Please attach the relevant Jira ticket

mshitrit:
Changes are missing in the CSV (just run `make full-gen`)

mshitrit:
> We need to decide if we prefer two replicas on the _only available_ node, when it happens, or one replica in forever pending state in the same scenario.

No strong preference on my part, I am leaning to ` two replicas on the _only available_ node`  though, just because it seems like it adds just a bit more stability for this edge use case.

kattz-kawa:
> but... taking a step back: "multiple replicas of the controller-manager can be scheduled on the same node"
Does this happen actually? Doesn't the default scheduler spread pods automatically across nodes already these days?


Yes, this can happen.
In production environments, one common situation is when all OpenShift cluster nodes must be temporarily shut down.
For example, during a planned power outage in an on-premises setup.

Let’s imagine a user who has configured the controller-manager to run only on Infra nodes using `nodeSelector: Infra` setting via OLM. After the outage, the user carefully brings up each node one by one, checking the health of each Infra node before starting the next.
In this case, if only first Infra node is ready at the beginning, both replicas of the controller-manager may be scheduled on that single node.
This may sound like a rare or extreme case, but it’s just one example of a situation where this behavior can occur, if the configuration allows it.
One of the main goals of this PR is to prevent such behavior from happening without the user noticing it.

kattz-kawa:
@mshitrit
> Please attach the relevant Jira ticket
> Changes are missing in the CSV (just run make full-gen)

Thank you for the feedback.
I've created the relevant Jira ticket and linked it in the PR description.
Also, I've run make full-gen and committed the updated CSV to reflect the changes in the controller-manager deployment. Hope this addresses the feedback.

@slintes
> So I think this boils down to: does requiredDuringScheduling always work with OLM...

Thank you for your comment. I’d like to share some findings from our recent tests regarding the behavior of OLM when only one Infra nodes are available.

In our environment, we observed that the CSV entered a Failed phase when there was only one Infra node. However, this was expected. We confirmed that uninstalling the CSV in this state worked successfully. After restoring three Infra nodes to Ready, we were able to reinstall the operator without issues.


Additionally, when the CSV was in Failed state, then we recovered one Infra node, the CSV automatically transitioned to Succeeded without any manual operation. This suggests that OLM retries the installation process.

So far, we have not observed any side effects or OLM-specific parameters that would be negatively affected by using `requiredDuringSchedulingIgnoredDuringExecution`.

Please let me know if this testing is enough, or if there are any other points I may have missed. We’ll continue checking similar test in [NHC PR#377](https://github.com/medik8s/node-healthcheck-operator/pull/377) as well.

kattz-kawa:
@slintes I've updated this PRs to use `preferredDuringScheduling` as discussed. Please let me know if you have any further feedback or if anything needs adjustment.

PR #200, opened 2026-04-16
[WIP] Add Automatic remediation strategy

Tzif-Morgen:
#### Why we need this PR

Implement automatic remediation strategy selection based on Kubernetes version.

#### Changes made

- Add AutomaticRemediationStrategy constant and set as default
- Add version detection for OutOfServiceTaint support (>=1.26) and GA (>=1.28)
- Add getRuntimeStrategy() to select strategy at runtime based on k8s version
- Remove OutOfServiceTaintValidator from validation package (replaced by utils approach)
- Move and enhance version detection tests to pkg/utils (test both Supported and GA flags)
- Update CRD manifests with Automatic enum and default

The Automatic strategy selects OutOfServiceTaint on k8s >=1.28 (GA), otherwise ResourceDeletion.

Fixes [RHWA-690](https://redhat.atlassian.net/browse/RHWA-690)

## Summary by CodeRabbit

* **New Features**
  * Added "Automatic" remediation strategy that auto-selects the best approach at runtime.

* **Configuration Changes**
  * CRD default for remediationStrategy changed to "Automatic"; allowed values updated to include it.

* **Improvements**
  * Runtime strategy resolution and centralized retry-based taint-support detection for more reliable startup and remediation.

* **Tests**
  * Test suites updated to use shared taint feature flags and common taint utilities.

* **Chores**
  * Removed legacy out-of-service taint validator and deprecated helpers/tests.


Tzif-Morgen:
Will check ocp supported versions

mshitrit:
@coderabbitai review

mshitrit:
looks like csv wasn't regenerated.
you can run `make full-gen` to fix this

pkg/utils/taints.go
```diff
@@ -117,3 +139,58 @@ func RemoveTaint(r client.Client, nodeName string, taint corev1.Taint) error {
...
+		loggerTaint.Error(err, "couldn't parse k8s minor version", "minor version", version.Minor)
+		return err
+	}
+
+	IsOutOfServiceTaintSupported = majorVer > minK8sMajorVersionOutOfServiceTaint || (majorVer == minK8sMajorVersionOutOfServiceTaint && minorVer >= minK8sMinorVersionSupportingOutOfServiceTaint)
+	loggerTaint.Info("out of service taint strategy", "isSupported", IsOutOfServiceTaintSupported, "k8sMajorVersion", majorVer, "k8sMinorVersion", minorVer)
+	IsOutOfServiceTaintGA = majorVer > minK8sMajorVersionOutOfServiceTaint || (majorVer == minK8sMajorVersionOutOfServiceTaint && minorVer >= minK8sMinorVersionGAOutOfServiceTaint)
+	loggerTaint.Info("out of service taint strategy", "isGA", IsOutOfServiceTaintGA, "k8sMajorVersion", majorVer, "k8sMinorVersion", minorVer)
+	return nil
+}
```
mshitrit: I think this shared (SNR>FAR) logic block is big enough so it would make sense to put it in common

pkg/validation/validation.go
```diff
@@ -4,30 +4,8 @@ import (
...
-	leadingDigits    = regexp.MustCompile(`^(\d+)`)
-)
-
-const (
-	//out of service taint strategy const (supported from 1.26)
-	minK8sMajorVersionOutOfServiceTaint = 1
-	minK8sMinorVersionOutOfServiceTaint = 26
 )
 
-type OutOfServiceTaintValidator struct {
```
mshitrit, line 27: why switch out the validator pattern ?

pkg/utils/taint_test.go
```diff
@@ -56,3 +58,40 @@ func getControlPlaneRoleTaint() corev1.Taint {
 		Effect: corev1.TaintEffectNoExecute,
 	}
 }
+
+func Test_setOutOfTaintFlags(t *testing.T) {
```
mshitrit: Nit: I think it would be nicer to check the whole remediation process similar to how we do it in the [controller tests](https://github.com/medik8s/fence-agents-remediation/blob/main/controllers/fenceagentsremediation_controller_test.go) instead of checking specific method.

pkg/utils/taints.go
Tzif-Morgen: agree, unless we decide to remove the check of 1.26 version (for SNR we still need to check it)

main.go
```diff
@@ -107,15 +108,13 @@ func main() {
...
+	// Detect OutOfService taint support for runtime strategy selection and webhook validation
+	taintInfo, err := commonTaints.DetectOutOfServiceTaintInfoWithRetry(context.Background(), mgr.GetConfig())
 	if err != nil {
-		setupLog.Error(err, "unable to verify Kubernetes version for indicating the out-of-service taint support. out-of-service taint isn't supported")
-	}
-	isOutOfServiceTaintSupported := outOfServiceTaintValidator.IsOutOfServiceTaintSupported()
-	if isOutOfServiceTaintSupported {
-		setupLog.Info("out-of-service taint is supported on this cluster")
+		setupLog.Error(err, "unable to detect out-of-service taint support")
+		os.Exit(1)
```
slintes, line 115: do we really want to fail FAR completely when we're not able to check if the OOS taint is supported? 🤔

main.go
```diff
@@ -107,15 +108,13 @@ func main() {
...
 	}
-	fenceagentsremediationv1alpha1.InitOutOfServiceTaintSupportedFlag(isOutOfServiceTaintSupported)
+	fenceagentsremediationv1alpha1.OutOfServiceInfo = taintInfo
```
slintes: Please avoid introducing global public vars. They are bad design and dangerous, because anyone can overwrite them...

slintes (changes requested):
review main.go only so far...

main.go
Tzif-Morgen: that was the actual behavior before; it was only shadowed by a nil pointer dereference

Tzif-Morgen: we can of course continue with not-supported

main.go
Tzif-Morgen: Absolutely right. changed, wdyt?

Tzif-Morgen:
Not relevant right now

PR #205, opened 2026-05-31
fix: remove finalizer when node is deleted and remediation is timed out by NHC

mshitrit:
#### Why we need this PR

In an escalation remediation scenario (FAR → MDR), FAR may fail to remediate in time, causing NHC to time it out and trigger MDR. MDR remediates by deleting the machine, which removes the node. When NHC subsequently tries to delete the FAR CR, the finalizer blocks the deletion forever — the controller returns early on node == nil before it ever reaches the NHC-timeout + DeletionTimestamp check. The FAR CR is then permanently stuck, leaking a finalizer on a remediation that is already abandoned.

#### Changes made
- Controller: handle the case where the node is gone but NHC has timed out and is deleting the CR — remove the finalizer directly so the deletion can complete.
- Tests: add regression test for the scenario; harden node teardown to tolerate already-deleted nodes.

[RHWA-697](https://redhat.atlassian.net/browse/RHWA-697)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved FenceAgentsRemediation deletion handling so reconciliation completes reliably when NHC timeout is set and the target node is missing. Finalizers are removed correctly, in-flight executor work is stopped during deletion, and taint-cleanup is skipped when the node is already gone.

* **Tests**
  * Added/extended reconciliation coverage for (1) NHC timeout with node deleted before FAR deletion and (2) node deleted before FAR CR deletion, verifying eventual full CR cleanup and finalizer removal.

### DIFF
--- a/internal/controller/fenceagentsremediation_controller.go
+++ b/internal/controller/fenceagentsremediation_controller.go
@@ -119,43 +119,54 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		r.Log.Error(err, "Unexpected error when validating CR's name with nodes' names", "CR's Name", req.Name)
 		return emptyResult, err
 	}
-	if node == nil {
-		// Node not found — it may have been deleted by MDR (or another remediator).
-		// If NHC has already timed out and is also trying to delete this CR, the
-		// finalizer would block that deletion forever since there is no node to clean up.
-		// Remove the finalizer directly so the pending deletion can complete.
-		if isTimedOutByNHC(far) && far.DeletionTimestamp != nil {
-			r.Executor.Remove(far.GetUID())
-			r.Log.Info("Node not found; removing finalizer to unblock NHC deletion of timed-out remediation", "remediation name", far.GetName())
-			controllerutil.RemoveFinalizer(far, v1alpha1.FARFinalizer)
-			if err := r.Client.Update(ctx, far); err != nil {
-				return emptyResult, fmt.Errorf("failed to remove finalizer from CR - %w", err)
-			}
-			return emptyResult, nil
+
+	// Handle deletion first - this is the standard reconcile pattern
+	// Check if CR is being deleted (has DeletionTimestamp)
+	if !far.ObjectMeta.DeletionTimestamp.IsZero() {
+		r.Log.Info("CR deletion timestamp is set", "CR Name", req.Name)
+
+		// Log remediation state if it didn't complete successfully
+		if !meta.IsStatusConditionPresentAndEqual(far.Status.Conditions, commonConditions.SucceededType, metav1.ConditionTrue) {
+			processingCondition := meta.FindStatusCondition(far.Status.Conditions, commonConditions.ProcessingType).Status
+			fenceAgentActionSucceededCondition := meta.FindStatusCondition(far.Status.Conditions, utils.FenceAgentActionSucceededType).Status
+			succeededCondition := meta.FindStatusCondition(far.Status.Conditions, commonConditions.SucceededType).Status
+			r.Log.Info("FAR didn't finish remediate the node", "CR Name", req.Name, "processing condition", processingCondition,
+				"fenceAgentActionSucceeded condition", fenceAgentActionSucceededCondition, "succeeded condition", succeededCondition)
 		}
-		r.Log.Error(err, "Could not find CR's target node", "CR's Name", req.Name, "Expected node name", v1alpha1.GetNodeName(far))
-		utils.UpdateConditions(utils.RemediationFinishedNodeNotFound, far, r.Log)
-		commonEvents.WarningEvent(r.Recorder, far, utils.EventReasonCrNodeNotFound, utils.EventMessageCrNodeNotFound)
-		return emptyResult, err
+
+		// Always stop the executor on deletion - it's a no-op if already stopped
+		// This ensures cleanup even if conditions are corrupted/missing
+		r.Executor.Remove(far.GetUID())
+
+		// handleFARDeletion is nil-safe: if node is missing (e.g., deleted by MDR or another remediator),
+		// it skips taint removal and proceeds directly to finalizer removal
+		return r.handleFARDeletion(ctx, far, node)
 	}
 
 	// Check NHC timeout annotation
+	// When NHC times out a remediation, it annotates and then deletes the CR
 	if isTimedOutByNHC(far) {
-		r.Executor.Remove(far.GetUID())
-		if far.DeletionTimestamp != nil {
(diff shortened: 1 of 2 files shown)


slintes (changes requested):
Concern: code duplication. The new block inside node == nil duplicates the existing NHC timeout+deletion logic

 Existing:

```
 if isTimedOutByNHC(far) {
     if far.DeletionTimestamp != nil {
         return r.handleFARDeletion(ctx, far, node)
     }
     // ...conditions, events...
 }
```

 New:
```
 if isTimedOutByNHC(far) && far.DeletionTimestamp != nil {
     controllerutil.RemoveFinalizer(far, v1alpha1.FARFinalizer)
     r.Client.Update(ctx, far)
     return emptyResult, nil
 }
```

Same intent, two code paths. Any future change to NHC-timeout cleanup must be made in both places.

Recommended alternative: reorder, don't duplicate

Move the isTimedOutByNHC check before the node == nil check. Then make handleFARDeletion nil-safe for
 node: when node is nil, skip taint removal (node is gone, no taints to clean). This gives one code path for NHC timeout regardless of whether node exists. Matches SNR pattern. No duplication.

 handleFARDeletion change would be minimal — guard node.Name usages:

```
func (r *FenceAgentsRemediationReconciler) handleFARDeletion(ctx context.Context, far *v1alpha1.FenceAgentsRemediation, node
 *corev1.Node) (ctrl.Result, error) {
     if node != nil {
         // existing taint removal logic (lines 257-287)
     }
     // finalizer removal (lines 289-296) — unchanged
 }
```

internal/controller/fenceagentsremediation_controller.go
```diff
@@ -119,20 +119,15 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
 	}
-	if node == nil {
-		r.Log.Error(err, "Could not find CR's target node", "CR's Name", req.Name, "Expected node name", v1alpha1.GetNodeName(far))
-		utils.UpdateConditions(utils.RemediationFinishedNodeNotFound, far, r.Log)
-		commonEvents.WarningEvent(r.Recorder, far, utils.EventReasonCrNodeNotFound, utils.EventMessageCrNodeNotFound)
-		return emptyResult, err
-	}
 
-	// Check NHC timeout annotation
+	// Check NHC timeout annotation first, regardless of node existence
```
clobrano: The description could clarify that we need to be able to remove the remediation CR regardless the existence of the Node.

internal/controller/fenceagentsremediation_controller.go
```diff
@@ -253,39 +260,44 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 func (r *FenceAgentsRemediationReconciler) handleFARDeletion(ctx context.Context, far *v1alpha1.FenceAgentsRemediation, node *corev1.Node) (ctrl.Result, error) {
 	emptyResult := ctrl.Result{}
 
-	// remove out-of-service taint when using OutOfServiceTaint remediation
-	if far.Spec.RemediationStrategy == v1alpha1.OutOfServiceTaintRemediationStrategy {
-		r.Log.Info("Removing out-of-service taint", "Fence Agent", far.Spec.Agent, "Node Name", node.Name)
-		taint := utils.CreateOutOfServiceTaint()
+	// Only attempt taint removal if node still exists (it may have been deleted by MDR or another remediator)
+	if node != nil {
```
clobrano, line 268: Checking first `if node == nil` would allow early return, improving readability
```go
if node == nil {
  // log and return
}

// rest of the code

**Review by clobrano — CHANGES_REQUESTED** [review]
I left a some comments

File: internal/controller/fenceagentsremediation_controller.go (hunk 2 above)
Comment (mshitrit) [reply internal/controller/fenceagentsremediation_controller.go:268]: early return will prevent finilizer removal later on

**Review by clobrano — DISMISSED** [review]

File: internal/controller/fenceagentsremediation_controller.go
Comment (slintes) [inline internal/controller/fenceagentsremediation_controller.go:183]: node can be nil here when deletion timestamp is set, finalizer was removed, but CR wasn't garbage collected yet...

File: internal/controller/fenceagentsremediation_controller.go [hunk 3]
```diff
@@ -141,6 +137,18 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
 		return emptyResult, err
 	}
 
+	if node == nil {
+		// Only error out if FAR is not being deleted
+		// If FAR is being deleted, allow it to proceed to the deletion logic below
+		if far.ObjectMeta.DeletionTimestamp.IsZero() {
+			r.Log.Error(err, "Could not find CR's target node", "CR's Name", req.Name, "Expected node name", v1alpha1.GetNodeName(far))
```
slintes: err is always nil here, so no need to log it (or create an error)

slintes (changes requested):
The execution flow was already hard to read and maintain and now it's worse. The number of checks of the deletion timestamp is a perfect sign for this.
This is worth a refactoring IMHO.

The usual reconcile flow is to check and handle deletion timestamp very early.
Then check NHC timeout and nil node.
After that we are clear of "special" cases and can on with the "normal" reconcile flow, setting finalizer etc.

WDYT?

internal/controller/fenceagentsremediation_controller.go
```diff
@@ -119,30 +119,51 @@ func (r *FenceAgentsRemediationReconciler) Reconcile(ctx context.Context, req ct
...
+		r.Log.Info("CR deletion timestamp is set", "CR Name", req.Name)
+
+		// Stop any ongoing remediation if it didn't complete successfully
+		if !meta.IsStatusConditionPresentAndEqual(far.Status.Conditions, commonConditions.SucceededType, metav1.ConditionTrue) {
+			processingCondition := meta.FindStatusCondition(far.Status.Conditions, commonConditions.ProcessingType).Status
+			fenceAgentActionSucceededCondition := meta.FindStatusCondition(far.Status.Conditions, utils.FenceAgentActionSucceededType).Status
+			succeededCondition := meta.FindStatusCondition(far.Status.Conditions, commonConditions.SucceededType).Status
+			r.Log.Info("FAR didn't finish remediate the node", "CR Name", req.Name, "processing condition", processingCondition,
+				"fenceAgentActionSucceeded condition", fenceAgentActionSucceededCondition, "succeeded condition", succeededCondition)
+			r.Executor.Remove(far.GetUID())
```
slintes: would it make sense to always remove the executor? I feel like we could miss something when conditions are not as expected...

slintes (changes requested):

slintes (approved):

PR #209, opened 2026-07-09
Integrate OpenShift cluster TLS security profile

mshitrit:
#### Why we need this PR

Operators running on OpenShift should respect cluster-wide TLS security policies configured by administrators. Currently, FAR uses hardcoded TLS defaults regardless of the cluster's security profile, potentially violating compliance requirements or organizational security policies.

#### Changes made

- Added cluster TLS profile detection that reads OpenShift APIServer configuration
- Integrated library-go for converting TLS profiles to Go crypto/tls configuration
- Implemented graceful fallback to library defaults on vanilla Kubernetes
- Added fail-closed error handling for security-critical configuration failures
- Created comprehensive test coverage for all profile types and edge cases

[RHWA-1309](https://redhat.atlassian.net/browse/RHWA-1309)

#### Test plan

Unit tests verify TLS profile detection, conversion logic, and error handling for Old/Intermediate/Modern/Custom profiles. Tests also cover vanilla K8s detection and invalid configuration scenarios.


cmd/main.go
```diff
@@ -96,17 +104,25 @@ func main() {
...
 
-	// TLS options for metrics server: disable HTTP/2 for mitigating CVEs
-	var tlsOpts []func(*tls.Config)
-	if !enableHTTP2 {
-		tlsOpts = append(tlsOpts, func(c *tls.Config) {
-			c.NextProtos = []string{"http/1.1"}
-		})
+	// Fetch cluster TLS profile and create TLS options (once for both servers)
+	ctx := context.Background()
+	tlsOpts, err := tlsconfig.CreateTLSOptsForServer(ctx, tempClient, !enableHTTP2)
```
razo7, line 119: suggestion: `context.Background()` provides no deadline. If the API server is slow or unresponsive during startup (e.g., during cluster upgrade), the operator hangs indefinitely before `CrashLoopBackOff` recovers it.

Consider:
```go
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()
```

pkg/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// getEffectiveTLSProfile returns the effective TLS profile based on the APIServer configuration.
+// Returns an error if the profile type is invalid or misconfigured.
+func getEffectiveTLSProfile(profileSpec *configv1.TLSSecurityProfile, profileType configv1.TLSProfileType) (*configv1.TLSProfileSpec, error) {
+	// Handle custom profile separately
+	if profileType == configv1.TLSProfileCustomType {
+		if profileSpec == nil || profileSpec.Custom == nil {
+			return nil, fmt.Errorf("Custom TLS profile is invalid or incomplete")
+		}
```
razo7, line 92: nit: Go convention -- error strings should not be capitalized (per `go vet` / `errcheck` guidelines). Should be `"custom TLS profile is invalid or incomplete"`.

pkg/tlsconfig/tlsconfig_test.go
```diff
@@ -0,0 +1,446 @@
...
+
+			if tlsConfig.MinVersion != tt.expectMinTLS {
+				t.Errorf("MinVersion = %v, want %v", tlsConfig.MinVersion, tt.expectMinTLS)
+			}
+		})
+	}
+}
+
+func TestGetClusterTLSConfig_VanillaK8s(t *testing.T) {
+	// Test vanilla K8s behavior with proper scheme registration
```
razo7: This test doesn't exercise the production vanilla K8s path. The fake client with `configv1` in its scheme but no `APIServer` object returns `IsNotFound`, which takes the error path (`tlsconfig.go:67`). The production vanilla K8s path is `IsNoMatchError` -> `return nil, nil` (`tlsconfig.go:64`).

The test name says "VanillaK8s" but it actually tests "OpenShift with deleted APIServer config." Consider using `interceptor.Funcs` to return a `&meta.NoKindMatchError{GroupKind: schema.GroupKind{Group: "config.openshift.io", Kind: "APIServer"}}` and verify the `nil, nil` return.

pkg/tlsconfig/tlsconfig_test.go
```diff
@@ -0,0 +1,446 @@
...
+			ctx := context.Background()
+			tlsConfig, err := getClusterTLSConfig(ctx, fakeClient)
+
+			if (err != nil) != tt.expectError {
+				t.Errorf("getClusterTLSConfig() error = %v, expectError %v", err, tt.expectError)
+				return
+			}
+
+			if tlsConfig.MinVersion != tt.expectMinTLS {
+				t.Errorf("MinVersion = %v, want %v", tlsConfig.MinVersion, tt.expectMinTLS)
```
razo7, line 119: suggestion: `tlsConfig` could be nil if a future test case triggers an unexpected `(nil, nil)` return. Add a nil guard before field access:
```go
if tlsConfig == nil {
    t.Fatal("getClusterTLSConfig() returned nil config, expected non-nil")
}
```

config/rbac/role.yaml
```diff
@@ -39,6 +39,14 @@ rules:
...
   verbs:
   - create
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - apiservers
+  verbs:
+  - get
+  - list
+  - watch
```
razo7: The code only calls `client.Get()` once. Do we really need the other verbs? Can we reduce RBAC to `get` only?

pkg/tlsconfig/tlsconfig_test.go
```diff
@@ -0,0 +1,446 @@
...
+		}
+
+		if len(testConfig.NextProtos) > 0 {
+			t.Errorf("NextProtos should be empty when HTTP/2 enabled, got %v", testConfig.NextProtos)
+		}
+	})
+}
+
+// TestApplyTLSConfigWithHTTP2Control removed - this functionality is now handled
+// directly in CreateTLSOptsForServer which is tested separately
```
razo7: nit: dead comment about a removed test -- can be deleted.

razo7 (commented):
Nice PR, see below for some things to consider with regard to Vanilla K8s (and none OCP cluster) and my inline comments

1. **Vanilla K8s path is not actually tested** -- `TestGetClusterTLSConfig_VanillaK8s` acknowledges the fake client can't produce `IsNoMatchError`. The test asserts the opposite of production behavior (expects an error instead of graceful nil return). Use `interceptor.Funcs` or a custom stub returning `meta.NoKindMatchError`.

2. **`CreateTLSOptsForServer` vanilla K8s path** -- only tested with an OpenShift `APIServer` object present. Add a test verifying that when `getClusterTLSConfig` returns nil (vanilla K8s), only the HTTP/2 opt is returned.

3. **Test with real Intermediate profile ciphers** -- the profile includes DHE ciphers that Go silently drops. A test using `configv1.TLSProfiles[configv1.TLSProfileIntermediateType]` directly would verify the cipher conversion handles real-world profiles correctly.

pkg/tlsconfig/tlsconfig_test.go
mshitrit: I don't think it's relevant in a deterministic test code

razo7:
I am closing the PR after the replacement PR was merged #214 yesterday.
See more at https://github.com/medik8s/fence-agents-remediation/pull/214#issuecomment-5015864939 for the new PR improvements that superseded the current PR implementation (plus the other operators adopted the 214 changes).

PR #221, opened 2026-09-11
make a standard deployment for src olm

weshayutin:
same as https://github.com/medik8s/storage-based-remediation/pull/108
https://github.com/medik8s/self-node-remediation/pull/345
https://github.com/medik8s/node-maintenance-operator/pull/186
https://github.com/medik8s/machine-deletion-remediation/pull/195

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,28 +1,14 @@
 # Build the manager binary
-FROM quay.io/centos/centos:stream9 AS builder
-RUN dnf install -y jq git \
-    && dnf clean all -y
+FROM quay.io/konveyor/builder:ubi9-latest AS builder
+ARG TARGETOS
+ARG TARGETARCH
 
 WORKDIR /workspace
-# Copy the Go Modules manifests for detecting Go version
-COPY go.mod go.mod
-COPY go.sum go.sum
+COPY go.mod go.sum ./
 
-RUN \
-    # get Go version from mod file
-    export GO_VERSION=$(grep -oE "toolchain go[[:digit:]]\.[[:digit:]]+\.[[:digit:]]" go.mod | awk '{print $2}') && \
-    echo ${GO_VERSION} && \
-    # find filename for latest z version from Go download page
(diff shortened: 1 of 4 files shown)


mpryc:
Same comment as https://github.com/medik8s/storage-based-remediation/pull/108#issuecomment-5639436319

Makefile
```diff
@@ -523,7 +523,7 @@ container-build: docker-build bundle-build ## Build containers
 
 .PHONY: bundle-build-community
 bundle-build-community: bundle-community ## Run bundle community changes in CSV, and then build the bundle image.
-	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
+	$(CONTAINER_TOOL) build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
```
mshitrit: CONTAINER_TOOL is only defined in `dev.mk` , in case `dev.mk`  does not exist (because tools wasn't fetched) we'll get `"build: command not found"` error.

Maybe either:
- set a default value (i.e something like `CONTAINER_TOOL ?= docker`) - in case missing tool is acceptable
- or validate CONTAINER_TOOL is set with a clear error - in case tool is mandatory

[edit] just noticed a similar comment by coderabbit over [here](https://github.com/medik8s/fence-agents-remediation/pull/221/changes#r4086109874) (second one) so sorry for extra noise

Makefile
mshitrit: Probably should use CONTAINER_TOOL here instead of hardcode `docker`

mshitrit: Probably should use CONTAINER_TOOL here instead of hardcode `docker`

Dockerfile
```diff
@@ -1,28 +1,14 @@
 # Build the manager binary
-FROM quay.io/centos/centos:stream9 AS builder
-RUN dnf install -y jq git \
-    && dnf clean all -y
+FROM quay.io/konveyor/builder:ubi9-latest AS builder
+ARG TARGETOS
+ARG TARGETARCH
```
mshitrit: IIUC atm those variables are unused,  in order to support multi-arch those need to be consumed in the go build command (i.e something like `GOOS=$TARGETOS GOARCH=$TARGETARCH`)

